PR #14, opened 2018-01-09
Changes gridSize to be a static variable

danielgordon10:
Without this change, every time a new scene is loaded, the grid has to be reset because the Discrete object is created anew. Making it a static variable applies the change to any future scenes also loaded, but it can be changed later by calling initialize again.

This also adds in a check that the grid size is greater than some small value in case the user doesn't ever specify the grid size.


ekolve:
Can you modify the defaulting behavior so that the grid size is set at definition time since the character will get moved (in MoveCharacter) by the csharp default of 0, and then snap to grid back to zero. 

 I think it would make more sense to just enforce a range in the Initialize() method that checks that the value is above and below a min/max.  Then emit an error if its outside the range and set it to the default.

PR #15, opened 2018-01-09
fixes freeze after random_initialize is called

danielgordon10:


ekolve:
Good catch, thank you!

PR #63, opened 2018-06-20
add `Controller.fullscreen` for fullscreen/windowed option

valtron:


ekolve:
Thanks for the PR. I'll test this out and merge. Are you creating custom builds for Windows to run ai2thor?

valtron:
Yes.

ekolve:
K. I'll will look at adding Windows to the build script so that Windows users can skip creating custom builds.

ekolve:
i added the fullscreen option in a slightly different way - your code should continue to work. My method just passes the parameter whether its True or False so there isn't an assumption about what the default is for the player.

PR #69, opened 2018-06-28
added ability to see overlays in unity

danielgordon10:


ekolve:
Is this something you want exposed in the API?

danielgordon10:
Fair enough. I would say always on is a waste of compute, so I would add it as an additional shader only when in unity editor.

danielgordon10:
Closing to be replaced by https://github.com/allenai/ai2thor/pull/226

PR #74, opened 2018-07-19
Small fixes for Linux and earlier versions of Unity

jsanmiya:


ekolve:
For all the texturescom assets, we still use those in the project, but due to the licensing restrictions from textures.com (they don't all redistribution of the textures).  They can be downloaded from their website and placed into the directory unity/Assets/QuickMaterials/SourceTextures/

jsanmiya:
Thanks for the clarification about the meta files. I've reverted that change.

I've also clarified in the check-in comment that I'm bringing back support for Unity 2017.2, and noted that 2017.1 should also work. 

Thanks!

PR #89, opened 2018-09-14
ExhaustiveBFSController and OfflineController

mitchellnw:
Includes the ExhaustiveBFSController and OfflineController.
Main features:
- Large speed up for non-interactive navigation agents.
- Builds a graph so that you can compute shortest paths and optimal plans.

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -1065,7 +1065,7 @@ class BFSController(Controller):
         self.initialize_scene()
         while self.queue:
             self.queue_step()
-            #self.visualize_points(scene_name)
+            # self.visualize_points(scene_name)
 
         self.prune_points()
         #self.visualize_points(scene_name)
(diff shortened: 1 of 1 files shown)


ai2thor/controller.py
```diff
@@ -1293,3 +1293,627 @@ def queue_step(self):
...
+                best_path_len = len(path)
+
+        best_plan = []
+        if get_plan:
+            best_plan = self.optimal_plan(source_state, t)
+       
+        return best_path, best_path_len, best_plan
+
+    def visualize_plan(self, source, plan):
+        """ Visualize the best path from source to plan. """
```
mitchellnw: Visualize the plan (a series of actions) which begins at source.

mitchellnw: This comment should be "Visualizes a plan (series of actions) which begins at source."

PR #99, opened 2018-11-26
Added stability to PutObject checks and added a new field for individual object masks by class

danielgordon10:
Also checks for masks that may be visible through closed receptacles and ignores them if that's the case. This won't deal with things like going through the side of the trash can.


ekolve:
Thanks for the PR. Looks good.  Could you write up a short (1 - 2 sentence description) for class_masks_individual that I can add to http://ai2thor.allenai.org/tutorials/event-metadata.  Similar to class_detections2D.

PR #112, opened 2019-02-21
WebGL builds 

AlvaroHG:
* Adds communication between C# and Javascript
* Adds task to create WebGL builds with one or multiple scenes
* First steps to tackle WebGL quality settings and differences in rendering


unity/ProjectSettings/QualitySettings.asset
```diff
@@ -211,7 +240,7 @@ QualitySettings:
     PSP2: 2
     Standalone: 6
     Tizen: 2
-    WebGL: 3
+    WebGL: 5
```
ekolve: I think we need to bump all the default values >= 5.  So prior to your change 6 == Ultra, but now 7 == Ultra, so the standalone setting needs to get updated.

AlvaroHG: Right, changed.

AlvaroHG:
Merging, Thanks!

PR #115, opened 2019-02-23
Unity 2018.3.6

AlvaroHG:
* Upgrades thor with physics to Unity 2018.3.6f1
* Fixes WebGL shader compilation that caused lights to bleed
* Upgrades deprecated API calls


AlvaroHG:
Merging!

PR #132, opened 2019-03-14
Fix build_url

jkulhanek:


ekolve:
Fixed with separate commit before I noticed your PR.

Issue #154, opened 2019-04-01
an envirionment bug


YunlianMoon:
There maybe an bug in AI2-THOR, when I use controller.step(dict(action='TeleportFull', x=, y=, z=, rotation=, horizon=)) function.
1. I found horizon (-60, -30) are look up degrees, while (30, 60) are look down degrees.
2. controller.step(dict(action='TeleportFull', x=, y=, z=, rotation=, horizon=60.0)) also works, while official website (http://ai2thor.allenai.org/documentation/actions) said that horizon valid values are -60 to +30 degrees.

winthos:
Hello! 
So the `TeleportFull` action is a little strange, because it allows you to hard set the horizon value of the agent without any restriction. Normally, the `LookUp` and `LookDown` actions limit the agent's horizons in 30 degree increments. `TeleportFull` allows you to set whatever horizon angle you want, even something like 14.8 as there is no restriction of 30 degree increments.

The difference in the negative (-30, -60) "upward" look angles is due to the way horizon angles are defined within the Unity Framework.

![THOR_Horizon_Diagram](https://user-images.githubusercontent.com/17502636/55345247-47d33e80-5464-11e9-9a01-bb8c89c0cbd2.png)

I threw together this diagram to better describe why those numbers seem wrong. Because the X-axis is defined to be "right" facing given that the forward axis is defined by the Z-axis, the negative values indicate a rotation about the X-axis that ends up making the Z-axis point "upward." This is why the -30 horizon value angles the Agent's forward angle 30 degrees "up," while a 30 horizon angle will change the forward angle to 30 degrees "down."

We will update the documentation to more accurately describe how `TeleportFull` is used, as the way angles are set with `TeleportFull` are slightly different than `LookUp` and `LookDown`.

PR #155, opened 2019-04-01
Added conversion for TeleportFull horizon values.

winthos:
Users now can use 30, 0, -30, -60 as expected rather than dealing with Unity's weird x axis rotation orientation nonsense

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -62,6 +62,17 @@ namespace UnityStandardAssets.Characters.FirstPerson
 		protected float[] headingAngles = new float[] { 0.0f, 90.0f, 180.0f, 270.0f };
 		protected float[] horizonAngles = new float[] { 60.0f, 30.0f, 0.0f, 330.0f };
 
+		//When passing in the TeleportFull action's horizon value, these are the only values that should be allowed
+		//they will be translated to the actual horizonAngles used above, as 30 should rotate the agent's camera 30 degrees up, but actually the
+		//value 330 or -30 about the x axis is what will angle the camera up... so yeah this is just to make it less confusing for the user
+		protected Dictionary<int, float> validHorizonAngleValues = new Dictionary<int, float>()
+		{
+			{30, -30.0f},
+			{0, 0.0f},
+			{-30, 30.0f},
+			{-60, 60.0f},
+		};
+
 		//allow agent to push sim objects that can move, for physics
 		protected bool PushMode = false;
 		protected int actionCounter;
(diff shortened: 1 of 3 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -62,6 +62,17 @@ public bool IsVisible
...
+		//they will be translated to the actual horizonAngles used above, as 30 should rotate the agent's camera 30 degrees up, but actually the
+		//value 330 or -30 about the x axis is what will angle the camera up... so yeah this is just to make it less confusing for the user
+		protected Dictionary<int, float> validHorizonAngleValues = new Dictionary<int, float>()
+		{
+			{30, -30.0f},
+			{0, 0.0f},
+			{-30, 30.0f},
+			{-60, 60.0f},
+		};
+
```
Lucaweihs: I think it makes more sense for a horizon of 30 to indicate a rotation of 30 degrees downwards rather than having to do this conversion to do the opposite. I also worry that this breaks backwards compatibility.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -1052,6 +1052,21 @@ public void TeleportObjectToFloor(ServerAction action)
 
         public void TeleportFull(ServerAction action) {
 			targetTeleport = new Vector3 (action.x, action.y, action.z);
+            float targetHorizon;
+
+            //check to make sure valid horizon angles are passed in
+            if(!validHorizonAngleValues.ContainsKey(action.horizon))
```
Lucaweihs: You should do `action.horizon % 360` here as someone might be doing things programmatically and thus have generated a horizon that is off by a constant multiple of 360 degrees. This means you'll need to update the keys of `validHorizonAngleValues` also.

Lucaweihs: If `forceAction` is `true` then all of these checks should be ignored.

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: hmm, maybe there is a better way to structure this. Basically I was trying to address this post: https://github.com/allenai/ai2thor/issues/154

Lucaweihs: Ah, I would just mark #154 as #wontfix as this isn't a bug, if anything I would just make the documentation clearer.

ekolve:
The motivation for how TeleportFull works is that you should be able to take the values from the metadata['agent']['position'] metadata['agent]['rotation'] from some previous event and pass them back to TeleportFull.  Flipping the sign of the horizon I think will just add further confusion since it now makes the values inconsistent with what we return the to user.

PR #157, opened 2019-04-09
New RemoveFromScene action...

winthos:
Added a new action allowing specific sim objects (by string id) to be removed from a scene.

Also found a couple scene fixes - moved some objects, added in some missing objects

### DIFF
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -5821,7 +5821,37 @@ namespace UnityStandardAssets.Characters.FirstPerson
                 }
             }
             actionFinished(true);
-		}  
+		}
+
+        //remove a given sim object from the scene. Pass in the object's uniqueID string to remove it.
+        public void RemoveFromScene(ServerAction action)
+        {
+            //pass name of object in from action.objectId
+            if (action.objectId == null)
+            {
+                Debug.Log("Hey, actually give me an object ID to open, yeah?");
+                errorMessage = "objectId required for OpenObject";
+                actionFinished(false);
+                return;
+            }
+
+            PhysicsSceneManager script = GameObject.Find("PhysicsSceneManager").GetComponent<PhysicsSceneManager>();
+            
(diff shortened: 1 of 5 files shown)


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5821,7 +5821,37 @@ Vector3[] reachablePositions
...
+
+            PhysicsSceneManager script = GameObject.Find("PhysicsSceneManager").GetComponent<PhysicsSceneManager>();
+            
+            //see if the object exists in this scene
+            foreach (SimObjPhysics sop in script.PhysObjectsInScene)
+            {
+                if(sop.uniqueID == action.objectId)
+                {
+                    sop.transform.gameObject.SetActive(false);
+                    script.SetupScene();
```
AlvaroHG: Do we need to call SetupScene here?

winthos: yes, it updates the active list of sim objects in the scene on the PhysicsSceneManager object.

AlvaroHG:
Looked mostly at the code. Not sure what the scene file changes are. Left a comment, LGTM after that.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
AlvaroHG: Maybe we should give it a more specific name if it's all it does, since it sounds like a "run at startup only" kind of function that makes you think it's expensive to call.

winthos: hmm, it's probably fine as is. It does some other cleanup beyond just updating the active list. It is required whenever any sim object in the scene needs to be destructively changed in some way. It is rather expensive and shouldn't be run unless major changes have happened to the scene, like removing or adding objects.

AlvaroHG: Fair enough, we should look at refactoring it if more actions like this are added to only do the necessary things.

PR #161, opened 2019-04-11
Fixing some warnings about unused variables and an error when determining if a target position was within scene bounds.

Lucaweihs:
Also updated the `OpenObjectAtLocation` method.


winthos:
looks good! go ahead and merge

PR #162, opened 2019-04-17
Removed extra right angled bracket in 4 statements in README.md

beduffy:
Simple but important fix. Noticed it when I was copying these commands to test ai2thor 1.0


winthos:
fixed in 2.0 release

PR #185, opened 2019-05-14
Input refactor for Turk

AlvaroHG:
* Refactors player input into components that govern all the input at one time, so that input code in multiple places does not compete with each other.
* System for enabling and disabling components that handle input
* Moves input code out of DebugInputField, and into a specific discrete debug input component
* Adds new point and click input for hide n seek evaluation tasks
* Cleans up DebugFPS component for new input system
* Abstracts object highlight and pickup functionality to be used by different controllers
* Uses preprocessor macros to inform the build system of a particular controller build, e.g. a build for turk.
* Changes in agent manager and base fps controller to send player input events as metadata to javascript.
* In editor you can cycle between different controllers by pressing `~`, `1` and `2` keys.


AlvaroHG:
FYI @Lucaweihs, and @winthos would be good if you can give it a quick look :)

winthos:
looks good to me. I'm going to need a quick breakdown of how to control things in editor so I can use the new inputs to help debug!

AlvaroHG:
Yes, I tried to change the least possible; it defaults to text input mode, `~` still enables fps mode `1` gets you to text input mode and `2` gets you into the new point and click mode.

AlvaroHG:
merging

PR #225, opened 2019-06-12
made messages variable length and increased bad JSON robustness

danielgordon10:


ekolve:
Can you provide a test case for how the previous implementation would fail? Currently we read the HTTP header from the cilent into a 1K buffer and parse the content-length. Then construct a buffer of that size for the body.  The only time this would fail would be if the header grew greater than 1KB, which would very strange and would only happen if we started sending cookies.

danielgordon10:
I made this change before you moved to HTTP 1.1, so it might not be necessary anymore. Some of the changes I added are still good though. I will revert the socket logic.

danielgordon10:
Was this change actually needed after your reworking?

ekolve:
The main change with the PR was the exception logging.  We technically should never have malformed JSON, but I also don't like the experience where if we do run in to this ever again Unity basically hangs and the user has to do a bunch of debugging to narrow it down to a JSON issue.  At least now its logged explicitly.

PR #226, opened 2019-06-12
Image overlays

danielgordon10:
Adding flow back and enabling overlays in Unity editor.


danielgordon10:
Can this be merged? At least into WIPFeatures? I was able to merge the changes no problem on my end.

winthos:
All the functionality of these changes have already been implemented into WIPFeatures. It was part of the image synthesis update. We had to rewrite a bit of your original logic into our own version to account for some additional changes on our end, but all the functionality should already exist. The biggest change is that the image synth isn't turned on by default in editor. It must be toggled on using the debug field "initsynth".

danielgordon10:
You also added flow?

winthos:
Yes, we copied your original shader. Check the current image synthesis script to confirm.

PR #299, opened 2019-09-06
Moveable objects

ekolve:


winthos:
already pulled via command line

PR #310, opened 2019-10-31
Interact controller upgrade, image fixes

AlvaroHG:
* Upgrades interact invoke task to save images of real agent at every step
* Flips RB values for agent's color image
* Option to flip image vertically and horizontally
* Flips depth image horizontally, as it was incorrect


AlvaroHG:
Merging

PR #318, opened 2019-12-04
Shortest Path Setup

AlvaroHG:
Setup and utility functions for getting Shortest Path to objects using navmeshes.


drschwenk:
I don't see any issues or conflicts with my plans

PR #344, opened 2020-01-29
Depth fixes

AlvaroHG:
* Unifies depth with real
* Supports different units
* Shader rewrites without hacks and magic numbers

### DIFF
--- a/ai2thor/server.py
+++ b/ai2thor/server.py
@@ -27,6 +27,8 @@ import werkzeug.serving
 import werkzeug.http
 import numpy as np
 
+from enum import Enum
+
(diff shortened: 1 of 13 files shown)


ai2thor/server.py
```diff
@@ -27,6 +27,8 @@
 import werkzeug.http
 import numpy as np
 
+from enum import Enum
```
ekolve: Is there a python 2 equivalent? enum isn't available in python 2.7.

AlvaroHG: ok will look for that, we already are using `Enum` in `interact.py`

PR #350, opened 2020-02-03
Depth noise

AlvaroHG:
* task for analyzing depth components in real image and save noise representation
* task for transforming real metadata into a sim position and getting depth
* flag in controller to add noise to sim images
* fixes on controllers, and expanded flags for interact and get-depth task


tasks.py
```diff
@@ -941,22 +951,34 @@ def get_depth(
...
+            t_size = len(teleport)
+            if 1 <= t_size:
+                teleport_arg['x'] = teleport[0]
+            if 2 <= t_size:
+                teleport_arg['z'] = teleport[1]
+            if 3 <= t_size:
+                teleport_arg['y'] = teleport[2]
+
+        from pprint import pprint
+        pprint(teleport_arg)
```
ekolve: remove!

AlvaroHG: done

PR #354, opened 2020-02-11
Dataset generation

AlvaroHG:
* Scripts to generate dataset
* filter and validate dataset
* spl tweaks
* better error for reset


ai2thor/util/scene_utils.py
```diff
@@ -0,0 +1,9 @@
+import re
+
+def scene_names_key_func(scene_name):
```
ekolve: This looks identical to the function in reset()

PR #363, opened 2020-03-02
Release 2.3.5

AlvaroHG:
Bug fixes:
* Deprecated `continuous` and `continuousMode` flags for initialization, fixed non-stochastic agent set to `snapToGrid=False` to work properly
* ApplyNoise flag for the stochastic agent now also controls the small random rotation inside of the move action that tried to mimic the robot.
* Increased navmesh coverage for shortest paths
* Nav mesh voxel resolution increment
* Set navmesh agent bake height to match new model
* Manual fixes for navmeshes in scenes: 
    - FloorPlan_Train7_1, made navmesh agent radius smaller to fit between table where reachable 
       positions returned valid locations
    - FloorPlan_Train_11_3, unmade bed blocking walkability
    - FloorPlan_Val2_3, unmade bed blocking walkability

Features:
* Dataset Generation, filtering tasks
* Visualization of shortest_paths for point set
* Visualization of path
* Tasks for visualization


ekolve:
Where are the changes that deprecate the continuous flags? We shouldn't remove this functionality without emitting warnings for a few versions so users are aware. Have you talked with Roozbeh about removing continuous? Users can get the same behavior with the stochastic agent, but requires knowing that you need to remove noise.

AlvaroHG:
will do

AlvaroHG:
done

AlvaroHG:
tested locally merging

PR #378, opened 2020-03-19
Video controller merge

mattdeitke:
Allows for basic animations with the video controller and changes to the color of the skybox from python.

### DIFF
--- a/ai2thor/video_controller.py
+++ b/ai2thor/video_controller.py
@@ -144,7 +144,7 @@ class VideoController(Controller):
             # keep the position the same
             p = self.last_event.metadata['agent']['position']
             if smoothAnimation:
-                if agentId == None:
+                if agentId is None:
                     yield self.step(action='TeleportFull',
                         rotation=y0 + rotateDegrees * self._linear_to_smooth(i + 1, frames, std_dev=1),
                         agentId=agentId,
@@ -154,7 +154,7 @@ class VideoController(Controller):
                         rotation=y0 + rotateDegrees * self._linear_to_smooth(i + 1, frames, std_dev=1),
                         **p)
             else:
-                if agentId == None:
+                if agentId is None:
                     yield self.step(action='TeleportFull',
                         rotation=y0 + rotateDegrees * ((i + 1) / frames),
                         agentId=agentId, **p)
--- a/ai2thor/video_controller.py
+++ b/ai2thor/video_controller.py
@@ -78,22 +78,20 @@ class VideoController(Controller):
         self.ceiling_off = not self.ceiling_off
         return self.step(action='ToggleMapView')
 
-    def _cdf(self, x, std=0.5, mean=0.0):
+    def _cdf(self, x, std_dev=0.5, mean=0.0):
         '''Cumulative distribution function'''
         return (1.0 + erf((x - mean) / sqrt(2.0 * std_dev**2))) / 2.0
 
-    def _linear_to_smooth(self, curr_frame, total_frames, std=0.5, min_val=3):
+    def _linear_to_smooth(self, curr_frame, total_frames, std_dev=0.5, min_val=3):
         # start at -3 STD on a normal gaussian, go to 3 STD on gaussian
         # curr frame should be 1 indexed, and end with total_frames
         assert min_val > 0, "Min val should be > 0"
-
-        distribution = st.norm(0, std)
 
         if curr_frame == total_frames:
             # removes drifting
             return 1
 
-        return self._cdf(- min_val + 2 * min_val * (curr_frame / total_frames))
+        return self._cdf(- min_val + 2 * min_val * (curr_frame / total_frames), std_dev=std_dev)
 
     def _move(self, actionName, moveMagnitude, frames, smoothAnimation, agentId=None):
         """Yields a generator full of move commands to move the agent incrementally.
@@ -102,7 +100,7 @@ class VideoController(Controller):
         for i in range(frames):
             # smoothAnimation = False => linear animation
             if smoothAnimation:
-                next_moveMag = self._linear_to_smooth(i + 1, frames, std=1) * moveMagnitude
+                next_moveMag = self._linear_to_smooth(i + 1, frames, std_dev=1) * moveMagnitude
                 if agentId == None:
                     yield self.step(action=actionName, moveMagnitude=next_moveMag - last_moveMag)
                 else:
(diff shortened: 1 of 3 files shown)


ai2thor/video_controller.py
```diff
@@ -0,0 +1,336 @@
...
+            p = os.path.join(path, f'{i}.jpg')
+            if os.path.exists(p):
+                os.remove(p)
+            Image.fromarray(self.saved_frames[i]).save(p)
+        print('done')
+
+    def mergeVideo(self, otherVideoPath):
+        """Concatenates the frames of `otherVideoPath` to the presently
+           generated video within this class."""
+        import cv2
```
ekolve: This import can be removed since its done at the top.

ai2thor/video_controller.py
```diff
@@ -0,0 +1,336 @@
...
+                30,
+                (self.saved_frames[0].shape[1], self.saved_frames[0].shape[0])
+            )
+            for i, frame in enumerate(self.saved_frames):
+                print('|', end='')
+                # assumes that the frames are RGB images. CV2 uses BGR.
+                video.write(cv2.cvtColor(frame, cv2.COLOR_BGR2RGB))
+            cv2.destroyAllWindows()
+            video.release()
+            print('done')
```
ekolve: Instead of print statements these should either be removed or replaced with logger.info statements so that end users can control whether they want them to appear.

ai2thor/video_controller.py
```diff
@@ -0,0 +1,336 @@
...
+import os
+from PIL import Image
+import scipy.stats as st  # for smooth non-linear animations
+from queue import Queue
+import math
+
+
+class VideoController(Controller):
+    def __init__(self,
+                 initial_camera_rotation=dict(x=85, y=225, z=0),
```
ekolve: mutable default arguments is an anti-pattern:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

I'm terribly worried in this case, but its good to consider it.  For instance the following is possible:
```python
a = VideoController()
b = VideoController()
b.initial_camera_rotation['x'] = 100
print(a.initial_camera_rotation['x']) # this will equal 100 as well
c = VideoController()
print(c.initial_camera_rotation['x']) # new instances will equal 100 since the modification was done at the class level

```

The link above provides one solution, another way would be to simply create copy the default argument
```python
self.initial_camera_position = initial_camera_position.copy()
self.initial_camera_rotation = initial_camera_rotation.copy()
```

ai2thor/video_controller.py
```diff
@@ -0,0 +1,336 @@
...
+from ai2thor.controller import VideoController
+with VideoController() as vc:
+    vc.transform(vc.moveAhead())
+    vc.exportVideo('thor.mp4')
+Known issues:
+- Multi agent rotations don't work (since TeleportFull breaks when passing in an AgentID)
+"""
+
+from ai2thor.controller import Controller
+import cv2
```
ekolve: Can you add opencv-python to setup.py? Do you know if you need a particular version?

ai2thor/video_controller.py
```diff
@@ -0,0 +1,348 @@
...
+    def _cdf(self, x, std=0.5, mean=0.0):
+        '''Cumulative distribution function'''
+        return (1.0 + erf((x - mean) / sqrt(2.0 * std_dev**2))) / 2.0
+
+    def _linear_to_smooth(self, curr_frame, total_frames, std=0.5, min_val=3):
+        # start at -3 STD on a normal gaussian, go to 3 STD on gaussian
+        # curr frame should be 1 indexed, and end with total_frames
+        assert min_val > 0, "Min val should be > 0"
+
+        distribution = st.norm(0, std)
```
ekolve: can be removed

ai2thor/video_controller.py
```diff
@@ -0,0 +1,348 @@
...
+            cv2.destroyAllWindows()
+            video.release()
+            print('done')
+
+    def exportFrames(self, path):
+        """Exports all of the presently frames to the `path` directory.
+        
+        The frames are numbered in sequential order (starting with 0)."""
+        for i in range(len(self.saved_frames)):
+            p = os.path.join(path, f'{i}.jpg')
```
ekolve: might make more sense to use PNG instead of JPG to make this lossless.

ai2thor/video_controller.py
```diff
@@ -0,0 +1,346 @@
...
+        assert direction == 'left' or direction == 'right'
+        if direction == 'left': rotateDegrees *= -1
+
+        # get the initial rotation
+        y0 = self.last_event.metadata['agent']['rotation']['y']
+        for i in range(frames):
+            # keep the position the same
+            p = self.last_event.metadata['agent']['position']
+            if smoothAnimation:
+                if agentId == None:
```
ekolve: All == None should be changed to 'is None'

```python
if agentId is None
```

See https://stackoverflow.com/questions/14247373/python-none-comparison-should-i-use-is-or

ai2thor/video_controller.py
```diff
@@ -0,0 +1,346 @@
...
+            if os.path.exists(path):
+                os.remove(path)
+            print((self.saved_frames[0].shape[1], self.saved_frames[0].shape[0]))
+            video = cv2.VideoWriter(
+                path,
+                cv2.VideoWriter_fourcc(*'DIVX'),
+                30,
+                (self.saved_frames[0].shape[1], self.saved_frames[0].shape[0])
+            )
+            for i, frame in enumerate(self.saved_frames):
```
ekolve: Style thing, ```i``` is unused and can be replaced with ```_```

ai2thor/video_controller.py
```diff
@@ -0,0 +1,346 @@
...
+                print('|', end='')
+                # assumes that the frames are RGB images. CV2 uses BGR.
+                video.write(cv2.cvtColor(frame, cv2.COLOR_BGR2RGB))
+            cv2.destroyAllWindows()
+            video.release()
+            print('done')
+
+    def exportFrames(self, path):
```
ekolve: For method names that we aren't trying to match to the Unity actions it makes sense to follow the python convention of snake casing method names. So this would be ```export_frames```

ai2thor/video_controller.py
mattdeitke: Just going to remove the enumeration, which will remove i. Previously used i to internally track export progress.

PR #379, opened 2020-03-19
Adding Drone functionality to master

winthos:
adds new 'drone' mode agent is able to initialize into.

while in drone mode, physics and time updates are paused until an emitFrame executes, allowing for simulation of continuous flight between metadata generation events.

many new drone actions:

FlyTo, FlyAhead, FlyUp, FlyDown, FlyLeft, FlyRight, FlyBack, LaunchDroneObject, SpawnDroneLauncher, FlyAssignStart, FlyRandomStart, ChangeDronePositionRandomNoiseSigma,

New drone specific metadata:
object metadata: isCaught, numSimObjHits, numFloorHits, numStructureHits, lastVelocity
event metadata: droneCurrentTime


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -448,6 +457,41 @@ public void SetAgentMode(string mode)
...
             }
+
+            else if(whichMode == "drone")
+            {
+                //toggle first person character cull
+                fpcc.SwitchRenderersToHide(whichMode);
+
+                VisibilityCapsule = DroneVisCap;
+                m_CharacterController.center = new Vector3(0,0,0);
+                m_CharacterController.radius = 0.2f;
```
AlvaroHG: should this magic constant be in the prefab of the drone?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -448,6 +457,41 @@ public void SetAgentMode(string mode)
...
+                CapsuleCollider cc = this.GetComponent<CapsuleCollider>();
+                cc.center = m_CharacterController.center;
+                cc.radius = m_CharacterController.radius;
+                cc.height = m_CharacterController.height;
+
+                m_Camera.GetComponent<PostProcessVolume>().enabled = false;
+                m_Camera.GetComponent<PostProcessLayer>().enabled = false;
+
+                //camera position set forward a bit for drone
+                m_Camera.transform.localPosition = new Vector3(0, 0, 0.2f);
```
AlvaroHG: same comment as above

unity/Assets/Scripts/FlyingDrone.cs
```diff
@@ -25,20 +24,53 @@ void Update ()
...
+    {
+        bool caught_object_bool = false;
+        foreach (SimObjPhysics obj in caught_object)
+        {
+            if(obj.Type == check_obj.Type)
+            {
+                if(obj.name == check_obj.name)
+                {
+                    caught_object_bool = true;
+                    //Debug.Log("catch!!!");
```
AlvaroHG: remove comment

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -39,10 +40,14 @@ public class PhysicsRemoteFPSAgentController : BaseFPSAgentController {
...
+        //DRONE parameters
         [SerializeField] public bool FlightMode = false;
+        public bool hasFixedUpdateHappened = true;//track if the fixed physics update has happened
         protected Vector3 thrust;
-
-        [SerializeField] Camera[] FlightCameras;
+        public float dronePositionRandomNoiseSigma = 0f;
+        //count of fixed updates for use in droneCurrentTime
+        public float fixupdateCnt = 0f;
+        public float autoResetTimeScale = 1.0f;
```
AlvaroHG: I think these variables should be in a controller that inherits from `PhysicsRemoteFPSAgentController` just like the stochastic controller, something like `DroneAgentController`. Any sub class now has `dronePositionRandomNoiseSigma` which is not used, and drone actions should throw unsupported exceptions for the other controllers

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: I dont think so, the drone isn't a separate prefab, it's a modification of the FPSController prefab that toggles on all the drone settings. Both the "tall" and"bot" modes set the center and radius to the appropriate values for their modes as well.

unity/Assets/Scripts/FlyingDrone.cs
winthos: removed in push

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -203,12 +207,43 @@ public void SetDecayTemperatureBool(ServerAction action)
 
         // Update is called once per frame
         void Update() {
-            if (FlightMode) {
-                if (thrust.magnitude > 0.1) {
-                    m_CharacterController.Move(thrust * Time.deltaTime);
+
+        }
+
+        void FixedUpdate(){
```
AlvaroHG: you could override this and simplify the logic, and have each controller decide what fixed update

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: same as above as well, although maybe there is a way to store these values on a drone controller script like you suggested that inherits from `PhysicsRemoteFPSAgent` but I think there will be some issues with execution order... will have to double check that.

unity/Assets/Scripts/BaseFPSAgentController.cs
AlvaroHG: got it

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: agreed, I will start working on a drone controller that inherits from PhysicsRemoteFPSAgentController and see where that goes

AlvaroHG:
Looks good, some suggestions for trying to avoid future pains in the FPSController class

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: agreed, its messy enough as it is, and this might make it a bit more organized at the very least.

unity/Assets/Scripts/BaseFPSAgentController.cs
AlvaroHG: I guess they could be in the DroneAgentController class if you inherited PhysicsFPSController

winthos: yes agreed, I will move them there

PR #397, opened 2020-05-20
upgrading jsonnet to 12.3 and switching .NET to version 4; actionRetu…

ekolve:
…rn is now serialized using jsonnet and metadat uses unity3d jsonutility


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1390,7 +1390,6 @@ public virtual MetadataWrapper generateMetadataWrapper()
             metaMessage.lastAction = lastAction;
             metaMessage.lastActionSuccess = lastActionSuccess;
             metaMessage.errorMessage = errorMessage;
-            metaMessage.actionReturn = this.actionReturn;
```
AlvaroHG, line 1393: Should we still keep this? and set it to `actionReturns[0]` for the first agent to not break client code we have all around where we read `actionReturn` on the metadata to do certain things?

AlvaroHG:
Looks great! My only concern is deprecating `actionReturn` from metadata. I agree we need multiple returns and we should move to that, but there are probably many changes e.g. tasks in the repository where we check if an action fails by accessing `actionReturn`, that were not updated in the PR, and other external client code. I am thinking we should set `actionReturn` on the metadata object to the return of the first agent for now? or see how many places in the code it is and maybe fix those too before we submit this.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: We still pass actionreturn values as a separate form parameter and copy this in to the correct place in the metadata in server.py

PR #433, opened 2020-08-03
Update setup.py

jonborchardt:
this appears to be needed to install dependencies, but i don't know anything about the project...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 from setuptools import setup, find_packages
 
-exec(open('ai2thor/_version.py').read())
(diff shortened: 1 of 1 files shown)


ekolve:
This will break our pip packages. We can apply the change from the ExpRoom branch I committed earlier: https://github.com/allenai/ai2thor/commit/fca34ef8babdd8c35e9d62e302f0679443624b59

I'm on vacation so I can't do it at the moment.

jonborchardt:
closing, not the right sln.

PR #435, opened 2020-08-07
Jon fix 410

### LINKED ISSUE #410: Add FOV to thirdPartyCamera metadata / add support to update thirdPartyCamera FOV

jonborchardt:
fixes #410 

This code adds the ability to update the FOV for 3rd party cameras and tests it.

I have a few review questions in the code ill call out via comments.

### DIFF
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -33,6 +33,20 @@ class UnityTestController(ai2thor.controller.Controller):
     def prune_releases(self):
         pass
 
+# Defining const classes to lessen the possibility of a misspelled key
+# Review: This pattern is inconsistent with this file, but I think its a better direction
+# let me know if I should revert to match the rest of the file
+class Actions:
+    AddThirdPartyCamera = 'AddThirdPartyCamera'
+    UpdateThirdPartyCamera = 'UpdateThirdPartyCamera'
+
+class MultiAgentMetadata:
+    thirdPartyCameras = 'thirdPartyCameras'
+
+class ThirdPartyCameraMetadata:
+    position = 'position'
+    rotation = 'rotation'
+    fieldOfView = 'fieldOfView'
 
 controller = UnityTestController()
 controller.reset('FloorPlan28')
@@ -45,10 +59,10 @@ def teardown_module(module):
 #def controller():
 #    return c
 
-def assert_near(point1, point2):
-    assert point1.keys() == point2.keys()
+def assert_near(point1, point2, error_message=''):
+    assert point1.keys() == point2.keys(), error_message
     for k in point1.keys():
-        assert round(point1[k], 3) == round(point2[k], 3)
+        assert round(point1[k], 3) == round(point2[k], 3), error_message
 
 def test_rectangle_aspect():
     controller = UnityTestController(width=600, height=300)
@@ -61,7 +75,7 @@ def test_small_aspect():
     controller.reset('FloorPlan28')
     event = controller.step(dict(action='Initialize', gridSize=0.25))
     assert event.frame.shape == (64, 128, 3)
-    
+
 def test_lookdown():
 
     e = controller.step(dict(action='RotateLook', rotation=0, horizon=0))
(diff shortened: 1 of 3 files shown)


ai2thor/tests/test_unity.py
```diff
@@ -33,6 +33,20 @@ def __init__(self,**kwargs):
     def prune_releases(self):
         pass
 
+# Defining const classes to lessen the possibility of a misspelled key
+# Review: This pattern is inconsistent with this file, but I think its a better direction
+# let me know if I should revert to match the rest of the file
```
jonborchardt, line 38: This pattern is inconsistent with this file, but I think its a better direction
let me know if I should revert to match the rest of the file

ai2thor/tests/test_unity.py
```diff
@@ -110,18 +124,54 @@ def test_rotate_left():
...
+    expectedFieldOfViewDefault = 90.0
+    assert len(controller.last_event.metadata[MultiAgentMetadata.thirdPartyCameras]) == 1, 'there should be 1 camera'
+
+    e = controller.step(dict(action=Actions.UpdateThirdPartyCamera, thirdPartyCameraId=0, position=expectedPosition, rotation=expectedRotation))
+    camera = e.metadata[MultiAgentMetadata.thirdPartyCameras][0]
+    assert_near(camera[ThirdPartyCameraMetadata.position], expectedPosition, 'position should have been updated')
+    assert_near(camera[ThirdPartyCameraMetadata.rotation], expectedRotation, 'rotation should have been updated')
+    assert camera[ThirdPartyCameraMetadata.fieldOfView] == expectedInitialFieldOfView, 'fieldOfView should not have changed'
+
+    # 0 is a special case, since nullable float does not get encoded properly, we need to pass 0 as null
```
jonborchardt, line 158: note, it would be preferable to allow serializing nullables, but it looks like nullables dont get passed... so i need to reserve 0 as null.

ai2thor/video_controller.py
```diff
@@ -27,7 +27,7 @@ def __init__(self,
                  **controller_kwargs):
         super().__init__(continuous=True, **controller_kwargs)
         self.step(
-            action='AddThirdPartyCamera', 
+            action='AddThirdPartyCamera',
```
jonborchardt, line 30: lots of small formatting cleanup, they are all good... but let me know if you dont like them

ai2thor/video_controller.py
```diff
@@ -287,9 +262,14 @@ def AbsoluteCameraAnimation(self,
                                  'y': p0['y'] + (py - p0['y']) / frames * i,
                                  'z': p0['z'] + (pz - p0['z']) / frames * i},
                 }
-                if not (smartSkybox and maxY > p0['y'] + (py - p0['y']) / frames * i):
-                    kwargs['skyboxColor'] = 'black'
-                yield self.step(**kwargs)
+
+            # enables linear animation changes to the camera FOV
+            if FOVstart != None and FOVend != None:
```
jonborchardt, line 267: im not sure if/where this code is used, but this should work to update fov now

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -104,11 +104,11 @@ void Start()
         physicsSceneManager = GameObject.Find("PhysicsSceneManager").GetComponent<PhysicsSceneManager>();
 	}
 
-	private void initializePrimaryAgent() 
+	private void initializePrimaryAgent()
     {
```
jonborchardt, line 108: i did not clean up all the formatting issues... is there a reason the { are misaligned?
should we add linting?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -300,6 +300,11 @@ public BaseFPSAgentController ReturnPrimaryAgent()
 		ResetSceneBounds();
 	}
 
+	// If fov is <= min or > max, return defaultVal, else return fov
+	private float ClampFieldOfView(float fov, float defaultVal = 90f, float min = 0f, float max = 180f) {
```
jonborchardt, line 304: this could be used in other files, but i did not want to make the change larger

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -316,24 +321,15 @@ public BaseFPSAgentController ReturnPrimaryAgent()
...
-        {
-            //default to 90 fov on third party camera if nothing passed in, or if value is too large
-            fov = 90f;
-        }
-        else
-        {
-            fov = action.fieldOfView;
-        }
         if (action.orthographic) {
+			// REVIEW: should the orthographicSize use the clamped fov or the original value passed in?
```
jonborchardt, line 328: REVIEW: should the orthographicSize use the clamped fov or the original value passed in?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -952,6 +954,7 @@ public class ThirdPartyCameraMetadata
 	public int thirdPartyCameraId;
 	public Vector3 position;
 	public Vector3 rotation;
+	public float fieldOfView;
```
jonborchardt, line 957: this should be nullable, but when i tried, the serializer stopped emitting it.

ai2thor/tests/test_unity.py
AlvaroHG: you can make the `fieldOfView` a `nullable` in C#, using `?` and then if you don't pass anything or pass `None`, you can detect the nulled variable and have a special case for it. See `gridColor` in `ServerAction`.

ai2thor/video_controller.py
AlvaroHG: Can you include @mattdeitke in the review? he wrote this video controller to make sure the functionality is the same

unity/Assets/Scripts/AgentManager.cs
AlvaroHG: Orthographic should just use the FoV as passed by the action

unity/Assets/Scripts/AgentManager.cs
AlvaroHG: to python? I believe it works for `gridColor` but I am not sure.

If you already tried this then it's ok to use zero. You could also wrap it in another class `class Float` that uses the `[Serializable]` directive to box it and allow for null values.

ai2thor/video_controller.py
```diff
@@ -287,9 +262,14 @@ def AbsoluteCameraAnimation(self,
...
+                kwargs['fieldOfView'] = FOVstart + (FOVend - FOVstart) / frames * i
```
AlvaroHG, line 268: Don't you have to call `reset` and `TeleportFull` here too?

AlvaroHG:
Looks good! left a few comments, thank you for contributing :)
@mattdeitke can you verify the changes to the `RelativeCameraAnimation`? LGTM from me after that

unity/Assets/Scripts/AgentManager.cs
AlvaroHG: yes, we should. We are working on big backlog of features for different projects, let's talk about it next time we meet with Eric.

ai2thor/tests/test_unity.py
jonborchardt: i tried that, and when i did, the c# return object (`ThirdPartyCameraMetadata` in AgentManager.cs)
stopped returning a fieldOfView even when I was setting it... reverting to not use nullable fixed the issue. I would like this to work as you say, but i would need some help understanding why the serialized ThirdPartyCameraMetadata that is returned is missing the field even when it is not null.

ai2thor/video_controller.py
jonborchardt: done

unity/Assets/Scripts/AgentManager.cs
jonborchardt: added a note

ai2thor/video_controller.py
jonborchardt: i do not believe so, this was getting called because it was a new camera, now it is an updated camera
@mattdeitke can confirm

unity/Assets/Scripts/AgentManager.cs
jonborchardt: gridColor is never read in python... my guess is that this is not working as expected... but i could be wrong
maybe ill dig in and make a test to see

ai2thor/video_controller.py
mattdeitke: The video controller updates look good to me! (Assuming updating the ThirdPartyCamera's FOV on the Unity side works as expected.)

jonborchardt:
@ekolve is taking this and will merge it when he is ready, Thanks!

PR #439, opened 2020-08-10
using json.net for all serialization due to the use of primitive list…

ekolve:
…s (SceneBounds.cornerPoints), which get ignored by JsonUtility

### DIFF
--- a/ai2thor/tests/data/metadata-schema.json
+++ b/ai2thor/tests/data/metadata-schema.json
@@ -1,0 +1,666 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "type": "object",
+    "properties": {
+        "objects": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "type": "string"
+                    },
+                    "position": {
+                        "type": "object",
+                        "properties": {
+                            "x": {
+                                "type": "number"
+                            },
+                            "y": {
+                                "type": "number"
+                            },
+                            "z": {
+                                "type": "number"
+                            }
+                        },
+                        "required": [
+                            "x",
+                            "y",
+                            "z"
+                        ]
+                    },
+                    "rotation": {
+                        "type": "object",
+                        "properties": {
+                            "x": {
+                                "type": "number"
+                            },
+                            "y": {
+                                "type": "number"
+                            },
+                            "z": {
+                                "type": "number"
+                            }
+                        },
+                        "required": [
+                            "x",
+                            "y",
+                            "z"
+                        ]
+                    },
+                    "visible": {
+                        "type": "boolean"
+                    },
+                    "obstructed": {
+                        "type": "boolean"
+                    },
+                    "receptacle": {
+                        "type": "boolean"
+                    },
+                    "toggleable": {
+                        "type": "boolean"
+                    },
+                    "isToggled": {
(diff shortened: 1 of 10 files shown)


tasks.py
```diff
@@ -56,7 +56,7 @@ def push_build(build_archive_name, archive_sha256, include_private_scenes):
 def _local_build_path(prefix="local"):
     if platform.system() == "Darwin":
         suffix = "OSXIntel64.app"
-        build_path = "unity/builds/thor-{}-{}/Contents/MacOS/thor-local-OSXIntel64".format(
+        build_path = "unity/builds/thor-{}-{}/Contents/MacOS/AI2-Thor".format(
```
AlvaroHG, line 59: Does this break anything?

ai2thor/tests/data/metadata-schema.json
```diff
@@ -0,0 +1,666 @@
...
+        "objectIdsInBox",
+        "objects",
+        "reachablePositions",
+        "sceneBounds",
+        "sceneName",
+        "screenHeight",
+        "screenWidth",
+        "segmentedObjectIds",
+        "thirdPartyCameras",
+        "visibleRange"
```
AlvaroHG, line 664: nice!

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1382,3 +1383,28 @@ public class TypedVariable {
     public string type;
     public object value;
 }
+
+public class ShouldSerializeContractResolver : DefaultContractResolver
+{
+   public static readonly ShouldSerializeContractResolver Instance = new ShouldSerializeContractResolver();
+
+   protected override JsonProperty CreateProperty( MemberInfo member,
```
AlvaroHG, line 1391: Hopefully this does not make it much slower. Why did we not need this before, or did we? Is this to the net framework change and support with a particular Newtonsoft version?

ai2thor/tests/test_server.py
```diff
@@ -13,12 +13,10 @@ def generate_multi_agent_form(metadata, sequence_id=1):
     agent2['agentId'] = 1
     agent1 = metadata
     agents = [agent1, agent2]
-    action_returns = [None, None]
     boundary = b'--OVCo05I3SVXLPeTvCgJjHl1EOleL4u9TDx5raRVt'
     data = b'\r\n' + boundary + b'\r\nContent-Type: text/plain; charset="utf-8"\r\nContent-disposition: form-data; name="metadata"\r\n\r\n'
     data += json.dumps(dict(agents=agents, sequenceId=sequence_id, activeAgentId=1)).encode('utf8')
     data += b'\r\n' + boundary + b'\r\nContent-Type: text/plain; charset="utf-8"\r\nContent-disposition: form-data; name="actionReturns"\r\n\r\n'
-    data += json.dumps(action_returns).encode('utf8')
     data += b'\r\n' + boundary + b'\r\nContent-Type: text/plain; charset="utf-8"\r\nContent-disposition: form-data; name="token"\r\n\r\n'
```
AlvaroHG, line 20: do we still have an `actionReturn` value per agent? why is this different?

AlvaroHG:
A couple of questions, but looks as expected LGTM

tasks.py
ekolve: it was broken before - the binary name changed with the update to 2019

PR #442, opened 2020-08-12
Expert actions and RoboTHOR LookUp/LookDown fixes.

Lucaweihs:
Three main changes:

1. Brings ExpRoom up to date with the current master
2. Adds a new `ObjectNavExpertAction` to the stochastic controller.
3. Fixes LookUp/LookDown in RoboTHOR so that it doesn't let you look behind yourself.


Lucaweihs:
I agree but I'm not sure what the best way of dealing with this is as these types of actions should be computed at the unity level (it would be too slow to do it via repeated calls to the python API). Currently we don't, to the best of my knowledge, have a good place to put these types of "task specific" functions. Perhaps we should have a `extensions` or `tasks` file that provides experimental access to these types of functions? Any thoughts @ekolve ?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1866,6 +1866,29 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
             return Time.time;
         }
 
+        public bool objectIsWithinViewport(SimObjPhysics sop) {
```
ekolve, line 1869: Can this be made private or protected? Does it need to be public?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1866,6 +1866,29 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
...
+                    if (viewPoint.z > 0 &&
+                        viewPoint.x < ViewPointRangeHigh && viewPoint.x > ViewPointRangeLow && //within x bounds of viewport
+                        viewPoint.y < ViewPointRangeHigh && viewPoint.y > ViewPointRangeLow //within y bounds of viewport
+                    ) {
+                            return true;
+                    }
+                }
+            } else {
+                #if UNITY_EDITOR
+                Debug.Log("Error! Set at least 1 visibility point on SimObjPhysics prefab!");
```
ekolve, line 1886: Would it make sense to just log all the time even when not in the editor so that user's can see if an error occurred? Or is that not expected?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2544,23 +2567,141 @@ public void InitialRandomSpawn(ServerAction action)
...
+        public void ObjectNavExpertAction(ServerAction action) {
             SimObjPhysics sop = getSimObjectFromTypeOrId(action);
-            if (sop == null) {
+            var path = getShortestPath(sop, true);
+            if (path.status == UnityEngine.AI.NavMeshPathStatus.PathComplete) {
+
+                int parts = (int) Math.Round(360f / rotateStepDegrees);
+                if (Math.Abs((parts * 1.0f) - 360f / rotateStepDegrees) > 1e-5) {
+                    errorMessage = "Invalid rotate step degrees for agent, must divide 360 without a remainder.";
+                    actionFinished(false);
```
ekolve, line 2578: Add a return statement after the action finished so the method stops executing.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2544,23 +2567,141 @@ public void InitialRandomSpawn(ServerAction action)
...
+                        }
+                        m_Camera.transform.localEulerAngles = startCameraRot;
+                        transform.rotation = startRotation;
+                    }
+
+                    #if UNITY_EDITOR
+                    Debug.Log("Expert rotate and horizon:");
+                    Debug.Log(relRotate);
+                    Debug.Log(relHorizon);
+                    m_Camera.transform.localEulerAngles = new Vector3(startCameraRot.x + 30f * relHorizon, 0.0f, 0.0f);
```
ekolve, line 2617: Is it intentional to both log and transform the camera only when in Editor mode? Is this supposed to be outside the condition.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2544,23 +2567,141 @@ public void InitialRandomSpawn(ServerAction action)
...
+                
+                int numLeft = parts / 2;
+                int numRight = numLeft + (parts % 2 == 0 ? 1 : 0);
+                Vector3 startPosition = this.transform.position;
+                Quaternion startRotation = this.transform.rotation;
+                Vector3 startCameraRot = m_Camera.transform.localEulerAngles;
+
+                if (path.corners.Length <= 1) {
+                    if (objectIsWithinViewport(sop)) {
+                        actionFinished(true, null);
```
ekolve, line 2589: Should be able to change this to just actionFinished(true);

ekolve (changes requested):
Couple of comments and two small changes.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Made it protected.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Good point, removed the `#if UNITY_EDITOR` condition.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Good catch! Done.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Yes this is intentional (in the editor the agent takes a step in the direction of the expert action which is helpful for debugging, when running on the python side we will only want the name of the expert action, not to take it).

unity/Assets/Scripts/BaseFPSAgentController.cs

ekolve:
Merging. I am opening a separate issue to refactor the ObjectNavExpertAction into something more generic and less project specific.

PR #450, opened 2020-08-18
Erick/fifo server

ekolve:
This is a pretty big PR.  I add a new FIFO based server option (off by default) and use messagepack in place of json.net.  To provide some context, messagepack is about 30x faster than json.net (for our use cases) and 2x faster than Unity's JsonUtility (which doesn't always work).  The FIFO based server is roughly 10x - 20x faster than the flask based approach.  The new option to Controller() is server_type which controls whether this is enabled and we read an environment variable in the AgentManager which lets us know which backend has been requested.


ai2thor/server.py
```diff
@@ -403,10 +410,224 @@ class DepthFormat(Enum):
...
+    RGB_IMAGE = 4
+    DEPTH_IMAGE = 5
+    NORMALS_IMAGE = 6
+    FLOWS_IMAGE = 7
+    CLASSES_IMAGE = 8
+    IDS_IMAGE = 9
+    THIRD_PARTY_IMAGE = 10
+    END_OF_MESSAGE = 255
+
+class FifoServer(Server):
```
AlvaroHG, line 506: Should we move this subclass to a separate file?

ai2thor/server.py
```diff
@@ -403,10 +410,224 @@ class DepthFormat(Enum):
...
+            fifo_server_pipe_path=self.server_pipe_path,
+            fifo_client_pipe_path=self.client_pipe_path
+        )
+        return params
+
+    def stop(self):
+        self.client_pipe.close()
+        self.server_pipe.close()
+
+class WsgiServer(Server):
```
AlvaroHG, line 627: Same comment as above, have in a separate file

unity/Assets/Scripts/MessagePack/ExtensionHeader.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) All contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
AlvaroHG, line 2: Can we include a message pack distribution in manifest.json? instead of including all these source files .

ekolve: I believe the copyright header is a requirement from legal when we open sourced the project.

unity/Assets/Scripts/FifoServer.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        public void SendMessage(FieldType t, byte[] body) {
+            if (this.serverPipe == null) {
+                this.serverPipe = new NamedPipeClientStream(".", this.serverPipePath, PipeDirection.Out);
+                this.serverPipe.Connect();
+            }
+            //Console.WriteLine("server pipe + connected " + this.serverPipe.IsConnected );
+            byte[] header = new byte[headerLength];
+            header[0] = (byte)t;
+            PackNetworkBytes(body.Length).CopyTo(header, 1);
```
AlvaroHG, line 86: Why is the header copied to every location of the buffer starting at position `1` of the byte array?

ekolve: Each field is comprised of a HEADER + BODY.  The HEADER consists of a FIELD_TYPE + FIELD_LENGTH. The size of the HEADER is always 5 bytes (1 for the type and 4 (32 bits) for the length). So that is why we always have to read the header this way.

unity/Assets/Scripts/FifoServer.cs
```diff
@@ -0,0 +1,126 @@
...
+            serverPipe.Write(header, 0, header.Length);
+            serverPipe.Write(body, 0, body.Length);
```
AlvaroHG, line 88: Does making 2 `Write` calls vs one change anything? the server is still waiting at this point right, so it's just for clarity?

unity/Assets/Scripts/FifoServer.cs
AlvaroHG: Reading ReadMessage shouldn't this line be? 

`PackNetworkBytes(body.Length).CopyTo(body.Lenght, 1);`

or `IPAddress.HostToNetworkOrder(body.Lengh)` for correct endianness 

Like if I understand you are using the first byte to store the header `FieldType` and the rest of the bytes to store the integer representing the size of the body to be read.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -643,41 +651,20 @@ private void addImageSynthesisImageForm(WWWForm form, ImageSynthesis synth, bool
 		if (shouldRender) {
 			// we should only read the screen buffer after rendering is complete
 			yield return new WaitForEndOfFrame();
-			if (synchronousHttp) {
-				// must wait an additional frame when in synchronous mode otherwise the frame lags
-				yield return new WaitForEndOfFrame();
-			}
+			yield return new WaitForEndOfFrame();
```
AlvaroHG, line 654: Can you add the comment back to have reference of why we are waiting twice?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -723,28 +706,61 @@ private void addImageSynthesisImageForm(WWWForm form, ImageSynthesis synth, bool
...
+			// we should only read the screen buffer after rendering is complete
+			yield return new WaitForEndOfFrame();
+		}
+
+        MultiAgentMetadata multiMeta = new MultiAgentMetadata ();
+		
+
+		ThirdPartyCameraMetadata[] cameraMetadata = new ThirdPartyCameraMetadata[this.thirdPartyCameras.Count];
+        List<KeyValuePair<string, byte[]>> renderPayload = new List<KeyValuePair<string, byte[]>>();
+        createPayload(multiMeta, cameraMetadata, renderPayload, shouldRender);
```
AlvaroHG, line 741: this is a lot cleaner

unity/Assets/Plugins/System.Buffers.dll.meta
```diff
@@ -0,0 +1,33 @@
+fileFormatVersion: 2
```
AlvaroHG, line 1: What are these `dll`'s for? is this a Windows only thing?

ai2thor/controller.py
```diff
@@ -461,6 +455,29 @@ def __init__(
                     event.metadata['errorMessage'])
                 )
 
+    def _build_server(self, host, port, width, height):
+
+        if self.server is not None:
+            return
+
+        if self.server_type == 'wsgi':
```
AlvaroHG, line 463: Could we make these an `enum` or constants of a class?

unity/Assets/Scripts/FifoServer.cs
ekolve: 1 write call vs 2 really showed no difference - you would also have to create a new byte[] array and copy the data from each into the new one in order to get the single write call, so you lose any time you save doing that.  As well, serverPipe buffers the write calls internally and flushes, so it ended up not making any difference.

unity/Assets/Plugins/System.Buffers.dll.meta
ekolve: Those are came in from the MessagePack unity package: https://github.com/neuecc/MessagePack-CSharp/releases/download/v2.1.165/MessagePack.Unity.2.1.165.unitypackage

ai2thor/tests/fifo_client.py
```diff
@@ -0,0 +1,52 @@
...
+        header = struct.pack(ai2thor.fifo_server.FifoServer.header_format, field_type, len(body))
+
+        self.server_pipe.write(header + body)
+
+    def send_eom(self):
+        header = struct.pack(ai2thor.fifo_server.FifoServer.header_format, ai2thor.fifo_server.FieldType.END_OF_MESSAGE, 0)
+        self.server_pipe.write(header)
+        self.server_pipe.flush()
+
+    def recv(self):
```
AlvaroHG, line 26: Feels like the send and receive functionality is almost the same between the Client and Server, the client is just for testing right?

ai2thor/tests/test_unity.py
```diff
@@ -64,7 +61,8 @@ def test_small_aspect():
     event = controller.step(dict(action='Initialize', gridSize=0.25))
     assert event.frame.shape == (64, 128, 3)
     
-def test_lookdown():
+@pytest.mark.parametrize("controller", [wsgi_controller, fifo_controller])
```
AlvaroHG, line 64: nice

AlvaroHG:
A couple of comments and small suggestions. It looks really good, LGTM!

unity/Assets/Scripts/MessagePack/ExtensionHeader.cs
AlvaroHG: I meant including the binary instead of the source, discussed through slack and resolved

unity/Assets/Scripts/FifoServer.cs
AlvaroHG: makes sense

unity/Assets/Plugins/System.Buffers.dll.meta
AlvaroHG: ok

ai2thor/tests/fifo_client.py
ekolve: Yeah, its just for testing. If we needed for anything else I would make it part of the main package, but I kind of like having it separate since it essentially represents an external implementation, similar to the csharp one and will help a bit to find potential bugs if we change the server but neglect to change the client.

ai2thor/tests/test_unity.py
ekolve: Yeah, I just discovered that parametrize for this.  Its pretty nice.

ai2thor/controller.py
ekolve: I modified the field to 'server_class' and you pass in the class which will be WsgiServer or FifoServer and we default to WsgiServer

unity/Assets/Scripts/AgentManager.cs
ekolve: Done

PR #456, opened 2020-08-26
adding ActionDispatcher which allows actions to be defined without us…

ekolve:
…ing ServerAction

Please take a look at the comment at the top of ActionDispatcher.cs for more details.

### DIFF
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -231,27 +231,42 @@ def test_action_dispatch_invalid_action():
     assert controller.last_event.metadata['errorCode'] == 'InvalidAction'
 
 
-def test_action_dispatch():
+def test_action_dispatch_empty():
     event = controller.step(dict(action='TestActionDispatchNoop'))
     assert event.metadata['actionReturn'] == 'emptyargs'
+
+def test_action_disptatch_one_param():
     event = controller.step(dict(action='TestActionDispatchNoop', param1=True))
-    print(event.metadata['errorMessage'])
     assert event.metadata['actionReturn'] == 'param1'
+
+def test_action_disptatch_two_param():
     event = controller.step(dict(action='TestActionDispatchNoop', param1=True, param2=False))
     assert event.metadata['actionReturn'] == 'param1 param2'
+
+def test_action_disptatch_two_param_with_default():
     event = controller.step(dict(action='TestActionDispatchNoop', param3=True, param4='foobar'))
     assert event.metadata['actionReturn'] == 'param3 param4/default foobar'
+
+
+def test_action_disptatch_two_param_with_default_empty():
     event = controller.step(dict(action='TestActionDispatchNoop', param3=True))
     assert event.metadata['actionReturn'] == 'param3 param4/default foo'
+
+def test_action_disptatch_serveraction_default():
     event = controller.step(dict(action='TestActionDispatchNoopServerAction'))
     assert event.metadata['actionReturn'] == 'serveraction'
+
+
+def test_action_disptatch_serveraction_with_object_id():
     event = controller.step(dict(action='TestActionDispatchNoopServerAction', objectId='candle|1|2|3'))
     assert event.metadata['actionReturn'] == 'serveraction'
+
+def test_action_disptatch_all_default():
     event = controller.step(dict(action='TestActionDispatchNoopAllDefault'))
     assert event.metadata['actionReturn'] == 'alldefault'
 
+def test_action_disptatch_some_default():
     event = controller.step(dict(action='TestActionDispatchNoopAllDefault', param12=9.0))
-    print(event.metadata['errorMessage'])
     assert event.metadata['actionReturn'] == 'somedefault'
 
 def test_moveahead():
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -211,6 +211,26 @@ def test_teleport():
     controller.step(dict(action='Teleport', x=-2.0, z=-2.5, y=1.0), raise_for_failure=True)
     position = controller.last_event.metadata['agent']['position']
     assert_near(position, dict(x=-2.0, z=-2.5, y=0.901))
+
+def test_action_dispatch_find_conflicts_stochastic():
+    event = controller.step(dict(action='TestActionDispatchFindConflicts'), typeName='UnityStandardAssets.Characters.FirstPerson.StochasticRemoteFPSAgentController')
+    known_conflicts = {
+        'GetComponent': ['type'],
+        'StopCoroutine': ['routine'],
+        'TestActionDispatchConflict': ['param22']
+    }
+    assert event.metadata['actionReturn'] == known_conflicts
+
+def test_action_dispatch_find_conflicts_physics():
+    event = controller.step(dict(action='TestActionDispatchFindConflicts'), typeName='UnityStandardAssets.Characters.FirstPerson.PhysicsRemoteFPSAgentController')
+    known_conflicts = {
+        'GetComponent': ['type'],
+        'StopCoroutine': ['routine'],
+        'TestActionDispatchConflict': ['param22']
+    }
+    assert event.metadata['actionReturn'] == known_conflicts
+
+
 def test_action_dispatch_missing_args():
     caught_exception = False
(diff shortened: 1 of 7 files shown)


unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -0,0 +1,233 @@
...
+                    foreach(var p in method.GetParameters()) {
+                        if (actionParams.Contains(p.Name)) {
+                            matchCount++;
+                        }
+                    }
+                }
+
+                // preference is given to the method that matches all parameters for a method
+                // even if another method has the same matchCount (but has more parameters)
+                if (matchCount > bestMatchCount) {
```
AlvaroHG, line 139: cool!

unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -0,0 +1,233 @@
...
+            System.Reflection.MethodInfo[] allMethods = target.GetType().GetMethods(BindingFlags.Public | BindingFlags.Instance);
+            List<Type> hierarchy = new List<Type>();
+            // not completely generic
+            Type ht = t;
+            while (ht != typeof(object)) {
+                hierarchy.Add(ht);
+                ht = ht.BaseType;
+            }
+
+            Dictionary<string, List<MethodInfo>> methodDispatch = new Dictionary<string, List<MethodInfo>>();
```
AlvaroHG, line 59: We might be able to cache this after an `Init` call when the runtime type of the agent changes. Just an optimization thought for the future, but we should do this first.

unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -0,0 +1,233 @@
...
+            result = paramCount(a).CompareTo(paramCount(b));
+        }
+        return result;
+    }
+
+    private int paramCount(MethodInfo method) {
+        return method.GetParameters().Length;
+    }
+
+    private int requiredParamCount(MethodInfo method) {
```
AlvaroHG, line 210: Make static

unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -0,0 +1,233 @@
...
+        int requiredParamCountA = requiredParamCount(a);
+        int requiredParamCountB = requiredParamCount(b);
+        int result = requiredParamCountA.CompareTo(requiredParamCountB);
+        if (result == 0) {
+            result = paramCount(a).CompareTo(paramCount(b));
+        }
+        return result;
+    }
+
+    private int paramCount(MethodInfo method) {
```
AlvaroHG, line 206: Make static

unity/Assets/Scripts/ActionDispatcher.cs
ekolve: Can you elaborate on this? I'm not sure how I would make this method static since its part of the comparer.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1469,49 +1469,58 @@ public virtual MetadataWrapper generateMetadataWrapper()
...
+            this.actionComplete = false;
+
+            try
+            {
+                ActionDispatcher.Dispatch(this, controlCommand);
+            }
+            catch (MissingArgumentsActionException e)
+            {
+                errorMessage = "action: " + controlCommand.action + " is missing the following arguments: " + string.Join(",", e.ArgumentNames.ToArray());
+                errorCode = ServerActionErrorCode.MissingArguments;
```
AlvaroHG, line 1493: this is great

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2972,5 +2981,36 @@ void OnDrawGizmos()
...
+        public void TestActionDispatchNoop(bool param1, bool param2) {
+            actionFinished(true, "param1 param2");
+        }
+
+        public void TestActionDispatchNoop(bool param1) {
+            actionFinished(true, "param1");
+        }
+        public void TestActionDispatchNoop() {
+            actionFinished(true, "emptyargs");
+        }
```
AlvaroHG, line 3014: Maybe we should consider also having a test agent? just this class is becoming really big

ai2thor/tests/test_unity.py
```diff
@@ -211,6 +211,48 @@ def test_teleport():
...
+    event = controller.step(dict(action='TestActionDispatchNoopServerAction'))
+    assert event.metadata['actionReturn'] == 'serveraction'
+    event = controller.step(dict(action='TestActionDispatchNoopServerAction', objectId='candle|1|2|3'))
+    assert event.metadata['actionReturn'] == 'serveraction'
+    event = controller.step(dict(action='TestActionDispatchNoopAllDefault'))
+    assert event.metadata['actionReturn'] == 'alldefault'
+
+    event = controller.step(dict(action='TestActionDispatchNoopAllDefault', param12=9.0))
+    print(event.metadata['errorMessage'])
+    assert event.metadata['actionReturn'] == 'somedefault'
```
AlvaroHG, line 255: I feel like in general each test case should be a different function. If you run your tests and this big test fails at the beginning it may hide other failures, and pytest reports the amount of passed tests where the test boundary is the function . Like getting `6 out of 9 passed` vs `0 out of 1 failed` for the above test

AlvaroHG:
A couple of small suggestions. LGTM after that!

ai2thor/tests/test_unity.py
ekolve: Good point, i'll break them out

unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -0,0 +1,233 @@
...
+    Creating the following overloaded set of functions will not work as expected:
+
+    public void Teleport(int x, int y)
+    public void Teleport(int x, short y)
+
+    as well the following scenario should also be avoided:
+
+    public void ObjectVisible(bool foo, int x, int y)
+    public void ObjectVisible(int x, int y, bool foo)
+
```
AlvaroHG, line 31: What happens in this case? the sort function would just rank them the same way right? and choose one. maybe would be good to be able to distinguish this case from the correct override, by checking `==` case in the ranking function and propagate an error. What do you think?

unity/Assets/Scripts/ActionDispatcher.cs
AlvaroHG: Because it does not use any internal state of the `MethodParamComparer` instance, it is like a library function that just takes a `method` variable and returns a value, independent of the instance. It's a general practice to prefer static methods due to object oriented opacity, and makes it clear that the method is like a utility and not really doing anything with the internal state of the instance.

AlvaroHG: A minor point really, just easy to make static

unity/Assets/Scripts/ActionDispatcher.cs
ekolve: Yeah, this wouldn't fail. The getDispatchMethod would add both methods to the dispatchTable since the signatureMatch would be false between the two methods (we compare the parameters in order).  They would both end up with the same matchCount so we would just pick the first one.  I think for this case I don't want the error happening at runtime since that isn't real great for a user.  I am going to add a separate test that looks for these signature collisions and execute with the unit tests.  So if a collision is created, the build will fail and the committer will have to resolve the conflict?  Sound good?

AlvaroHG: Sounds good!

ekolve: Okay, added a unit test to look for these types of conflicts.  If we need/want to have the aforementioned method overloading the dispatcher can be made more type aware and attempt to map the JSON types to csharp types to reduce the conflicts.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: I agree.  I am going to create a separate issue that addresses Agent refactoring with the goal of making all the agentType, subclassing a little better.  I'm not sure what approach we should take, but I think we need to make things a little more extensible.

PR #457, opened 2020-08-28
Navmeshes for iTHOR scenes

AlvaroHG:
Updates Automatic navmesh Generation scripts, ran script to tag objects walkable, not walkable and built navmeshes for iTHOR tall agent.

Enables spl and shortest path functions in iTHOR


AlvaroHG:
There is not super easy way to test every scene, and it is sort of testing the layout to do a shortest path function or an spl that I would compute with the same spl function in the first place. These navmesh tests should use the new test room.
I started a document to list what we would test in the test room and as a design doc for it: https://docs.google.com/document/d/16QxkVge0Umut4-HsiMqfn23m5yCfovrOmc89il-37PQ/edit?usp=sharing

PR #460, opened 2020-08-31
add the visualization for 3D bounding box

KuoHaoZeng:
Add a visualization script for 3D bounding box on an 2D image.
The script is located in `ai2thor/util/visualize_3D_bbox.py`
The main function is `draw_3d_bbox(event)`, which takes an `event` object and draw all visible objects' 3D bounding boxes and store it back to the `event` by `event.bbox_3d_frame`.
This function would return the `event` object containing the `bbox_3d_frame`.


ai2thor/util/visualize_3D_bbox.py
```diff
@@ -0,0 +1,248 @@
...
+
+                # draw the 3D bbox 12 lines in the 2D RGB image
+                for iii, line in enumerate(lines):
+                    if valid_lines[iii]:
+                        draw.line((line[0], line[1], line[2], line[3]), fill=color, width=2)
+
+    # store the result back to the event
+    bbox_frame = np.array(img)
+    event.bbox_3d_frame = bbox_frame
+    return event
```
ekolve, line 211: Why is the event returned here? As well, modifying the event is not something a user would expect.  I think it would make more sense to return the bbox_frame.

ekolve: Another option, might be to move this code to a property named bbox_3d_frame off of the Event class in server.py.  Since the only dependency it has is the event it might make sense to move it there.

KuoHaoZeng: Having a property named `bbox_3d_frame` in `server.py` sounds good to me!
I guess we can still keep the source script in `util/visualize_3D_bbox.py` and make a function call in `server.py`.
It might make more sense if we have more visualization tools in the future?

Lucaweihs:
Closing this for now, but let me know if you want this functionality still @KuoHaoZeng

PR #464, opened 2020-09-11
Receptacle snapshot

winthos:
There are two main changes with this pull:

1) Receptacle Behavior - receptacles used to update their contents with each fixed update using a series of OnTrigger calls. This has now been replaced with Physics Overlap Boxes that snapshot the contents of the receptacle whenever an actionFinished() call is made. This should improve physics performance somewhat (hopefully a lot) due to minimizing the number of fixed update checks are happening across all receptacle objects.

2) New (undocumented for now?) Initialization parameter to allow more control over how long actions yield to return when waiting for an object to come to rest- Previously all actions that could move an object in a way that it should wait for the object to come to rest before returning actionFinished() defaulted to 2 second wait window. This is good for efficiency but could cause inaccuracies because some objects won't stop moving after only 2 seconds, leading to an actionFinished() call that is not synced up with how the world state actually is. 

**Example of Current Wait for Rest Behavior:** drop apple, wait 2 seconds, apple is not finished moving until 10 seconds. The result of this will be the actionFinished() fires at 2 seconds, giving a snapshot of the state of the apple and scene at t = 2, but once t = 10 the apple will be in an entirely new location not reflected in the metadata.

This change allows users to specify how long they want to yield for physics resolution via the `TimeToWaitForObjectsToComeToRest` float that can be specified on init. An additional change with this, is for more consistency, when the wait time is up all objects will be forced to stop moving and immediately go to rigidbody sleep, halting movement. This means that a new behavior will happen given the example above:

**Example of New Wait for Rest Behavior:** drop apple, wait 2 seconds, apple normally would not finish moving until 10 seconds. After 2 seconds, actionFinished() is called but the apple becomes frozen wherever it was at t = 2. This will make the metadata reflect the scene state without being desynced, but may cause situations like a bouncing ball after 2 seconds being frozen in midair if it happened to be midair at t = 2. To counteract this, users will be able to set the wait time to longer, like 20 seconds, to ensure the ball finishes bouncing. I think this default behavior, while less realistic, is more desirable than the current behavior where the object just becomes desynced so that the last returned metadata is just not accurate at all. Notably, if an object would come to rest sooner than the specified wait time, the action will short circuit and immediately return without having to wait the entire wait time.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -1060,7 +1060,14 @@ namespace UnityStandardAssets.Characters.FirstPerson
                 }
                 ObjectMetadata meta = ObjectMetadataFromSimObjPhysics(simObj, visibleSimObjsHash.Contains(simObj));
                 if (meta.receptacle) {
-                    List<string> roid = simObj.Contains();
+                    
+                    List<string> containedObjectsAsID = new List<String>();
+                    foreach(GameObject go in simObj.ContainedGameObjects())
+                    {
+                        containedObjectsAsID.Add(go.GetComponent<SimObjPhysics>().ObjectID);
+                    }
+                    List<string> roid = containedObjectsAsID;//simObj.Contains();
+
                     foreach (string oid in roid) {
                         if (!parentReceptacles.ContainsKey(oid)) {
                             parentReceptacles[oid] = new List<string>();
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -54,6 +54,8 @@ namespace UnityStandardAssets.Characters.FirstPerson
 		[SerializeField]
 		protected float m_GravityMultiplier;
 		protected static float gridSize = 0.25f;
+        //time the checkIfObjectHasStoppedMoving coroutine waits for objects to stop moving
+        protected float TimeToWaitForObjectsToComeToRest = 0.0f;
         //determins default move distance for move actions
 		protected float moveMagnitude;
         //determines rotation increment of rotate functions
@@ -421,7 +423,7 @@ namespace UnityStandardAssets.Characters.FirstPerson
                 	Time.timeScale = action.timeScale;
 				}
             } else {
-                errorMessage = "Time scale must be >0";
+                errorMessage = "Time scale must be > 0";
                 Debug.Log(errorMessage);
                 actionFinished(false);
                 return;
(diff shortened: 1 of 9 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -467,11 +469,15 @@ public void Initialize(ServerAction action)
...
             }
+
             else
             {
                 gridSize = action.gridSize;
                 StartCoroutine(checkInitializeAgentLocationAction());
             }
+
+            //initialize how long the default wait time for objects to stop moving is
+            this.TimeToWaitForObjectsToComeToRest = action.TimeToWaitForObjectsToComeToRest;
```
AlvaroHG, line 480: To auto detect if an object has been moving I thought we had mentioned using it's velocity as the derivative of it's position checking if the difference between the positions between two or more frames was zero. Is this used somewhere else? do we want this extra parameter for wait-time or is it that to get that non-movement it would take too long?

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -2018,11 +2020,14 @@ private IEnumerator checkIfObjectHasStoppedMoving(SimObjPhysics sop, float lengt
 
             if(sop != null)
             {
+                // print("here?");
+                // print("useTimeout?: " + useTimeout);
```
AlvaroHG, line 2024: remove comments

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -2032,12 +2037,12 @@ private IEnumerator checkIfObjectHasStoppedMoving(SimObjPhysics sop, float lengt
                     //ok the accel is basically zero, so it has stopped moving
                     if(Mathf.Abs(accel) <= 0.001f)
                     {
-                        stoppedMoving = true;
-
+                        // print("accel world");
```
AlvaroHG, line 2040: remove comments

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: it does both. What will happen is the action will return depending on which happens first:
1) action times out because it waited the entire duration of TimeToWait
2) object comes to rest via velocity check, and the action returns regardless of time left

I got a request to limit this to some default time because the default behavior of only waiting until all objects come to rest could result in cases where an action must delay 30+ seconds. This new behavior will still allow this, but it will require users to set the TimeToWait to somethine like a minute themselves.

AlvaroHG:
LGTM

unity/Assets/Scripts/BaseFPSAgentController.cs
AlvaroHG: cool

PR #466, opened 2020-09-15
Erick/simobj filter

ekolve:
Allow users to filter all or some of the outbound objects in the metadata.  This has shown to result in a dramatic performance increase.  Using the FifoServer (new backend) and filtering all objects we go from 82 FPS to 172 FPS (Pass action).  Using WSGI we go from 22 FPS to 27 FPS (this is due to the poorer performance of the existing WSGI backend and JSON.net).  Users will want to filter all objects when training a point nav task and can filter down to one or two objects for object nav tasks.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -70,6 +70,8 @@ namespace UnityStandardAssets.Characters.FirstPerson
         public GameObject DroneBasket;//reference to the drone's basket object
         private bool isVisible = true;
         public bool inHighFrictionArea = false;
+        // outbound object filter
+        private HashSet<SimObjPhysics> simObjFilter = null;
 
         public bool IsVisible
         {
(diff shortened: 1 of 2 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1073,16 +1075,39 @@ protected bool IsCollided()
...
 
+        public void ResetObjectFilter() {
+            this.simObjFilter = null;
+            actionFinished(true);
+        }
+        public void SetObjectFilter(string[] objectIds) {
+            SimObjPhysics[] simObjects = GameObject.FindObjectsOfType<SimObjPhysics>();
+            simObjFilter = new HashSet<SimObjPhysics>();
+            foreach(var simObj in simObjects) {
+                if(Array.IndexOf(objectIds, simObj.ObjectID) >= 0) {
```
AlvaroHG, line 1086: having a dictionary of the simobjects by id could be good so you don't have to look in the `objectIds` array every time, the array is probably often small so no big deal.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1073,16 +1075,39 @@ protected bool IsCollided()
...
+                    if (isSimObjVisible(m_Camera, sop)) {
+                        visibleSimObjsHash.Add(sop);
+                    }
+                }
+                simObjects = this.simObjFilter.ToArray();
+            } else {
+                foreach (SimObjPhysics sop in VisibleSimObjs(false)) {
+                    visibleSimObjsHash.Add(sop);
+                }
+                simObjects = GameObject.FindObjectsOfType<SimObjPhysics>();
```
AlvaroHG, line 1108: At line 1122, below (GitHub does not let me comment there), there is a reference to `excludeObject` which uses an array of ids  `this.excludeObjectIds` which probably we can deprecate and I don't see any other references to.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: A hashset seems like the right structure here and this array lookup is done so rarely (only on setting the filter
on each request to retrieve the actual object.  Doing this Array.IndexOf() is only done when you set the filter

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1073,16 +1075,39 @@ protected bool IsCollided()
...
         public virtual ObjectMetadata[] generateObjectMetadata()
 		{
-            SimObjPhysics[] visibleSimObjs = VisibleSimObjs(false); // Update visibility for all sim objects for this agent
             HashSet<SimObjPhysics> visibleSimObjsHash = new HashSet<SimObjPhysics>();
-            foreach (SimObjPhysics sop in visibleSimObjs) {
-                visibleSimObjsHash.Add(sop);
+            SimObjPhysics[] simObjects = null;
+            if (this.simObjFilter != null) {
+                foreach (SimObjPhysics sop in this.simObjFilter) {
+                    if (isSimObjVisible(m_Camera, sop)) {
```
AlvaroHG, line 1099: Since you added a function that works with a single sim object, you probably don't need the `visibleSimObjsHash` in line 1125 you can replace `visibleSimObjsHash.Contains(simObj)` with a call to this new function `isSimObjVisible(m_Camera, sop)`

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1979,47 +2043,15 @@ protected SimObjPhysics[] GetAllVisibleSimObjPhysics(Camera agentCamera, float m
                     if (sop != null && !testedSops.Contains(sop)) 
                     {
                         testedSops.Add(sop);
-                        //check against all visibility points, accumulate count. If at least one point is visible, set object to visible
-                        if (sop.VisibilityPoints == null || sop.VisibilityPoints.Length > 0) 
+                        if (isSimObjVisible(agentCamera, sop) && !currentlyVisibleItems.Contains(sop))
```
AlvaroHG, line 2046: `currentlyVisibleItems` could be a HashSet for performance

AlvaroHG:
This is great, improvement in clarity as well, left a few comments some minimal optimizations and deprecation of a similar unused system. LGTM!

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: If you look at the GetAllVisibleSimObjPhysics (which VisibleSimObjs calls), there is some logic there that does a spherecast first and then does a visibility check on the actual object.  This is done to avoid having to send raycasts out to every visibility point of every simobject on every frame.

AlvaroHG: Makes sense

AlvaroHG:
LGTM

PR #467, opened 2020-09-16
removing useAgentTransform and converting GetShortestPath and GetShor…

ekolve:
…testPathToPoint to new actiondispatch style

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -2537,16 +2537,15 @@ namespace UnityStandardAssets.Characters.FirstPerson
             }
         }
 
-        private SimObjPhysics getSimObjectFromTypeOrId(ServerAction action) {
-            var objectId = action.objectId;
-            if (!String.IsNullOrEmpty(action.objectType) && String.IsNullOrEmpty(action.objectId)) {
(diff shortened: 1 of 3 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2598,6 +2590,18 @@ public void InitialRandomSpawn(ServerAction action)
...
 
+        public void GetShortestPath(Vector3 position, Vector3 rotation, string objectType = null, string objectId = null) {
+            getShortestPath(objectType, objectId, position, Quaternion.Euler(rotation));
+        }
+
+        public void GetShortestPath(Vector3 position, string objectType = null, string objectId = null) {
+            getShortestPath(objectType, objectId, position, Quaternion.Euler(Vector3.zero));
+        }
+
+        public void GetShortestPath(string objectType = null, string objectId = null) {
```
AlvaroHG, line 2601: Makes sense

AlvaroHG:
LGTM!

A couple of things to note:
We have to update the robothor dataset generation in tasks.py, this breaks compatibility, I'm ok with it.
There is also this function which may be used for similar purposes `ObjectTypeToObjectIds` which get's an object type (currently still a ServerAction) like "apple" and returns the id's on the scene of that object

PR #468, opened 2020-09-16
removing removeProb, objectName, objectRandom, and intensity from Ser…

ekolve:
…verAction


AlvaroHG:
LGTM, messaged you with some problems of calling actions dynamically that take a Vector3 in their params, so maybe we should merge this after we figure that stuff out?

PR #474, opened 2020-09-21
adding distance restriction on simobject filter visibility check

ekolve:


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2100,35 +2107,9 @@ protected SimObjPhysics[] GetAllVisibleSimObjPhysics(Camera agentCamera, float m
...
-                                    if (CheckIfVisibilityPointInViewport(sop, point, agentCamera, true)) 
-                                    {
-                                        visPointCount++;
-                                    }
-                                }
-
-                                //if we see at least one vis point, the object is "visible"
-                                if (visPointCount > 0) 
-                                {
-                                    #if UNITY_EDITOR
```
winthos, line 2123: I would add this editor define back so that debug highlighting of objects still works while in editor mode.

winthos: actually wait, I believe this is fine because this was all effectively moved to the isSimObjVisible call

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1950,18 +1950,25 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
...
                 int visPointCount = 0;
 
+
                 foreach (Transform point in visPoints) 
                 {
+                    if (maxDistance > 0.0f) {
+                        Vector3 agentPoint = new Vector3(this.transform.position.x, point.position.y, this.transform.position.z);
+                        if (Vector3.Distance(agentPoint, point.position) > maxDistance) {
+                            continue;
+                        }
```
Lucaweihs, line 1968: The type of distance we use for visibility checks actually isn't "sphere shaped" as it, more or less, ignores y positions. This is important so that objects on the ground are considered interactable. I worry that using the euclidean distance here might thus incorrectly throw away some visibility points.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1092,7 +1092,7 @@ public virtual ObjectMetadata[] generateObjectMetadata()
             SimObjPhysics[] simObjects = null;
             if (this.simObjFilter != null) {
                 foreach (SimObjPhysics sop in this.simObjFilter) {
-                    if (isSimObjVisible(m_Camera, sop)) {
+                    if (isSimObjVisible(m_Camera, sop, false, this.maxVisibleDistance)) {
```
Lucaweihs, line 1095: This sort of fixes the problem but not really as the `VisibleSimObjs` can result in an object being invisible using the filtered version and visible with the non-filtered version. I vote for adding a `simObjFilter` parameter to `VisibleSimObjs` and then having the `VisibleSimObjs` logic skip anything not in this filter. At the cost of some computation we get guaranteed consistency.

Lucaweihs (commented):
Some comments.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1950,18 +1950,25 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
             return Time.time;
         }
 
-        private bool isSimObjVisible(Camera agentCamera, SimObjPhysics sop) {
+        private bool isSimObjVisible(Camera agentCamera, SimObjPhysics sop, bool includeInvisible, float maxDistance = 0.0f) {
```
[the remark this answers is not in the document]
ekolve: I refactored to always require a distance - I feel like this makes more sense since that way the isSimObjVisible can reflect the definition of what we consider visible.  I could argue for just removing the maxDistance parameters that are passed all around in the class and simply use this.maxVisibleDistance so that this method would always return the correct value.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1950,18 +1950,25 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
...
             bool visible = false;
             //check against all visibility points, accumulate count. If at least one point is visible, set object to visible
-            if (sop.VisibilityPoints == null || sop.VisibilityPoints.Length > 0) 
+            if (sop.VisibilityPoints != null && sop.VisibilityPoints.Length > 0) 
             {
                 Transform[] visPoints = sop.VisibilityPoints;
                 int visPointCount = 0;
 
+
```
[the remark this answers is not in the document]
ekolve: fixed.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1950,18 +1950,25 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
...
                 foreach (Transform point in visPoints) 
                 {
+                    if (maxDistance > 0.0f) {
+                        Vector3 agentPoint = new Vector3(this.transform.position.x, point.position.y, this.transform.position.z);
```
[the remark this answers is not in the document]
ekolve: We should have chat with Winson, but doing the calculation this way makes it consistent with how the capsule cast limited the distance of what is visible the distance.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: See the change above where we fix the y value of the agentPosition to the visibility point's y position.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: I'm not clear how that could occur - the only difference between VisibleSimObjs and isSimObjVisible is the former performs a capsule cast to filter down the list of candidates that we then send to isSimObjVisible with the same distance value as the filtered list.

unity/Assets/Scripts/BaseFPSAgentController.cs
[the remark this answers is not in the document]
Lucaweihs: I misread this, you're correct.

PR #475, opened 2020-09-21
disabling unity audio

ekolve:
I noticed when running a large number of experiments that the pulseaudio process on Linux would use up to 100% of a CPU.  This PR disables the Audio in thor. When I ran a build with the audio disabled pulseaudio uses no CPU.


AlvaroHG:
LGTM

PR #478, opened 2020-09-23
Erick/simobj filter distance 2

ekolve:
Second try:

I rewrote the VisibleSimObs method to only examine the VisibilityPoints instead of using CapsuleColliders.  The benefit is around a 2x-10x speedup (when not using the filter) depending on the scene and where the agent is located. However, I am only using this approach when a filter is set, so the speedup is currently not realized by default.  My plan is to switch the existing visibility check to this one after I can do an exhaustive test of all scenes + points, comparing the old and new lists of objects that are reported as visible.

### DIFF
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -1355,6 +1355,7 @@ public class ServerAction
 	public float TimeToWaitForObjectsToComeToRest = 10.0f;
 	public float intensity;//used for light?
 	public float scale;
+    public string visibilityScheme;
 
     public SimObjType ReceptableSimObjType()
 	{
@@ -1364,6 +1365,13 @@ public class ServerAction
 		}
 		return (SimObjType)Enum.Parse(typeof(SimObjType), receptacleObjectType);
 	}
+
+    public VisibilityScheme GetVisibilityScheme() {
+        if (string.IsNullOrEmpty(visibilityScheme)) {
+            return VisibilityScheme.Collider;
+        }
+		return (VisibilityScheme)Enum.Parse(typeof(VisibilityScheme), visibilityScheme);
+    }
 
 	public SimObjType GetSimObjType()
 	{
@@ -1409,6 +1417,12 @@ public enum ServerActionErrorCode  {
 	InvalidAction,
     MissingArguments
 }
+
+public enum VisibilityScheme {
+    Collider,
+    Distance
+}
+
 
 
 [Serializable]
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -484,6 +484,7 @@ public class AgentManager : MonoBehaviour
     }
 
     private void LateUpdate() {
+
 		int completeCount = 0;
 		foreach (BaseFPSAgentController agent in this.agents) {
 			if (agent.actionComplete) {
@@ -1355,7 +1356,7 @@ public class ServerAction
 	public float TimeToWaitForObjectsToComeToRest = 10.0f;
 	public float intensity;//used for light?
 	public float scale;
-    public string visibilityScheme;
+    public string visibilityScheme = VisibilityScheme.Collider.ToString();
 
     public SimObjType ReceptableSimObjType()
 	{
@@ -1367,10 +1368,16 @@ public class ServerAction
 	}
 
     public VisibilityScheme GetVisibilityScheme() {
-        if (string.IsNullOrEmpty(visibilityScheme)) {
-            return VisibilityScheme.Collider;
+        VisibilityScheme result = VisibilityScheme.Collider;
+        try 
+        {
+            result = (VisibilityScheme)Enum.Parse(typeof(VisibilityScheme), visibilityScheme, true);
+        } 
+        catch (ArgumentException ex) { 
+            Debug.LogError("Error parsing visibilityScheme: '" + visibilityScheme + "' defaulting to Collider");
         }
-		return (VisibilityScheme)Enum.Parse(typeof(VisibilityScheme), visibilityScheme);
+
+		return result;
     }
 
 	public SimObjType GetSimObjType()
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -73,6 +73,7 @@ namespace UnityStandardAssets.Characters.FirstPerson
         public bool inHighFrictionArea = false;
         // outbound object filter
         private SimObjPhysics[] simObjFilter = null;
+        private VisibilityScheme visibilityScheme = VisibilityScheme.Collider;
 
         public bool IsVisible
         {
@@ -492,6 +493,8 @@ namespace UnityStandardAssets.Characters.FirstPerson
                 }
 
             }
+
+            this.visibilityScheme = action.GetVisibilityScheme();
         }
 
         public void SetAgentMode(string mode)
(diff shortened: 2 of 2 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1965,18 +1962,42 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
...
+                    if (xdelta > maxDistance) 
+                    {
+                        continue;
+                    }
+
+                    float zdelta = Math.Abs(this.transform.position.z - point.position.z);
+                    if (zdelta > maxDistance) 
+                    {
+                        continue;
+                    }
```
Lucaweihs, line 1985: Why are these if statements necessary? Since `abs(xdelta) <= distance` and `abs(zdelta) <= distance` aren't these calls effectively redundant in view of the `if (distance > maxDistance) {` statement below?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2053,7 +2087,7 @@ protected SimObjPhysics[] GetAllVisibleSimObjPhysics(Camera agentCamera, float m
             }
             #endif
 
-            HashSet<SimObjPhysics> currentlyVisibleItems = new HashSet<SimObjPhysics>();
+            List<SimObjPhysics> currentlyVisibleItems = new List<SimObjPhysics>();
```
Lucaweihs, line 2090: Why make this a list? Won't this make the `if (!currentlyVisibleItems.Contains(sop))` call below into a linear time op rather than constant? Is the idea that the number of visible items is generally small so this is, in practice, is faster than not having it be a hashset?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2043,6 +2064,19 @@ public SimObjPhysics[] VisibleSimObjs(bool forceVisible)
...
+            foreach(var sop in simObjs)
+            {
+                if(isSimObjVisible(agentCamera, sop, this.maxVisibleDistance)) 
+                {
+                    visible.Add(sop);
+                }
+            }
+            return visible.ToArray();
+        }
+
```
Lucaweihs, line 2079: I still don't think this logic is equivalent to the old visibility logic. The main problem is with long objects, imagine the following setting:

1. We have a plunger laying on its side angled away from us so that half of the plunger is within the visibility distance and half is beyond the visibility distance.
2. The part of the plunger that is within the visibility distance is occluded while the portion of the plunger that is outside our visibility distance is not.

With this new logic, the plunger is considered **not visible** but, with the old logic, it **is visible**. This is because the OverlapCapsule call will add the plunger to the `colliders_in_view` list after which we only require that a single visibility point be visible at any distance for the object to be considered visible.

Lucaweihs (commented):
Some comments, I still don't think the two versions of visibility are equivalent.

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: This is true, but this new behavior may be more consistent as the default moving forward. It didn't make much sense in this example to have a plunger laying lengthwise still be visible in that way where the closer visibility points are obstructed and only the further away ones outside of the visibility distance are visible.

I think this is more of a question of which behavior seems more appropriate given the context of visual tasks that might require one behavior or another

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: The goal is to short circuit as early as possible and avoid having to compute the distance.  Let me benchmark without them though and see if it is actually contributing.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: All these changes are reverting back to the pre-simobject filter code to keep the visibility logic identical.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: I agree that the new behavior is more "reasonable" in some sense but as it fundamentally breaks backwards compatibility this isn't something we can just release. Perhaps it can be saved for our 3.0.0 release.

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: Hard to say... I think based on the scene and the actual angle you are looking at objects, one or the other may have better performance just depending on the exact situation. Overall I can't say definitively if the average list length is small enough that it will perform better or not...

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: In generally I think we have to be better in THOR about sticking to the semantic versioning guidelines (https://semver.org/) as, currently, even patch updates frequently break backwards compatibility.

ekolve: Luca, yes you are correct.  I hadn't considered that case.  I feel the new implementation makes more sense and is much easier to explain to others/document - it basically boils down > 1 visibility point within the maxVisibleDistance.  Under the old scheme you could also have something like an extremely long table that is obscured well outside the visibilitydistance, but one point 3m away is visible.  I don't think it makes sense to consider that "visible" as that implies its interactable.

We have a couple options:
1) I can modify the code to match the logic of the existing code
2) We do nothing and accept the new change
3) Modify the existing code to match the current logic, but also add a parameter to support the new scheme so that we can experiment with results

I would prefer we do 3 since I prefer the new logic, but there are unknown consequences to changes the visibility logic and we should run models with/without the new logic.

winthos: My vote is also for 3. I believe in the next big version update the default should be the new behavior with the option of the old, but for now making the new scheme optional while retaining the current behavior sounds good.

Lucaweihs: I'm happy with 3 and totally agree that the new logic is more sensible. I would even be happy with deprecating the old code visibility code on the next major version release.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: I just benchmarked this and keeping the shortcircuits in results in ~20% performance improvement.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1365,6 +1366,13 @@ public SimObjType ReceptableSimObjType()
 		return (SimObjType)Enum.Parse(typeof(SimObjType), receptacleObjectType);
 	}
 
+    public VisibilityScheme GetVisibilityScheme() {
+        if (string.IsNullOrEmpty(visibilityScheme)) {
+            return VisibilityScheme.Collider;
+        }
+		return (VisibilityScheme)Enum.Parse(typeof(VisibilityScheme), visibilityScheme);
+    }
+
```
Lucaweihs, line 1375: Rather than have null/empty represent `Collider`, why not have `visibilityScheme` have a default value of `"Collider"`?

Lucaweihs (commented):
One minor comment but otherwise LGTM!

unity/Assets/Scripts/AgentManager.cs
ekolve: I updated this to default to "Collider", but I also added some error handling if Enum can't parse the passed in param.  The original motivation was to handle the case if someone sent in an empty string for visibilityScheme, but checking for parse errors is more robust.

PR #485, opened 2020-10-01
seenPoints for reachablePositions and shortestPath

ekolve:
adding a seenPoints hashset to keep track of which points have been visited during the BFS for shortestPath and reachablePositions

I feel pretty confident about this, but want to be sure that this doesn't introduce any errors.  The output for both reachablePositions and shortestPath were identical when I tested this with and without the seenPoints hashset.  The benefit I saw is that the number of steps taken went from around 400-> 100 for reachable positions (FloorPlan28) and 190 -> 66 for shortestPath (FloorPlan28 mug).


AlvaroHG:
great, LGTM!

PR #486, opened 2020-10-01
adding state field and state machine to manage frame emitting instead…

ekolve:
This PR refactors away all the readyToEmit and actionComplete flags in favor of a state machine.  The states are:

Processing -> ActionComplete -> Emit (for physics>
Processing -> PendingFIxedUpdate -> ActionComplete -> Emit (drone)

This also dose away with the droneMode flag since the drone specific code could be moved to the  drone agent class.

The primary motivation for this refactor was to support render free emits.  This means that we can emit a patched metadata to the python side without actually have to run the entire render loop.  This allows us to make calls to the Unity side at a rate of 6000 - 8000 requests/s.  I only modified one action to use this so far, but we should be able to move anything that doesn't modify the scene and is just querying the environment to use this.

### DIFF
--- a/ai2thor/fifo_server.py
+++ b/ai2thor/fifo_server.py
@@ -27,6 +27,7 @@ class FieldType(IntEnum):
     CLASSES_IMAGE = 8
     IDS_IMAGE = 9
     THIRD_PARTY_IMAGE = 10
+    METADATA_PATCH = 11
     END_OF_MESSAGE = 255
 
 
@@ -42,6 +43,8 @@ class FifoServer(ai2thor.server.Server):
         self.client_pipe_path = os.path.join(self.tmp_dir.name, 'client.pipe')
         self.server_pipe = None
         self.client_pipe = None
+        self.raw_metadata = None
+        self.raw_files = None
         # allows us to map the enum to form field names
         # for backwards compatibility
         # this can be removed when the wsgi server is removed
@@ -96,11 +99,27 @@ class FifoServer(ai2thor.server.Server):
                 #print("body length %s" % len(body))
                 #print(body)
                 metadata = msgpack.loads(body, raw=False)
-                #print(metadata)
+            elif field_type is FieldType.METADATA_PATCH:
+                metadata_patch = msgpack.loads(body, raw=False)
+                agents = self.raw_metadata['agents']
+                metadata = dict(
+                    agents=[{} for i in range(len(agents))],
+                    thirdPartyCameras=self.raw_metadata['thirdPartyCameras'],
+                    sequenceId=self.sequence_id,
+                    activeAgentId=metadata_patch['agentId']
+                )
+                for i in range(len(agents)):
+                    metadata["agents"][i].update(agents[i])
+
+                metadata["agents"][metadata_patch['agentId']].update(metadata_patch)
+                files = self.raw_files
             elif field_type in self.image_fields:
                 files[self.form_field_map[field_type]].append(body)
             else:
                 raise ValueError("Invalid field type: %s" % field_type )
+
+        self.raw_metadata = metadata
+        self.raw_files = files
 
         return metadata, files
 
@@ -117,7 +136,12 @@ class FifoServer(ai2thor.server.Server):
         self.client_pipe.flush()
 
     def receive(self):
+
         metadata, files = self._recv_message()
+
+        if metadata is None:
+            raise ValueError("no metadata received from recv_message")
+
         return self.create_event(metadata, files)
 
     def send(self, action):
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -80,6 +80,25 @@ def test_small_aspect():
     event = controller.step(dict(action='Initialize', gridSize=0.25))
     assert event.frame.shape == (64, 128, 3)
 
+@pytest.mark.parametrize("controller", [fifo_controller])
+def test_fast_emit(controller):
+    event = controller.step(dict(action='RotateRight'))
+    event_fast_emit = controller.step(dict(action='TestFastEmit', rvalue='foo'))
+    # assert that when actionFastEmit is off that the objects are different
+    assert id(event.metadata['objects']) !=  id(event_fast_emit.metadata['objects'])
+
+    fast_controller = UnityTestController(server_class=FifoServer, fastActionEmit=True)
+    event = fast_controller.step(dict(action='RotateRight'))
+    event_fast_emit = fast_controller.step(dict(action='TestFastEmit', rvalue='foo'))
+    event_no_fast_emit = fast_controller.step(dict(action='LookUp'))
+    event_no_fast_emit_2 = fast_controller.step(dict(action='RotateRight'))
+
+    assert event.metadata['actionReturn'] is None
+    assert event_fast_emit.metadata['actionReturn'] == 'foo' 
+    assert id(event.metadata['objects']) ==  id(event_fast_emit.metadata['objects'])
+    assert id(event.metadata['objects']) !=  id(event_no_fast_emit.metadata['objects'])
+    assert id(event_no_fast_emit_2.metadata['objects']) !=  id(event_no_fast_emit.metadata['objects'])
+
 @pytest.mark.parametrize("controller", [wsgi_controller, fifo_controller])
 def test_lookdown(controller):
 
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -80,13 +80,7 @@ def test_small_aspect():
     event = controller.step(dict(action='Initialize', gridSize=0.25))
     assert event.frame.shape == (64, 128, 3)
 
-@pytest.mark.parametrize("controller", [fifo_controller])
-def test_fast_emit(controller):
-    event = controller.step(dict(action='RotateRight'))
-    event_fast_emit = controller.step(dict(action='TestFastEmit', rvalue='foo'))
-    # assert that when actionFastEmit is off that the objects are different
-    assert id(event.metadata['objects']) !=  id(event_fast_emit.metadata['objects'])
-
+def test_fast_emit():
     fast_controller = UnityTestController(server_class=FifoServer, fastActionEmit=True)
     event = fast_controller.step(dict(action='RotateRight'))
     event_fast_emit = fast_controller.step(dict(action='TestFastEmit', rvalue='foo'))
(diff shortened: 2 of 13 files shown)


ai2thor/fifo_server.py
```diff
@@ -96,12 +99,28 @@ def _recv_message(self):
                 #print("body length %s" % len(body))
                 #print(body)
                 metadata = msgpack.loads(body, raw=False)
-                #print(metadata)
+            elif field_type is FieldType.METADATA_PATCH:
+                metadata_patch = msgpack.loads(body, raw=False)
+                agents = self.raw_metadata['agents']
```
AlvaroHG, line 104: Can this be `None` the first time a message is received and is of type `FieldType.METADATA_PATCH`?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -728,147 +704,184 @@ private void addImageSynthesisImage(List<KeyValuePair<string, byte[]>> payload,
...
+        bool emit = true;
+        foreach (BaseFPSAgentController agent in this.agents) {
+            if (agent.agentState != AgentState.Emit) {
+                emit = false;
+                break;
+            }
+        }
 
-		frameCounter += 1;
+        return this.agentState == AgentState.Emit && emit;
```
AlvaroHG, line 716: why do you check for `this.agentState == AgentState.Emit` is it the primary agent? shouldn't it be in `this.agents` ?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -728,147 +704,184 @@ private void addImageSynthesisImage(List<KeyValuePair<string, byte[]>> payload,
...
-        MultiAgentMetadata multiMeta = new MultiAgentMetadata ();
-		
-
-		ThirdPartyCameraMetadata[] cameraMetadata = new ThirdPartyCameraMetadata[this.thirdPartyCameras.Count];
-        List<KeyValuePair<string, byte[]>> renderPayload = new List<KeyValuePair<string, byte[]>>();
-        createPayload(multiMeta, cameraMetadata, renderPayload, shouldRender);
 
+            frameCounter += 1;
+            if (this.agentState == AgentState.ActionComplete) {
+                this.agentState = AgentState.Emit;
```
AlvaroHG, line 728: Related to the commend above, It is a little bit confusing to me that we have both `agentState` in the `AgentManager` Class and in the `BaseFPSAgentController` class, since `AgentManager` just manages multiple agents and `AgentState` sounds like an Agent level property not a manager one?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -905,12 +918,16 @@ private void ProcessControlCommand(string msg)
         this.activeAgentId = controlCommand.agentId == null ? 0 : controlCommand.agentId;
 
 		if (controlCommand.action == "Reset") {
+            this.agentState = AgentState.Processing;
```
AlvaroHG, line 921: To not have to set the state like this for every action, maybe we should just set it outside of the `if` for the `string msg` version of `ProcessControlCommand`, and remove it from the others, minus the WebGL one which calls `ProcessControlCommand` not with a string due to the dynamic incompatibility.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -266,12 +283,21 @@ public void actionFinished(bool success, System.Object actionReturn=null)
...
             lastActionSuccess = success;
-			this.actionComplete = true;
+			this.agentState = newState;
 			this.actionReturn = actionReturn;
 			actionCounter = 0;
 			targetTeleport = Vector3.zero;
+
+        }
+
+		public virtual void actionFinished(bool success, System.Object actionReturn=null)
```
AlvaroHG, line 293: Why are there 2 `actionFinished` here? surprised the type system is not confused as I am of which one is being called from `actionFinishedEmit`

AlvaroHG: I see now, backcompat

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1563,7 +1601,7 @@ public void ProcessControlCommand(dynamic controlCommand)
             lastAction = controlCommand.action;
             lastActionSuccess = false;
             lastPosition = new Vector3(transform.position.x, transform.position.y, transform.position.z);
-            this.actionComplete = false;
+            this.agentState = AgentState.Processing;
```
AlvaroHG, line 1604: If moved to the `string` argument ProcessControlCommand, maybe remove this on here

unity/Assets/Scripts/DroneFPSAgentController.cs
```diff
@@ -88,6 +98,10 @@ void FixedUpdate()
             //here, the in-editor axisAlignedBoundingBox metadata for drone objects seems to be offset by some number of updates.
             //it's unclear whether this is only an in-editor debug draw issue, or the actual metadata for the axis
             //aligned box is messed up, but yeah.
+
+            if (this.agentState == AgentState.PendingFixedUpdate) {
+                this.agentState = AgentState.ActionComplete;
```
AlvaroHG, line 103: not very familiar with the dron controller, but semantically should this be done at the end of FixedUpdate?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -266,12 +283,21 @@ public void actionFinished(bool success, System.Object actionReturn=null)
...
 			this.actionReturn = actionReturn;
 			actionCounter = 0;
 			targetTeleport = Vector3.zero;
+
+        }
+
+		public virtual void actionFinished(bool success, System.Object actionReturn=null) 
+		{
+            actionFinished(success, AgentState.ActionComplete, actionReturn);
+            this.resumePhysics();
```
AlvaroHG, line 296: a lot cleaner for the Drone :)

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -728,147 +704,184 @@ private void addImageSynthesisImage(List<KeyValuePair<string, byte[]>> payload,
...
-        if(droneMode)
-        {
-            if (Time.timeScale == 0 && !Physics.autoSimulation && physicsSceneManager.physicsSimulationPaused)
-            {
-                DroneFPSAgentController agent_tmp = this.agents.ToArray()[0].GetComponent<DroneFPSAgentController>();
-                Time.timeScale = agent_tmp.autoResetTimeScale;
-                Physics.autoSimulation = true;
-                physicsSceneManager.physicsSimulationPaused = false;
-                agent_tmp.hasFixedUpdateHappened = false;
+                while (canEmit() && this.fastActionEmit) {
```
AlvaroHG, line 852: What happens for the case where `fastActionEmit` is passed as true to the action but the controller is not using the `fifoServer` ?

ai2thor/tests/test_unity.py
```diff
@@ -80,6 +80,25 @@ def test_small_aspect():
...
     assert event.frame.shape == (64, 128, 3)
 
+@pytest.mark.parametrize("controller", [fifo_controller])
+def test_fast_emit(controller):
+    event = controller.step(dict(action='RotateRight'))
+    event_fast_emit = controller.step(dict(action='TestFastEmit', rvalue='foo'))
+    # assert that when actionFastEmit is off that the objects are different
+    assert id(event.metadata['objects']) !=  id(event_fast_emit.metadata['objects'])
+
+    fast_controller = UnityTestController(server_class=FifoServer, fastActionEmit=True)
```
AlvaroHG, line 90: May be better to separate this into a second test, as it is testing different stuff and not using pytest's parametrization of the `controller` variable.

AlvaroHG:
Looking great, left a few comments and questions.

ai2thor/tests/test_unity.py
```diff
@@ -80,6 +80,25 @@ def test_small_aspect():
     event = controller.step(dict(action='Initialize', gridSize=0.25))
     assert event.frame.shape == (64, 128, 3)
 
+@pytest.mark.parametrize("controller", [fifo_controller])
+def test_fast_emit(controller):
+    event = controller.step(dict(action='RotateRight'))
+    event_fast_emit = controller.step(dict(action='TestFastEmit', rvalue='foo'))
+    # assert that when actionFastEmit is off that the objects are different
+    assert id(event.metadata['objects']) !=  id(event_fast_emit.metadata['objects'])
```
AlvaroHG, line 88: why are they different if the `TestFastEmit` action does not change the camera environment or anything? there is another metadata generation vs a metadata patch but shouldn't it be the same metadata generated?

ai2thor/fifo_server.py
ekolve: That can never happen since sending the patched data always occurs after the sending of a full metadata.

unity/Assets/Scripts/AgentManager.cs
ekolve: I agree it is a little weird in terms of naming, but we need to have state on the agentManager itself because it has actions like AddThirtpartyCamera, Initialize, etc.  So I figured it should go through the same state transitions.  I suppose it could be called agentManagerState to be less confusing?

unity/Assets/Scripts/AgentManager.cs
ekolve: Everything works as expected, you just don't get any speed improvements.  I just added a warning in controller.py so that users are aware that it only works with the FifoServer.

unity/Assets/Scripts/AgentManager.cs
ekolve: I just switched the AgentManager to use the ActionDispatcher and created a set of predefined agentManager action to clean that code up.

unity/Assets/Scripts/AgentManager.cs
ekolve: this.agentState (on AgentManager) should get renamed.  Since the agentManager also transitions through states because it handles actions, its state must also be checked.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: The string ProcessControlCommand only exists in the AgentManager, so we have to set this here since this is in the actual agent handling the request.

unity/Assets/Scripts/DroneFPSAgentController.cs
ekolve: At first I didn't follow what you meant, but I get it now.  Since the state transition indicates that we are no longer in PendingFixedUpdate and FixedUpdate only finishes at the end of the method, this transition should reside there.  I moved the block.  I don't think this changes the behavior, but I agree with your comment.

ai2thor/tests/test_unity.py
ekolve: I refactored the tests into two separate tests - note the comment above the comparison.  They are different in the case when fastActionEmit is disabled.

ai2thor/tests/test_unity.py
ekolve: agreed. refactored into a separate test.

PR #494, opened 2020-10-07
Erick/exp room arm height performance

ekolve:


AlvaroHG:
LGTM, we should put it in `moveArmTarget` too

winthos:
Yes, any changes should also be applied to `moveArmTarget` and `rotateHand`

PR #499, opened 2020-10-12
adding active collisions and method to retrieve active collisions aga…

ekolve:
…inst arm


AlvaroHG:
LGTM, here I feel I cannot suggest anything as I feel unity's physics system is pretty unpredictable, so we have to test this empirically.

PR #500, opened 2020-10-12
Exp room refactor

AlvaroHG:
* Adds optional restriction to moveArm function to not move behind shoulder.
* Refactors no render code and unifies into only 2 movement functions, IEnumerator and manual physics step

### DIFF
--- a/unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
+++ b/unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
@@ -51,6 +51,9 @@ public class IK_Robot_Arm_Controller : MonoBehaviour
     [SerializeField]
     private BoxCollider[] ArmBoxColliders;
 
+    // This value is wrong 0.6325f for the origin to shoulder, it should be the height of the z-oriented capsule, 0.34566f and we should get it dinamically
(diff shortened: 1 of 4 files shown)


unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -342,151 +347,117 @@ public void OnTriggerEnter(Collider col)
...
-
-        
-
-        controller.actionFinished(true);
+        controller.actionFinished(actionSuccess, debugMessage);
     }
 
-    public IEnumerator moveArmTarget(PhysicsRemoteFPSAgentController controller, Vector3 target, float unitsPerSecond,  GameObject arm, bool returnToStartPositionIfFailed = false, string whichSpace = "arm") {
+    // Restricts front hemisphere for arm movement
+    private bool validArmTargetPosition(Vector3 targetWorldPosition) {
```
winthos, line 440: So this will prevent the target position from being valid if behind the "base" joint of the arm, but we also need to restrict it so that  points beyond the forward hemisphere are not valid. If the targetWorldPosition's z is beyond the radius of this sphere. This radius is 0.6335, which is the distance from the base joint to the wrist.

PR #504, opened 2020-10-16
initializing properties on SimObjPhysics only once at startup

ekolve:


ekolve:
closing this as it doesn't result in much performance improvement.

ekolve:
Re-opened - i get around 10% improvement - not as much as I hoped, but its meaningful.

winthos (approved):
looks good to me

PR #508, opened 2020-10-20
using isSimObjVisible with shortestPath calculation if visibilityScheme is distance

ekolve:
Using visibilityScheme of Distance combined with this change and fastAction emit leads to ~10x speed improvement for GetShortestPath.  Before this at each point we check the visibility of all objects at each point, when only one object is needed.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -2933,6 +2933,9 @@ namespace UnityStandardAssets.Characters.FirstPerson
         }
 
         public bool  getReachablePositionToObjectVisible(SimObjPhysics targetSOP, out Vector3 pos, float gridMultiplier = 1.0f, int maxStepCount = 10000) {
+
+            Plane[] planes = GeometryUtility.CalculateFrustumPlanes(this.m_Camera);
+
             CapsuleCollider cc = GetComponent<CapsuleCollider>();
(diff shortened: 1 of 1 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2960,14 +2963,21 @@ int layerMask
                     //make sure to rotate just the Camera, not the whole agent
                     m_Camera.transform.LookAt(targetSOP.transform, transform.up);
 
-                    var visibleSimObjects = this.GetAllVisibleSimObjPhysics(this.maxVisibleDistance);
+                    bool isVisible = false;
+                    if (this.visibilityScheme == VisibilityScheme.Distance) {
+                        isVisible = isSimObjVisible(this.m_Camera, targetSOP, this.maxVisibleDistance, planes);
+                    } else {
+                        var visibleSimObjects = this.GetAllVisibleSimObjPhysics(this.maxVisibleDistance);
+                        isVisible = visibleSimObjects.Any(sop => sop.objectID == targetSOP.objectID);
```
AlvaroHG, line 2971: Have read around that it is not recommended to run Linq functions in frequent running functions in unity as it is quite slow, ok for initialization etc. but not great for update. 

Have not measured it so I am not sure, and regardless I vote for the new visibility scheme

AlvaroHG:
LGTM

PR #515, opened 2020-10-23
Continuous Move/Rotate and CollisionListener

AlvaroHG:
Refactors collision checking with the arm to a new component CollisionListener that can cascade collision events up to parent CollisionListeners, to achieve a uniform way of rotating both the arm and the agent and paving the way to allow self intersection.

* Adds CollisionListener
* Refactors collision checking code
* Adds ContinuousMovement class for generic utility continuous Transform update functions
* Removes old code in arm, one single function for moving
* Adds MoveContinuous/MoveRotate
* Re-implements RotateMidLevelHand using this and latest collision checking


winthos (approved):

unity/Assets/Scripts/CollisionListener.cs
```diff
@@ -0,0 +1,120 @@
...
+        public GameObject gameObject;
+
+        public SimObjPhysics simObjPhysics;
+
+         //indicates if gameObject a simObject
+        public bool isSimObj {
+            get { return simObjPhysics != null; }
+        }
+    }
+    private StaticCollision staticCollided;
```
ekolve, line 24: I think we can get rid of this.  It doesn't appear to be used.

unity/Assets/Scripts/CollisionListener.cs
```diff
@@ -0,0 +1,120 @@
...
+                if(sop.PrimaryProperty == SimObjPrimaryProperty.Static)
+                {
+
+                    if(!col.isTrigger)
+                    {
+                        // #if UNITY_EDITOR
+                        // Debug.Log("Collided with static sim obj " + sop.name);
+                        // #endif
+                        sc = new StaticCollision();
+                        sc.simObjPhysics = sop;
```
ekolve, line 86: In the previous code only a simObj or gameObject was stored on the StaticCollision.  I don't think it matters too much, but it forces us to only do the null check where we check for a simobj first and then a gameobject in a certain order.  We only do that during debugging, so I don't think it changes the actual logic.

unity/Assets/Scripts/ContinuousMovement.cs
```diff
@@ -0,0 +1,194 @@
...
+using UnityEngine;
+using UnityStandardAssets.Characters.FirstPerson;
+using System;
+
+
+namespace UnityStandardAssets.Characters.FirstPerson {
+
+    public class ContinuousMovement {
+
+    public static void unrollSimulatePhysics(IEnumerator enumerator, float fixedDeltaTime) {
```
ekolve, line 12: Nice! This is what I was referring to when using the IEnumerable to unify the code.

ekolve (commented):
I few comments, but otherwise looks good.

unity/Assets/Scripts/CollisionListener.cs
winthos: agreed, this was from the previous version of the logic so should no longer be needed

AlvaroHG: Done, sneaked past me

unity/Assets/Scripts/CollisionListener.cs
AlvaroHG: I wanted to simplify the check to not have an if that looked like `if (sc.simObjPhysics != null && sc.gameObject == null) { // then it is a simobject` which we had, so basically now if it is registered as a "StaticCollision" the gameObject will always be non-Null and there is a `isSimObject` field for quick checking if we collided against a SimObject.

The way to check if it is/is not a simObject you only call `sc.isSimObject`, and you avoid the null checking for the game object so it is the same one check, I just think is cleaner to have less nulls around and simplifies the above case. 
Does this make sense?

unity/Assets/Scripts/ContinuousMovement.cs
AlvaroHG: Yeah good idea, thanks for mentioning it

PR #534, opened 2020-11-25
integrate new ManipulateArm() command from IKSolver into continuous m…

ekolve:
This changes the following:

* restores the snapping to target
* removes waitForFixedUpdate param
* integrates Eli's solver into ContinuousMovement
* Removes the resetProperty Func in favor of using setProperty() - this works now that we detect collisions immediately after impact
* Prevent overshooting by calculating the nextProp value and snapping to the target if it is closer than the nextProp.


winthos (approved):

AlvaroHG:
LGTM

PR #536, opened 2020-12-02
Physics Determinism Test

AlvaroHG:
* Adds trial runner as a generic sequence generator that gathers a metric after running controller code
* Adds a object position variance average metric used for physics determinism
* Task for testing determinism of basic actions

Results for 120 trials on a 100 random actions of each category:
```
inv get-physics-determinism -s=FloorPlan212_physics -n=120 --samples=samples

actions: 'move', object_position_variance_average: 1.0702238729969497e-11 
actions: 'rotate', object_position_variance_average: 3.6856343950937414e-12 
actions: 'look', object_position_variance_average: 4.872177661890314e-12 
actions: 'all', object_position_variance_average: 0.00029112702368825385

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -392,7 +392,7 @@ class Controller(object):
             local_build = True
 
         self.last_event = None
-        self.current_scene_name = None
+        self.scene = None
         self._scenes_in_build = None
         self.killing_unity = False
         self.quality = quality
@@ -504,10 +504,10 @@ class Controller(object):
 
         return self._scenes_in_build
 
-    def reset(self, scene=''):
+    def reset(self, scene=None):
         # Case for just resetting the current scene
-        if scene == '' and self.current_scene_name is not None:
-            scene = self.current_scene_name
+        if scene is None and self.scene is not None:
+            scene = self.scene
 
         if re.match(r'^FloorPlan[0-9]+$', scene):
             scene = scene + "_physics"
@@ -530,7 +530,7 @@ class Controller(object):
         
         self.last_event = self.step(action='Initialize', **self.initialization_parameters)
 
-        self.current_scene_name = scene
+        self.scene = scene
 
         return self.last_event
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: ai2thor/controller.py [hunk 1]
```diff
@@ -503,7 +504,11 @@ def scenes_in_build(self):
 
         return self._scenes_in_build
 
-    def reset(self, scene='FloorPlan_Train1_1'):
+    def reset(self, scene=''):
```
ekolve, line 507: Can you make the default for scene None instead of an empty string?  With the current code you could do this:

```
controller.reset(None)
```
and you would get an error during concatenation

If you make the default None and someone did do something like pass in an empty string, they would get an error that the scene does not exist.

ai2thor/controller.py
```diff
@@ -392,6 +392,7 @@ def __init__(
             local_build = True
 
         self.last_event = None
+        self.current_scene_name = None
```
ekolve, line 395: Would you be opposed to naming this self.scene instead of self.current_scene_name?  Since there is only ever one scene loaded it won't conflict with anything and then the member variable will match the parameter to the constructor which makes the property a bit easier to discover by the end user.

ai2thor/util/metrics.py
```diff
@@ -2,6 +2,7 @@
 import json
 import math
 import copy
+import numpy as np
```
ekolve, line 5: It doesn't appear that numpy is used in the file.

ekolve (changes requested):
a couple small changes requested, otherwise looks good!

ai2thor/controller.py
AlvaroHG: Done

ai2thor/controller.py
AlvaroHG: Done

ai2thor/util/metrics.py
AlvaroHG: Removed

ekolve:
looks good, feel free to merge.

ekolve (approved):

PR #539, opened 2020-12-08
Improving the ShortestPath and PlaceObjectAtPoint functions.

Lucaweihs:
## Problem(s)

1. The `ShortestPath` functions frequently return that no path exists despite there very clearly being one.
2. The `PlaceObjectAtPoint` code did not allow for forcing the placed object to be kinematic.

## Solutions

1. I added checks and additional code to make the `ShortestPath` functions more robust. Namely I allow for a small error when specifying the starting/target positions, this handles a large collection of edge cases.
2. I refactored the `PlaceObjectAtPoint` function and added the `forceKinematic` option.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -3047,11 +3047,11 @@ namespace UnityStandardAssets.Characters.FirstPerson
                 new Vector3(targetPosition.x, targetHit.position.y, targetPosition.z)
             );
             if (startOffset > allowedError && targetOffset > allowedError) {
-                if (!startWasHit) {
-                    errorMessage = $"Closest point on NavMesh was too far from the agent: " + 
-                    $" (startPosition={startPosition}, closest navmesh position {startHit.position}) and" +
-                    $" (targetPosition={targetPosition}, closest navmesh position {targetHit.position}).";
-                }
+                errorMessage = $"Closest point on NavMesh was too far from the agent: " + 
+                    $" (startPosition={startPosition.ToString("F3")}," +
+                    $" closest navmesh position {startHit.position.ToString("F3")}) and" +
+                    $" (targetPosition={targetPosition.ToString("F3")}," +
+                    $" closest navmesh position {targetHit.position.ToString("F3")}).";
                 this.GetComponent<UnityEngine.AI.NavMeshAgent>().enabled = false;
                 return false;
             }
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -3070,7 +3070,8 @@ namespace UnityStandardAssets.Characters.FirstPerson
                 return true;
             }
             else {
-                errorMessage = "Path to target could not be found";
+                errorMessage = $"Could not find path between {startHit.position}" +
+                    " and {targetHit.position} using the NavMesh.";
                 this.GetComponent<UnityEngine.AI.NavMeshAgent>().enabled = false;
                 return false;
             }
@@ -3079,7 +3080,7 @@ namespace UnityStandardAssets.Characters.FirstPerson
             Vector3 position, float x, float y, float z, float allowedError = 0.01f
         ) {
             var path = new UnityEngine.AI.NavMeshPath();
-            if (SafelyComputeNavMeshPath(position, new Vector3(x,y,z), path, allowedError)) {
+            if (SafelyComputeNavMeshPath(position, new Vector3(x, y, z), path, allowedError)) {
                 actionFinished(true, path);
             } else {
                 actionFinished(false);
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -2644,13 +2644,19 @@ namespace UnityStandardAssets.Characters.FirstPerson
             actionFinished(true, reachablePositions);
         }
 
-        private void getShortestPath(string objectType, string objectId,  Vector3 startPosition, Quaternion startRotation) {
+        private void getShortestPath(
+            string objectType,
+            string objectId,
+            Vector3 startPosition,
+            Quaternion startRotation,
+            float allowedError
+        ) {
             SimObjPhysics sop = getSimObjectFromTypeOrId(objectType, objectId);
             if (sop == null) {
                 actionFinished(false);
                 return;
             }
-            var path = GetSimObjectNavMeshTarget(sop, startPosition, startRotation);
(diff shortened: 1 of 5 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2968,38 +2990,123 @@ int layerMask
...
-            this.GetComponent<UnityEngine.AI.NavMeshAgent>().enabled = false;
 
             return path;
         }
 
-        public void GetShortestPathToPoint(Vector3 position, float x, float y, float z) {
-            Vector3 startPosition = position;
-            var targetPosition = new Vector3(x, y, z);
+        protected float getFloorY(float x, float start_y, float z) {
+            int layerMask = ~(1 << 10 | 1 << 9);
```
AlvaroHG, line 2998: Could you use the `LayerMask.GetMask("TargetLayer")` function to get the layers, to make it easier to read in the future?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2968,38 +2990,123 @@ int layerMask
...
+            Ray ray = new Ray(new Vector3(x, y, z), -transform.up);
+            if (!Physics.Raycast(ray, out hit, 100f, layerMask)) {
+                errorMessage = "Could not find the floor";
+                return float.NegativeInfinity;
+            }
+            return hit.point.y;
+        }
+        
+        protected float getFloorY(float x, float z) {
+            int layerMask = ~(1 << 10);
```
AlvaroHG, line 3011: Same comment as above

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2968,38 +2990,123 @@ int layerMask
...
+            bool targetWasHit = UnityEngine.AI.NavMesh.SamplePosition(
+                targetPosition, out targetHit, 0.2f, UnityEngine.AI.NavMesh.AllAreas
+            );
+
+            if (!startWasHit || !targetWasHit) {
+                if (!startWasHit) {
+                    errorMessage = $"No point on NavMesh near {startPosition}.";
+                }
+                if (!targetWasHit) {
+                    errorMessage = $"No point on NavMesh near {targetPosition}.";
```
AlvaroHG, line 3052: This is great!

AlvaroHG:
minor comment and LGTM!

unity/Assets/Scripts/DebugInputField.cs
```diff
@@ -895,12 +895,41 @@ public void Execute(string command)
...
+                        BaseFPSAgentController agent = AManager.agents[0];
+
+                        GameObject itemInHand = AManager.agents[0].ItemInHand;
+                        if (itemInHand != null) {
+                            itemInHand.SetActive(false);
+                        }
+
+                        RaycastHit hit;
+                        Ray ray = agent.m_Camera.ViewportPointToRay(new Vector3(0.5f, 0.5f, 0.0f));
+                        bool raycastDidHit = Physics.Raycast(ray, out hit, 10f, (1 << 8) | (1 << 10));
```
winthos, line 910: Even though this is just for debug purposes, this functionality seems nice. Currently PlaceObjectAtPoint requires you to pass in a position that you obtain somewhere else, but it might be worth considering using a screenspace raycast as an optional way to "target" the position you are trying to place an object at.

This may require more testing of course, but we should keep a note that this functionality might be something good to build in to the actual action as well.

winthos (approved):
Also minor comment and looks good!

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: I did not know that `LayerMask.GetMask` existed! Done.

unity/Assets/Scripts/BaseFPSAgentController.cs

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2662,16 +2669,27 @@ public void InitialRandomSpawn(ServerAction action)
...
         }
 
-        public void GetShortestPath(Vector3 position, Vector3 rotation, string objectType = null, string objectId = null) {
-            getShortestPath(objectType, objectId, position, Quaternion.Euler(rotation));
+        public void GetShortestPath(
+            Vector3 position,
+            Vector3 rotation,
+            string objectType = null,
+            string objectId = null,
+            float allowedError = 0.01f
```
ekolve, line 2677: We have the default of 0.01 for allowedError in both python and Unity.  I'd recommend the default occur at the lowest layer (Unity) and just default None on the Python side.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4593,41 +4585,44 @@ public void PickupContainedObjects(SimObjPhysics target)
             }
         }
 
-        public void DropContainedObjects(SimObjPhysics target) {
+        public void DropContainedObjects(
```
ekolve, line 4588: I know you didn't write this method originally, but could it be made protected or package level visibility instead of public?

ekolve (approved):
a few comments, otherwise looks good.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
Lucaweihs: It seems to be used within the `InstatiatePrefabTest` class so I believe it needs to be public.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: The downside of doing this is that it's not immediately clear to the user what the default value is, they would have to look at the Unity code directly. I could specify that we use a default value of 0.01 in the python documentation but I feel that this is strictly worse as now we could change things on the Unity side and forget to change them on the python side.

I would make it a required parameter at the Unity level but this would break backwards compatibility.

ekolve: Here are a few options - 
1) make an overloaded version that has allowedError required (use the default from python).  The method without allowedError would have the backwards compatible behavior.  
2) Default allowedError to 0.0 if that would preserve existing behavior.  
3) Finally, just leave as is if you think the new behavior is better

Lucaweihs: 👍 Made `allowedError` default to a small positive number (worried about it being exactly 0 due to possible floating point issues) in Unity and the python documentation simply says that Unity will choose some small positive number so people (theoretically) should not depend on this being consistent.

PR #544, opened 2020-12-10
Exp room held object clip

winthos:
adding fix for objects held by the arm to also correctly work with the collision listener


winthos:
closing for now since we are using cascading collision listeners instead of the ConvertCollider logic. Don't close this branch for now because we may need to recycle the ConvertCollider logic at some point in the future.

Issue #546, opened 2020-12-14
Issue with Teleport behavior


zkytony:
It would be nice (and expected) if when calling `TeleportFull`, the coordinates that are not specified are kept the same.

```python
from ai2thor.controller import Controller
controller = Controller(scene="FloorPlan_Train1_1")

# Initial pose
event = controller.step("Pass")
print("position1:", event.metadata["agent"]["position"])
print("rotation1:", event.metadata["agent"]["rotation"])

# I want the agent to teleport to:
#   x=3.25, z=-1.5, rotation_y=270.0
# without changing other coordinates
event = controller.step('TeleportFull',
                        x=3.25, z=-1.5,
                        rotation=dict(y=270.0))
print("position2:", event.metadata["agent"]["position"])
print("rotation2:", event.metadata["agent"]["rotation"])
```

This gives me the following output:
```
position1: {'x': 3.0, 'y': 0.9009997, 'z': -1.5}
rotation1: {'x': 0.0, 'y': 269.999542, 'z': 0.0}
position2: {'x': 3.25, 'y': -39.24, 'z': -1.5}
rotation2: {'x': 0.0, 'y': 270.0, 'z': 0.0}
```
You can see that somehow `y` is set to `-39.24` even though I didn't specify this setting when I called `Teleport`. However, the angles are not subject to this unexpected behavior.

I am using ai2thor 2.4.22 on Ubuntu 18.04.

winthos:
I believe what is happening with that `y` value is the coordinate you teleport to is outside of the room bounds or perhaps in the half inside one of the walls, causing the agent to fall through the floor. That very large negative Y is expected if the agent has fallen through the floor, which can happen when teleporting.

After a teleport, a helper function is called that uses gravity to snap the agent to the ground, which is why the Y value will be reported back as `0.9009997` even if you were to pass in `1` as the y value, because the agent will fall slightly to settle on the surface of the floor. If there is no floor, the agent will fall so you will see that very large negative value.

mattdeitke:
I think the issue is more so that the unspecified parameters for `Teleport` behave super weirdly. For instance, by not specifying `y` for the position (which definitely shouldn't be necessary since the value is generally an arbitrary float), the value should simply remain the same on the agent.

I should be able to push a fix for this soon, though I'm checking if it'd break anything at the moment.

The fix would also solve #447, which depreciates TeleportFull (since it would no longer be necessary with better default parameters).

zkytony:
[...]

Issue #447, opened 2020-08-15
Merge Teleport and TeleportFull?


mattdeitke:
`TeleportFull` seems like a misuse of leveraging default key-word args.

---

For reference, `Teleport` is defined as
```python
controller.step(
    action='Teleport',
    x=0.999, y=1.01, z=-0.3541)
```
and `TeleportFull` is defined the same way, but with 2 extra args:
```python
event = controller.step(
    action='TeleportFull',
    x=0.999, y=1.01, z=-0.3541,
    rotation=dict(x=0.0, y=90.0, z=0.0),
    horizon=30.0)
```

---

What would seem easier from a user's perspective would be to allow users to **optionally** specify the `rotation`, `horizon`, and also `x`, `y`, `z` positions. And for all the options that are not specified, leave them **as is**.

For example, running
```python
controller.step(
    action='Teleport',
    x=0.999, z=-0.3541,
    horizon=-30)
```
Will change only the x/z position and horizon of the agent, and leave the y/rotation as is. This will make it so that users don't have to know some crazy float `y` value that's only accessible from `GetReachablePositions`.

Effectively, this should depreciate the need for `TeleportFull` since all it's functionality would be merged with `Teleport`.

PR #547, opened 2020-12-15
Teleport Enhancements

mattdeitke:
Solves #546 and #447:

- Sets unspecified parameters to their defaults on the agent.
- Depreciates TeleportFull (all functionality merged with Teleport).

---
Currently waiting on the local build before running a few tests.


winthos (approved):
looks good!

Lucaweihs:
I spoke with @ekolve about this a bit and would recommend instead creating a new function to have this functionality. I.e. we should have three methods:
* `Teleport` - Requires all parameters, if one isn't given, it results in an error.
* `TeleportFull` - Identical to Teleport, kept around for backwards compatibility.
* `TeleportPartial` - Allows for this defaulting behavior you've proposed.

PR #550, opened 2020-12-16
Exp room mass threshold

winthos:
adding toggle for using mass threshold on mid level arm collision listener

-new initialize params:

useMassThreshold = True
massThreshold = <some float>


unity/Assets/Scripts/AgentManager.cs
```diff
@@ -213,7 +213,22 @@ public void Initialize(ServerAction action)
...
                 SetUpArmController(true);
+
+				if(action.useMassThreshold)
+				{
+					if(action.massThreshold > 0.0)
+					SetUpMassThreshold(action.massThreshold);
+
+					else
+					{
+						var error = "massThreshold must have nonzero value if useMassThreshold = True";
```
winthos, line 224: I was getting something strange when testing this error check. I get an in-editor debug log error of:

"ActionFinished called with agentState not in procesing"

this is in `actionFinished` inside of the `BaseFPSAgentController.cs`

It seems this only throws an editor log error, but doesn't seem to change the actual behavior of actionFinished?

unity/Assets/Scripts/CollisionListener.cs
```diff
@@ -31,8 +31,20 @@ public class StaticCollision
             }
         }
     }
+
     private HashSet<Collider> activeColliders = new HashSet<Collider>();
+    public static bool useMassThreshold = false;
+    public static float massThreshold = 10f;
```
ekolve, line 37: Why is this static, but you set the value using an instance of collisionListener? If you want the value to apply to all the CollisionListeners, you could just change the setter to public static get/set property or set the value directly on CollisionListener (class not instance).

unity/Assets/Scripts/CollisionListener.cs
```diff
@@ -31,8 +31,20 @@ public class StaticCollision
...
+
+    public void Set_massThreshold (float f)
```
ekolve, line 39: Can you remove the underscore and rename to SetMassThreshold, same for Set

ekolve (changes requested):

unity/Assets/Scripts/CollisionListener.cs
winthos: Ah yes I see, I think we do want this to apply to all CollisionListeners in case we ever have multiple agents, which would then have multiple instances of the CollisionListener component existing in the same scene.

I will change this to set the values directly via the CollisionListener class.

unity/Assets/Scripts/CollisionListener.cs
winthos: removing these since it would be better to set the values directly as mentioned above.

PR #559, opened 2021-01-05
Enhance reset

mattdeitke:
Currently, the only documented way to make updates to initialization is to call:
```python
controller = Controller(init_params)
controller.stop()
controller = Controller(new_initialization_params)
```
However, this is highly undesirable when working with an interactive shell since: 
* It takes a non-trivial amount of time to create and destroy a controller. 
* If one doesn't first call `controller.stop()` and later initializes a controller with the same name `controller`, there is no way to ever stop it (without resetting the interactive session). For instance:
```python
controller = Controller(renderDepthImage=True)
controller = Controller()

# the first `controller` is no longer reachable, but it's Unity window still renders
```

This PR simply extends the abilities of `reset` to not only take in a `scene` but also take in any initialization parameter, such as:
```python
controller.reset(scene='FloorPlan1', renderDepthImage=True, agentMode='stochastic', width=4500)
```

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -386,6 +386,8 @@ class Controller(object):
         self.unity_pid = None
         self.docker_enabled = docker_enabled
         self.container_id = None
+        self.width = width
+        self.height = height
 
         self.last_event = None
         self._scenes_in_build = None
@@ -500,7 +502,7 @@ class Controller(object):
 
         return self._scenes_in_build
 
-    def reset(self, scene='FloorPlan_Train1_1'):
+    def reset(self, scene='FloorPlan_Train1_1', **init_params):
         if re.match(r'^FloorPlan[0-9]+$', scene):
             scene = scene + "_physics"
 
@@ -519,7 +521,22 @@ class Controller(object):
 
         self.server.send(dict(action='Reset', sceneName=scene, sequenceId=0))
         self.last_event = self.server.receive()
-        
+
+        # update the initialization parameters
+        init_params = init_params.copy()
+
+        # width and height are updates in 'ChangeResolution', not 'Initialize'
+        if 'width' in init_params or 'height' in init_params:
+            if 'width' in init_params:
+                self.width = init_params['width']
+                del init_params['width']
+            if 'height' in init_params:
+                self.height = init_params['height']
+                del init_params['height']
+            self.step(action='ChangeResolution', x=self.width, y=self.height)
+
+        # updates the initialization parameters
+        self.initialization_parameters.update(init_params)
         self.last_event = self.step(action='Initialize', **self.initialization_parameters)
 
         return self.last_event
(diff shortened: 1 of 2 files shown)


ai2thor/controller.py
```diff
@@ -519,7 +521,22 @@ def key_sort_func(scene_name):
...
+
+        # width and height are updates in 'ChangeResolution', not 'Initialize'
+        if 'width' in init_params or 'height' in init_params:
+            if 'width' in init_params:
+                self.width = init_params['width']
+                del init_params['width']
+            if 'height' in init_params:
+                self.height = init_params['height']
+                del init_params['height']
+            self.step(action='ChangeResolution', x=self.width, y=self.height)
```
Lucaweihs, line 536: Can we not run `ChangeResolution` if the resolution is already correct? I've had weird issues with `ChangeResolution` in the past and so want to run it as few times as possible.

Lucaweihs (commented):
One small comment, otherwise LGTM.

ai2thor/controller.py
mattdeitke: Agree. It already shouldn't run if it's already correct. It's inside of the `if 'width' in init_params or 'height' in init_params:` condition.

mattdeitke: Actually can also add if `width == self.width` and `height == self.height` do not run.

mattdeitke: Updated with https://github.com/allenai/ai2thor/pull/559/commits/c2cfbafbfc36a927d0d9e592be30951c06c35617

ai2thor/controller.py
```diff
@@ -519,7 +521,24 @@ def key_sort_func(scene_name):
...
+        # update the initialization parameters
+        init_params = init_params.copy()
+
+        # width and height are updates in 'ChangeResolution', not 'Initialize'
+        if ('width' in init_params and  init_params['width'] != self.width) or (
+            'height' in init_params and init_params['height'] != self.height
+        ):
+            if 'width' in init_params:
+                self.width = init_params['width']
+                del init_params['width']
```
ekolve, line 534: Is the reason for deleting the param here to avoid forwarding them to Unity?

ekolve (changes requested):
Looks good - can you also add a test? Take a look at this for an example:

https://github.com/allenai/ai2thor/blob/19f3e233a5a99021200f89530e992c545480d016/ai2thor/tests/test_unity.py#L55

I think just testing that you can change the resolution or add depth then remove depth, then asserting should be sufficient.

ai2thor/controller.py
mattdeitke: Yeah. `width` and `height` aren't parameters within `controller.step('Initialize', ...)`.

Also, a copy of `init_params` is made, so deleting something should not be noticeable by the user.

mattdeitke:
@erick tests added (and passed) with https://github.com/allenai/ai2thor/pull/559/commits/b13dfe2428da96529d6ead12e3e424811e1de90a. Looks good on my end.

ekolve (approved):

ekolve:
lgtm

Lucaweihs:
LGTM.

PR #561, opened 2021-01-05
print -> logging

mattdeitke:
Changes from `print` to `logging.info` for consistency with other logging across the repo.

---

For reference, every time a Controller is currently initialized it runs:
```python
print("Initialize return: {}".format(init_return))
```
Within notebooks, particularly, this is non-trivial to suppress and forever displays the print statement after initializing a Controller.

---

For one to access this info here, they'd have to set
```python
import logging
logging.basicConfig(level=logging.INFO)
```
within the script that they initialize a Controller.

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -454,7 +454,7 @@ class Controller(object):
                 init_return = event.metadata['actionReturn']
                 self.server.set_init_params(init_return)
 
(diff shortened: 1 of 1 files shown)


ekolve (approved):

ekolve:
lgtm

PR #566, opened 2021-01-10
enable ResetColors, alias with RandomColors

mattdeitke:
Now, instead of destroying a controller to reset from crazy colors, that is:
```python
controller = Controller(scene='FloorPlan28')
controller.step('RandomColors')

controller.stop()
controller = Controller(scene='FloorPlan28')
```

one simply can call `ResetColors`:
```python
controller = Controller(scene='FloorPlan28')

controller.step('RandomColors')  # crazy random colors
controller.step('ResetColors')  # reset colors
```

I've also created an alias from the undocumented `ChangeColorOfMaterials` to a new name: `RandomColors`. The name `ChangeColorOfMaterials` was too hard to remember and the plurals on colors vs. materials seemed arbitrary.

This should be enough to officially support both `RandomColors` and `ResetColors` officially.

---

Note that these actions have also moved to `BaseFPSAgentController`, instead of `PhysicsRemoteFPSAgentController` because they do not require the use of any particular agent.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -710,6 +710,23 @@ namespace UnityStandardAssets.Characters.FirstPerson
             }
         }
 
+        // alias to RandomColors, supported for backwards compatibility
+        public void ChangeColorOfMaterials() {
+            RandomColors();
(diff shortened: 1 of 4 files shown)


elimvb:
Looks good!

PR #568, opened 2021-01-11
GetInteractablePoses, dynamic rotateStepDegrees, rotation instead of eulerAngles bug fixes

mattdeitke:
Aliases `PositionsFromWhichItemIsInteractable` to its new name `GetInteractablePositions`, since the former is **super** long. `GetInteractablePositions` is also more consistent with `GetReachablePositions`.

Also updates rotations to use `rotateStepDegrees`, instead of assuming that the agent is rotating in 90-degree increments.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -2804,7 +2804,7 @@ namespace UnityStandardAssets.Characters.FirstPerson
             SimObjPhysics soInHand = ItemInHand.GetComponent<SimObjPhysics>();
 
             Vector3 handObjPosRelAgent =
-                Quaternion.Euler(0, rotation - transform.rotation.y, 0) *
+                Quaternion.Euler(0, rotation - transform.eulerAngles.y, 0) *
                 (transform.position - ItemInHand.transform.position);
 
             Vector3 newHandPosition = handObjPosRelAgent + newAgentPosition;
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -1513,7 +1513,7 @@ namespace UnityStandardAssets.Characters.FirstPerson {
                     return;
                 }
                 Quaternion oldRotation = sop.transform.rotation;
-                sop.transform.rotation = Quaternion.Euler(new Vector3(0.0f, (float) Math.Round(sop.transform.eulerAngles.y + 90f % 360), 0.0f));;
+                sop.transform.rotation = Quaternion.Euler(new Vector3(0.0f, (float) Math.Round((sop.transform.eulerAngles.y + 90f) % 360), 0.0f));;
                 if (!action.forceAction) {
                     if (action.maxAgentsDistance > 0.0f) {
                         for (int i = 0; i < agentManager.agents.Count; i++) {
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -6945,12 +6945,25 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             Vector3[] positions = null,
             float[] rotations = null,
             float[] horizons = null,
-            bool[] standings = null
+            bool[] standings = null,
+            float? maxDistance = null
         ) {
             if (360 % rotateStepDegrees != 0) {
                 errorMessage = "360 % rotateStepDegrees must be 0, unless 'rotations: float[]' is overwritten.";
                 actionFinished(false);
                 return;
+            }
+
+            // default "visibility" distance
+            float maxDistanceFloat;
+            if (maxDistance == null) {
+                maxDistanceFloat = maxVisibleDistance;
+            } else if ((float) maxDistance <= 0) {
+                errorMessage = "maxDistance must be >= 0 meters from the object.";
+                actionFinished(false);
+                return;
+            } else {
+                maxDistanceFloat = (float) maxDistance;
             }
 
             if (!physicsSceneManager.ObjectIdToSimObjPhysics.ContainsKey(objectId)) {
@@ -6992,12 +7005,11 @@ namespace UnityStandardAssets.Characters.FirstPerson {
                 // This may happen if the agent starts by teleports with the rotation of 10 degrees.
                 float offset = transform.rotation.y % rotateStepDegrees;
 
-                rotations = new float[(int) (360 / rotateStepDegrees)];
-                int i = 0;
-
                 // Examples:
                 // if rotateStepDegrees=10 and offset=70, then the paths would be [70, 80, ..., 400, 410, 420].
                 // if rotateStepDegrees=90 and offset=10, then the paths would be [10, 100, 190, 280]
+                rotations = new float[(int) (360 / rotateStepDegrees)];
+                int i = 0;
                 for (float rotation = offset; rotation < 360 + offset; rotation += rotateStepDegrees) {
                     rotations[i++] = rotation;
                 }
@@ -7016,7 +7028,7 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             }
             float fudgeFactor = objectBounds.extents.magnitude;
             List<Vector3> filteredPositions = positions.Where(
-                p => (Vector3.Distance(a: p, b: theObject.transform.position) <= maxVisibleDistance + fudgeFactor + gridSize)
+                p => (Vector3.Distance(a: p, b: theObject.transform.position) <= maxDistanceFloat + fudgeFactor + gridSize)
             ).ToList();
 
             // set each key to store a list
@@ -7051,7 +7063,7 @@ namespace UnityStandardAssets.Characters.FirstPerson {
 
                             // Each of these values is directly compatible with TeleportFull
                             // and should be used with .step(action='TeleportFull', **interactable_positions[0])
-                            if (objectIsCurrentlyVisible(theObject, maxVisibleDistance)) {
+                            if (objectIsCurrentlyVisible(theObject, maxDistanceFloat)) {
                                 validAgentPoses["x"].Add(position.x);
                                 validAgentPoses["y"].Add(position.y);
                                 validAgentPoses["z"].Add(position.z);
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -1513,7 +1513,7 @@ namespace UnityStandardAssets.Characters.FirstPerson {
                     return;
                 }
                 Quaternion oldRotation = sop.transform.rotation;
-                sop.transform.rotation = Quaternion.Euler(new Vector3(0.0f, (float) Math.Round(sop.transform.rotation.y + 90f % 360), 0.0f));;
+                sop.transform.rotation = Quaternion.Euler(new Vector3(0.0f, (float) Math.Round(sop.transform.eulerAngles.y + 90f % 360), 0.0f));;
                 if (!action.forceAction) {
                     if (action.maxAgentsDistance > 0.0f) {
                         for (int i = 0; i < agentManager.agents.Count; i++) {
@@ -7006,14 +7006,14 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             // populate the rotations based on rotateStepDegrees
             if (rotations == null) {
                 // Consider the case where one does not want to move on a perfect grid, and is currently moving
-                //  with an offsetted set of rotations like {10, 100, 190, 280} instead of the default {0, 90, 180, 270}.
+                // with an offsetted set of rotations like {10, 100, 190, 280} instead of the default {0, 90, 180, 270}.
                 // This may happen if the agent starts by teleports with the rotation of 10 degrees.
-                float offset = transform.rotation.y % rotateStepDegrees;
+                int offset = (int) Math.Round(transform.eulerAngles.y % rotateStepDegrees);
 
                 // Examples:
                 // if rotateStepDegrees=10 and offset=70, then the paths would be [70, 80, ..., 400, 410, 420].
                 // if rotateStepDegrees=90 and offset=10, then the paths would be [10, 100, 190, 280]
-                rotations = new float[(int) (360 / rotateStepDegrees)];
+                rotations = new float[(int) Math.Round(360 / rotateStepDegrees)];
                 int i = 0;
                 for (float rotation = offset; rotation < 360 + offset; rotation += rotateStepDegrees) {
                     rotations[i++] = rotation;
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -6870,16 +6870,6 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             toReturn["objectSeen"] = objectSeen ? 1 : 0;
             toReturn["positionsTried"] = positionsTried;
 
-            if (wasStanding) {
-                stand();
-            } else {
-                crouch();
-            }
-            transform.position = oldPosition;
-            transform.rotation = oldRotation;
-            if (ItemInHand != null) {
-                ItemInHand.gameObject.SetActive(true);
-            }
             actionFinished(true, toReturn);
         }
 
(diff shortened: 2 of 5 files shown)


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6936,129 +6936,149 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
-                errorMessage = "Horizon value for PositionsFromWhichItemIsInteractable must be in range [0, 30] inclusive";
+        // Get the poses with which the agent can interact with 'objectId'
+        // @rotations: if rotation is not specified, we use rotateStepDegrees, which results in [0, 90, 180, 270] by default.
+        public void GetInteractablePositions(
+            string objectId,
+            Vector3[] positions = null,
+            float[] rotations = null,
+            float[] horizons = null,
+            bool[] standings = null
+        ) {
```
mattdeitke, line 6949: These allow you to ONLY check for certain positions, rotations, horizons, or standing booleans (i.e., standing is `True` or `False`).  These should be super easy to use in Python with linspaces and ranges (which are inclusive at both ends by default, unlike ranges):
```python
import numpy as np
# ... controller setup  ...
controller.step('GetInteractablePositions'
    objectId=<MY_OBJECT_ID>,
    rotations=range(0, 360, 10),
    horizons=np.linspace(-30, 60, 20),
    standings=[True]
)
```

The soon to be deprecated `PositionsFromWhichItemIsInteractable` used to take `positions` as a position filter but didn't allow for any others. (It also took `horizon` as a parameter, which worked like a horizon incremented).

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6936,129 +6936,149 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
-#endif
+            #if UNITY_EDITOR
+                Debug.Log(validAgentPoses["x"].Count);
+                Debug.Log(validAgentPoses["x"]);
+            #endif
+
+            actionFinished(true, validAgentPoses);
+        }
+
+        [ObsoleteAttribute(message: "This action is deprecated. Call GetInteractablePositions instead.", error: false)]
```
mattdeitke, line 7066: This is super useful with a nice IDE (i.e., vscode). It basically gives you an inline warning that the method you're using is deprecated.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6936,129 +6936,149 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
+            // iterate over each reasonable agent pose
+            foreach (float horizon in horizons) {
+                // recall that horizon=60 is look down 60 degrees and horizon=-30 is look up 30 degrees
+                if (horizon > maxDownwardLookAngle || horizon < -maxUpwardLookAngle) {
+                    errorMessage = $"Each horizon must be in [{-maxUpwardLookAngle}:{maxDownwardLookAngle}]";
+                    actionFinished(success: false);
+                }
+                m_Camera.transform.localEulerAngles = new Vector3(horizon, 0f, 0f);
+
+                foreach (bool standing in standings) {
```
mattdeitke, line 7023: Note that we always treat `standing` as a boolean, instead of a float (0 or 1) because that is what's passed into TeleportFull and is most consistent with the rest of the physics controller.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6936,129 +6936,149 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
+            }
+
+            // populate the rotations based on rotateStepDegrees
+            if (rotations == null) {
+                rotations = new float[(int) (360 / rotateStepDegrees)];
+                int i = 0;
+                for (float rotation = 0; rotation < 360; rotation += rotateStepDegrees) {
+                    rotations[i++] = rotation;
+                }
+            }
```
mattdeitke, line 6989: Rotations are now dynamically set based on `rotateStepDegrees`. This has caused several silent bugs in the past when using non-90 degree rotations.

Lucaweihs: A few thoughts:

* What if I teleport my agent to a rotation that's not on the standard `[0, rotateStepDegrees, 2*rotateStepDegrees, ...]` set of rotations. Does it make more sense to have these rotations be relative this non-standard rotation, i.e. do something like
```
offset = cur_rotation % rotateStepDegrees
rotations = [offset, offset + rotateStepDegrees, 2 * rotateStepDegrees, ...]
```
* It looks like `rotateStepDegrees` is allowed to be any possible floating point. If I pick `rotateStepDegrees=any irrational number` then the set `{[0, rotateStepDegrees % 360, 2 * rotateStepDegrees  % 360, ...}` will never repeat, thus does it really make sense to limit the rotations in the way you are? Perhaps we should only allow this if `(360 / rotateStepDegrees) is an integer and otherwise throw an error.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6936,129 +6936,148 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
 
             // Don't want to consider all positions in the scene, just those from which the object
             // is plausibly visible. The following computes a "fudgeFactor" (radius of the object)
             // which is then used to filter the set of all reachable positions to just those plausible positions.
             Bounds objectBounds = new Bounds(
-                new Vector3(float.PositiveInfinity, float.PositiveInfinity, float.PositiveInfinity),
-                new Vector3(-float.PositiveInfinity, -float.PositiveInfinity, -float.PositiveInfinity)
+                center: new Vector3(float.PositiveInfinity, float.PositiveInfinity, float.PositiveInfinity),
+                size: new Vector3(-float.PositiveInfinity, -float.PositiveInfinity, -float.PositiveInfinity)
```
Lucaweihs, line 6996: With you writing out the parameters I'm suddenly surprised that this works. I know the intention is to have them be effectively undefined so that they become defined by the first call to `Encapsulate` but this does seem weird. Is it possible to have a small unit test that this behavior works correctly?

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6936,129 +6936,148 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
+            foreach (float horizon in horizons) {
+                // recall that horizon=60 is look down 60 degrees and horizon=-30 is look up 30 degrees
+                if (horizon > maxDownwardLookAngle || horizon < -maxUpwardLookAngle) {
+                    errorMessage = $"Each horizon must be in [{-maxUpwardLookAngle}:{maxDownwardLookAngle}]";
+                    actionFinished(success: false);
+                }
+                m_Camera.transform.localEulerAngles = new Vector3(horizon, 0f, 0f);
+
+                foreach (bool standing in standings) {
+                    if (standing) stand(); else crouch();
```
Lucaweihs, line 7024: Let's expand this to the full 
```
if (standing) {
    stand();
} else {
    crouch();
}
```

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6936,129 +6936,148 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
                 }
             }
 
-            if (wasStanding) {
-                stand();
-            } else {
-                crouch();
-            }
+            // restore old agent pose
+            if (wasStanding) stand(); else crouch();
```
Lucaweihs, line 7050: Again here. I understand the desire to save space but I consider these bad style as it's very easy to break things without clear code blocks.

Lucaweihs (changes requested):
Minor changes but otherwise LGTM.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: Yeah, will try it. Also was kinda surprised that these were the parameters on bounds, haha.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: For 1) I like the idea and would have the agent go through the full set of rotations from [offset : 360 + offset).

For instance:
* If `rotateStepDegrees=10` and `offset=70`, then the paths would be `[70, 80, ..., 400, 410, 420]`.
* If `rotateStepDegrees=90` and `offset=10`, then the paths would be `[10, 100, 190, 280]`

For 2) Seems reasonable to just give an error, since, for irrational rotations, users can override the `rotations` if they have something in mind.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2804,7 +2804,7 @@ float allowedError
             SimObjPhysics soInHand = ItemInHand.GetComponent<SimObjPhysics>();
 
             Vector3 handObjPosRelAgent =
-                Quaternion.Euler(0, rotation - transform.rotation.y, 0) *
+                Quaternion.Euler(0, rotation - transform.eulerAngles.y, 0) *
```
mattdeitke, line 2807: This is unrelated to this PR, but a bug that I discovered that uses `transform.rotation.y` instead of `transform.eulerAngles.y`, which are completely different and results in a silent bug.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -1513,7 +1513,7 @@ private bool checkForUpDownAngleLimit(string direction, float degrees)
                     return;
                 }
                 Quaternion oldRotation = sop.transform.rotation;
-                sop.transform.rotation = Quaternion.Euler(new Vector3(0.0f, (float) Math.Round(sop.transform.rotation.y + 90f % 360), 0.0f));;
+                sop.transform.rotation = Quaternion.Euler(new Vector3(0.0f, (float) Math.Round((sop.transform.eulerAngles.y + 90f) % 360), 0.0f));;
```
mattdeitke, line 1516: Same as above:
> This is unrelated to this PR, but a bug that I discovered that uses `transform.rotation.y` instead of `transform.eulerAngles.y`, which are completely different and results in a silent bug.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6936,109 +6926,184 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
 
             return allVisible;
-
         }
 
-        public void PositionsFromWhichItemIsInteractable(ServerAction action) {
+        // returns null on failure.
+        // @positions/@rotations/@horizons/@standings are used to override all possible values the agent
+        // may encounter with basic agent navigation commands (excluding teleport).
+        private List<Dictionary<string, object>> getInteractablePoses(
```
mattdeitke, line 6934: We are now returning `List<Dictionary<string, object>` since that is easiest to use with python. Here, the new syntax is:
```python
poses = controller.step('GetInteractablePoses', objectId=<>).metadata['actionReturn']
controller.step('TeleportFull', **poses[0])
```
instead of the old long equivalent
```python
poses = controller.step('PositionsFromWhichItemIsInteractable', objectId=<>).metadata['actionReturn']
i = 0
controller.step('TeleportFull',
    x=poses[i]['x'], y=poses[i]['y'], z=poses[i]['z'],
    horizon=poses[i]['horizon'],
    standing=bool(poses[i]['standing']),
    rotation=poses[i]['rotation']
)
```
Note that the latter still works for backward compatibility, but is deprecated!

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -7051,80 +7116,98 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
+                    if (key == "standing") {
+                        // standing is converted from true => 1 to false => 0, for backwards compatibility
+                        d[key].Add((bool) pose[key] ? 1 : 0);
+                    } else {
+                        // all other keys have float outputs
+                        d[key].Add((float) pose[key]);
+                    }
+                }
+            }
+            actionFinished(true, d);
```
mattdeitke, line 7187: Here, we convert from the cleaner output of `GetInteractablePoses` to the old, backwards compatible, output from `PositionsFromWhichItemIsInteractable`.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -7051,80 +7116,98 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
+                markActionFinished: false,
+                maxPoses: maxPoses
+            );
+
+            // object id might have been invalid, causing failure
+            if (markActionFinished) {
+                actionFinished(success: interactablePoses != null);
+            }
+            return interactablePoses == null ? 0 : interactablePoses.Count;
+        }
```
mattdeitke, line 7206: This action copied nearly the same functionality of `getInteractablePoses`, but returned the number instead of each pose. Here, we just modulate its functionality.

Note (cc: @Lucaweihs), this did not check for any other horizons other than 0, by default. I can have it check for others, if necessary, but I don't know the intention of only using a single horizon here.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: A test has been added (and passes)

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -7051,80 +7116,98 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
-            Debug.Log(4 * 2 * positions.Length);
-            Debug.Log(numTimesVisible);
-#endif
+            return validAgentPoses;
+        }
 
-            return numTimesVisible;
+        // Get the poses with which the agent can interact with 'objectId'
+        // @rotations: if rotation is not specified, we use rotateStepDegrees, which results in [0, 90, 180, 270] by default.
+        public void GetInteractablePoses(
```
mattdeitke, line 7133: Note that I have changed this action from `GetInteractablePositions` to `GetInteractablePoses` because it's not only returning and checking for `x/y/z` positions, rather, it checks for the entire pose of the agent.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Good catch!

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
Lucaweihs: This isn't correct I believe. The old code used whatever horizon you were currently using, not 0.

mattdeitke: That's true. I guess I just figured that these tasks never looked up or down, and hence the horizon was always 0.

But it has been updated to use the current horizon.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6795,51 +6794,25 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
...
             int maxPoses = int.MaxValue  // works like infinity
         ) {
-            if (360 % rotateStepDegrees != 0) {
-                errorMessage = "360 % rotateStepDegrees must be 0, unless 'rotations: float[]' is overwritten.";
-                if (markActionFinished) {
-                    actionFinished(success: false);
-                }
-                return null;
+            if (360 % rotateStepDegrees != 0 && rotations != null) {
+                throw new InvalidOperationException("360 % rotateStepDegrees must be 0, unless 'rotations: float[]' is overwritten.");
```
Lucaweihs, line 6798: Unnecessary for now but in future PR's it would be great if you could add some more information to these error message using string interpolation. I.e.
```
throw new InvalidOperationException($"360 % rotateStepDegrees (360 % {rotateStepDegrees} != 0) must be 0, unless 'rotations: float[]' is overwritten.");
```

Lucaweihs (approved):
LGTM, added one thought that can be ignored for now.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: Agreed. Though was an easy update, so I fixed it.

ekolve:
Can you add a test for each of the following cases?

* GetInteractablePoses for object that is in a cabinet or fridge and thus can't be interacted with? (assert poses==0)
* GetInteractablePoses for object that can be interacted with, but limit separately positions, horizons, rotations and standings to result in 0 poses.  For example, if an agent has to be in rotation 0 or 90, limit rotations to 180 and 270.
* GetInteractablePoses with a maxDistance passed in, assert some know number of poses.

It doesn't appear to be possible to pass in or use maxPoses. 1) Is that intentional? 2) was this feature requested by someone? I don't understand why it would be useful to have just say 100 arbitrary poses.

mattdeitke:
I just parameterized `maxPoses` inside of `GetInteractivePoses`. It's useful to assert that an object is interactable, without getting all the interactive positions (i.e., `maxPoses=1`).

I could also see it being useful in the case of, say, I want to build a test set that contains 50 different starting random poses such that this object is interactible from that pose from the start. Instead of getting all poses, and then randomly permuting them to choose a starting position, one can just use the 50 returned poses.

ekolve (approved):

PR #571, opened 2021-01-13
Open Object Enhancements!

mattdeitke:
This PR makes several enhancements to opening and closing objects:

* The ability to reopen an object at different openness values. Currently, to change the openness, after opening an object, users must fully close the object, and then reopen it at a different openness. This is unintuitive, requires multiple actions, and has several side effects discussed at the bottom.
* Adds new `openness` parameter, which works like `moveMagnitude` used to work, except corrects the case where `openness=0` corresponds to fully closed (instead of fully opened) and is more intuitively named. Note that `moveMagnitude` is still fully backward compatible, but now depreciated.
* Treats `CloseObject` as a special case of OpenObject with `openness=0` (instead of giving it its own functionality and checks).
* Sets `isOpen` dynamically, based on if the openness is non-0. Currently, it was set only when an object was closed.
* All actions relying on `OpenObject` and `CloseObject` have also been updated.
* Changes `openPercent` to `openness` in the metadata, since it's not a percentage as it's scaled from [0:100].

Note that no changes were made to the specifics of _how_ objects open and close. This was mostly just a big code refactor to support these updates.

---

**Testing**. These actions have tested successfully for every major open movement type:
* Rotate: FloorPlan1's fridge
* Position: FloorPlan1's drawers
* ScaleZ: FloorPlan401's shower curtains
* ScaleY: FloorPlan18's blinds

(Note that every shower curtain used ScaleZ and every blind I saw used ScaleY. I could not find any objects that open from ScaleX, but it should follow trivially if ScaleZ works.)

On failure, it has also been successfully tested to reset to the start openness, in the case where the agent is colliding with the target openness. The isOpen property has also been successfully tested in this case.

A new test was also added (and passed), which checks that:
1. The metadata updates correctly,
2. Consecutive `OpenObject` calls, with different openness values, work correctly.
3. CloseObject works correctly.

---

Side effects of the current setup:

* Objects that move in unison with the parent object move more than they supposed to (and potentially in the wrong direction)
* There is no guarantee that an object can reopen to a given openness, and may end up fully closed after the close -> reopen, if is blocking reopen's openness, but not blocking the current openness.
* Objects might not be able to fully close, but they can partially close or be reopened.
* It requires 2 actions, instead of 1, to change the openness.

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -664,34 +664,12 @@ class Controller(object):
         if 'AI2THOR_VISIBILITY_DISTANCE' in os.environ:
             action['visibilityDistance'] = float(os.environ['AI2THOR_VISIBILITY_DISTANCE'])
 
-        should_fail = False
         self.last_action = action
-
-        if ('objectId' in action and (action['action'] == 'OpenObject' or action['action'] == 'CloseObject')):
-
-            force_visible = action.get('forceVisible', False)
-            agent_id = action.get('agentId', 0)
-            instance_detections2D = self.last_event.events[agent_id].instance_detections2D
-            if not force_visible and instance_detections2D and action['objectId'] not in instance_detections2D:
-                should_fail = True
-
-            obj_metadata = self.last_event.events[agent_id].get_object(action['objectId'])
-            if obj_metadata is None or obj_metadata['isOpen'] == (action['action'] == 'OpenObject'):
-                should_fail = True
-
 
         rotation = action.get('rotation')
         if rotation is not None and type(rotation) != dict:
             action['rotation'] = {}
             action['rotation']['y'] = rotation
-
-        if should_fail:
-            new_event = copy.deepcopy(self.last_event)
-            new_event.metadata['lastActionSuccess'] = False
-            new_event.metadata['errorMessage'] = "FAILURE SHORT-CIRCUITED IN controller.py!!"
-            self.last_event = new_event
-            return new_event
-
 
         self.server.send(action)
         self.last_event = self.server.receive()
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -386,6 +386,8 @@ class Controller(object):
         self.unity_pid = None
         self.docker_enabled = docker_enabled
         self.container_id = None
+        self.width = width
+        self.height = height
 
         self.last_event = None
         self._scenes_in_build = None
@@ -500,7 +502,7 @@ class Controller(object):
 
         return self._scenes_in_build
 
-    def reset(self, scene='FloorPlan_Train1_1'):
+    def reset(self, scene='FloorPlan_Train1_1', **init_params):
         if re.match(r'^FloorPlan[0-9]+$', scene):
             scene = scene + "_physics"
 
@@ -519,7 +521,24 @@ class Controller(object):
 
         self.server.send(dict(action='Reset', sceneName=scene, sequenceId=0))
         self.last_event = self.server.receive()
-        
+
+        # update the initialization parameters
+        init_params = init_params.copy()
+
+        # width and height are updates in 'ChangeResolution', not 'Initialize'
+        if ('width' in init_params and  init_params['width'] != self.width) or (
+            'height' in init_params and init_params['height'] != self.height
+        ):
+            if 'width' in init_params:
+                self.width = init_params['width']
+                del init_params['width']
+            if 'height' in init_params:
+                self.height = init_params['height']
+                del init_params['height']
+            self.step(action='ChangeResolution', x=self.width, y=self.height)
+
+        # updates the initialization parameters
+        self.initialization_parameters.update(init_params)
         self.last_event = self.step(action='Initialize', **self.initialization_parameters)
 
         return self.last_event
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -688,6 +688,7 @@ class Controller(object):
         if should_fail:
             new_event = copy.deepcopy(self.last_event)
             new_event.metadata['lastActionSuccess'] = False
+            new_event.metadata['errorMessage'] = "FAILURE SHORT-CIRCUITED IN controller.py!!"
             self.last_event = new_event
             return new_event
 
--- a/ai2thor/server.py
+++ b/ai2thor/server.py
@@ -334,19 +334,22 @@ class Server(object):
                 if key in files:
                     image_mapping[key](files[key][i])
 
-            third_party_image_mapping = dict(
-                image=e.add_image,
-                image_thirdParty_depth=lambda x: e.add_third_party_image_depth(
+            third_party_image_mapping = {
+                # if we want to convert this param to underscores in Unity, we will need to 
+                # keep the mapping with the dash for bacwkwards compatibility with older
+                # Unity builds
+                'image-thirdParty-camera':e.add_third_party_camera_image,
+                'image_thirdParty_depth':lambda x: e.add_third_party_image_depth(
                     x,
                     depth_format=self.depth_format,
                     camera_near_plane=self.camera_near_plane,
                     camera_far_plane=self.camera_far_plane
                 ),
-                image_thirdParty_image_ids=e.add_third_party_image_ids,
-                image_thirdParty_classes=e.add_third_party_image_classes,
-                image_thirdParty_normals=e.add_third_party_image_normals,
-                image_thirdParty_flows=e.add_third_party_image_flows
-            )
+                'image_thirdParty_image_ids':e.add_third_party_image_ids,
+                'image_thirdParty_classes':e.add_third_party_image_classes,
+                'image_thirdParty_normals':e.add_third_party_image_normals,
+                'image_thirdParty_flows':e.add_third_party_image_flows
+            }
 
             if a['thirdPartyCameras'] is not None:
                 for ti, t in enumerate(a['thirdPartyCameras']):
@@ -360,10 +363,4 @@ class Server(object):
         else:
             self.last_event = event = events[0]
 
-        for img in files.get('image-thirdParty-camera', []):
-            self.last_event.add_third_party_camera_image(img)
-
         return self.last_event
-
-
-
--- a/ai2thor/server.py
+++ b/ai2thor/server.py
@@ -90,6 +90,7 @@ class Event(object):
         self.depth_frame = None
         self.normals_frame = None
         self.flow_frame = None
+        self.unity_proc = None
 
         self.color_to_object_id = {}
         self.object_id_to_color = {}
--- a/ai2thor/server.py
+++ b/ai2thor/server.py
@@ -90,7 +90,6 @@ class Event(object):
         self.depth_frame = None
         self.normals_frame = None
         self.flow_frame = None
-        self.unity_proc = None
 
         self.color_to_object_id = {}
         self.object_id_to_color = {}
@@ -296,6 +295,7 @@ class Server(object):
         self.sequence_id = 0
         self.started = False
         self.client_token = None
+        self.unity_proc = None
 
         if add_depth_noise:
             assert width == height,\
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -67,6 +67,21 @@ def test_small_aspect():
     event = controller.step(dict(action='Initialize', gridSize=0.25))
     assert event.frame.shape == (64, 128, 3)
 
+def test_reset():
+    controller = build_controller()
+    width = 520
+    height = 310
+    event = controller.reset(scene='FloorPlan28', width=width, height=height, renderDepthImage=True)
+    assert event.frame.shape == (height, width, 3), "RGB frame dimensions are wrong!"
+    assert event.depth_frame is not None, 'depth frame should have rendered!'
+    assert event.depth_frame.shape == (height, width), "depth frame dimensions are wrong!"
+
+    width = 300
+    height = 300
+    event = controller.reset(scene='FloorPlan28', width=width, height=height, renderDepthImage=False)
+    assert event.depth_frame is None, "depth frame shouldn't have rendered!"
+    assert event.frame.shape == (height, width, 3), "RGB frame dimensions are wrong!"
+
 def test_fast_emit():
     fast_controller = build_controller(server_class=FifoServer, fastActionEmit=True)
     event = fast_controller.step(dict(action='RotateRight'))
@@ -90,7 +105,6 @@ def test_fast_emit_disabled(controller):
 
 @pytest.mark.parametrize("controller", [wsgi_controller, fifo_controller])
 def test_lookdown(controller):
-
     e = controller.step(dict(action='RotateLook', rotation=0, horizon=0))
     position = controller.last_event.metadata['agent']['position']
     horizon = controller.last_event.metadata['agent']['cameraHorizon']
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -55,6 +55,16 @@ def test_multi_agent_with_third_party_camera():
     event = controller.step(dict(action='AddThirdPartyCamera', rotation=dict(x=0, y=0, z=90), position=dict(x=-1.0, z=-2.0, y=1.0)))
     assert not np.all(controller.last_event.events[1].frame == controller.last_event.events[0].frame)
 
+# Issue #526 thirdPartyCamera hanging without correct keys in FifoServer FormMap
+def test_third_party_camera_with_image_synthesis():
+    controller = build_controller(server_class=FifoServer, renderObjectImage=True, renderDepthImage=True, renderClassImage=True)
+    event = controller.step(dict(action='AddThirdPartyCamera', rotation=dict(x=0, y=0, z=90), position=dict(x=-1.0, z=-2.0, y=1.0)))
+    assert len(event.third_party_depth_frames) == 1
+    assert len(event.third_party_class_segmentation_frames) == 1
+    assert len(event.third_party_camera_frames) == 1
+    assert len(event.third_party_instance_segmentation_frames) == 1
+
+
 def test_rectangle_aspect():
     controller = build_controller(width=600, height=300)
     controller.reset('FloorPlan28')
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -185,8 +185,6 @@ def test_simobj_filter(controller):
 
 @pytest.mark.parametrize("controller", [wsgi_controller, fifo_controller])
 def test_add_third_party_camera(controller):
-
-
     expectedPosition = dict(x=1.2, y=2.3, z=3.4)
     expectedRotation = dict(x=30, y=40, z=50)
     expectedFieldOfView = 45.0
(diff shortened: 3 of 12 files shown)


ai2thor/server.py
```diff
@@ -295,6 +295,7 @@ def __init__(self, width, height, depth_format=DepthFormat.Meters, add_depth_noi
         self.sequence_id = 0
         self.started = False
         self.client_token = None
+        self.unity_proc = None
```
mattdeitke, line 298: This was added to allow one to debug with the Unity editor open from Python. That is,
```python
from ai2thor.wsgi_server import WsgiServer
Controller(start_unity=False, port=8200, server_class=WsgiServer)
```

ai2thor/tests/test_unity.py
```diff
@@ -276,6 +274,23 @@ def test_teleport(controller):
     position = controller.last_event.metadata['agent']['position']
     assert_near(position, dict(x=-2.0, z=-2.5, y=0.901))
 
+@pytest.mark.parametrize("controller", [wsgi_controller, fifo_controller])
+def test_open(controller):
```
mattdeitke, line 278: This tests trying consecutive openness values in [0.5, 0.7, 0]. Here, passing in `openness=0` is exactly what is done with `CloseObject`.

ai2thor/controller.py
```diff
@@ -664,34 +664,13 @@ def step(self, action=None, **action_args):
         if 'AI2THOR_VISIBILITY_DISTANCE' in os.environ:
             action['visibilityDistance'] = float(os.environ['AI2THOR_VISIBILITY_DISTANCE'])
 
-        should_fail = False
         self.last_action = action
 
-        if ('objectId' in action and (action['action'] == 'OpenObject' or action['action'] == 'CloseObject')):
```
mattdeitke, line 670: This removes a "short" circuiting call that effectively fails if an opened object tries to be reopened. It was previously executed after _every single step_, regardless of the action, and had a marginal effect, at best.

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -97,529 +78,218 @@ void Start ()
...
-                        "position", lerpToPosition,
-                        "islocal", true,
-                        "time", animationTime,
-                        "easetype", "linear"));
-                    }
-
-				}
+        startOpenness = currentOpenPercentage;
+        for (int i = 0; i < MovingParts.Length; i++) {
+            Hashtable args = new Hashtable() {
```
mattdeitke, line 152: On the left, notice that all of these animations were nearly identical, except for passing in `rotation`, `position`, or `scale`. This `args` parameter just stores the identical arguments up-front, instead of explicitly writing them out for every single different open movement animation.

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -97,529 +78,218 @@ void Start ()
...
-                        "islocal", true,
-                        "time", animationTime,
-                        "easetype", "linear"));
-                    }
-                }
-			}
-
+    // note: reset can interrupt the Interact() itween call because
+    // it will start a new set of tweens before onComplete is called from Interact()... it seems
+    public void Reset() {
```
mattdeitke, line 274: `Reset`, which is called within `OnTriggerEnter` used to also treat Open as a Toggle, where if the object is open, let's close it, and if the object is closed, let's open it. It duplicated much of the functionality as `Interact` (which animates the opening of the objects).

Now, on failure, we just call Interact, passing in the `startOpenness` before the failed action started.

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -97,529 +78,218 @@ void Start ()
...
-                        "oncompleteparams", openPercentage));
-                    }
-
-                    else
-                    iTween.ScaleTo(MovingParts[i], iTween.Hash(
-                    "scale", new Vector3(openPositions[i].x, closedPositions[i].y + (openPositions[i].y - closedPositions[i].y) * openPercentage, openPositions[i].z),
-                    "islocal", true,
-                    "time", animationTime,
-                    "easetype", "linear"));
+            // let's open the object!
```
mattdeitke, line 164: Note: all of these animations are _exactly_ the same as how they started. They are just refactored with `args` so they aren't as repetitive.

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -97,529 +78,218 @@ void Start ()
...
-    }
 
-    private void setisOpen(float openPercentage)
-	{
-        //print("set is open happening");
-		isOpen = !isOpen;
+        setIsOpen(openPercentage: openPercentage);
+    }
 
+    private void setIsOpen(float openPercentage) {
```
mattdeitke, line 202: Here, instead of setting `isOpen` in the metadata to `!isOpen` (i.e., the toggling of open and close), we dynamically set it to `openPercentage != 0`.

ai2thor/controller.py
Lucaweihs: This short circuit also results in a failure if the `objectId` is not in the set of `instance_dections2D` (if `instance_detections2D` is available). Any idea why that was necessary (@winthos, @ekolve)? Are there really cases where an object is visible but the instance isn't visible? In either case, the behavior should probably not change depending on if `instance_detections2D` being available so I support this removal.

mattdeitke: (Also, a _terrible_ pain to debug, when everything works perfectly in Unity, but not in Python).

ai2thor/tests/test_unity.py
Lucaweihs: LGTM.

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -42,53 +41,35 @@ protected enum MovementType { Slide, Rotate, ScaleX, ScaleY, ScaleZ};
...
-    #if UNITY_EDITOR
-    void OnEnable ()
-    {
-		//debug check for missing CanOpen property
-        if (!gameObject.GetComponent<SimObjPhysics>().DoesThisObjectHaveThisSecondaryProperty(SimObjSecondaryProperty.CanOpen))
-        {
-            Debug.LogError(gameObject.transform.name + " is missing the Secondary Property CanOpen!");
-        }
-    }
-    #endif
```
Lucaweihs, line 77: Can you comment on why this was removed?

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -7957,23 +7805,40 @@ public Vector3 GetSurfacePointBelowPosition(Vector3 position)
...
-        public void RandomlyOpenCloseObjects(int randomSeed = 0, bool simplifyPhysics = false) {
-            System.Random rnd = new System.Random(randomSeed);
-            List<CanOpen_Object> toInteractWith = new List<CanOpen_Object>();
+        // @pOpen is the probability of opening an openable object.
+        // @randOpenness specifies if the openness for each opened object should be random, between 0% : 100%, or always 100%.
+        public void RandomlyOpenCloseObjects(
+                int? randomSeed = null,
+                bool simplifyPhysics = false,
+                float pOpen = 0.5f,
+                bool randOpenness = true
```
mattdeitke, line 7814: Note that there are 2 potentially breaking changes with this beta action:
* `randomSeed` is not set to 0 by default. Hence, if it is not passed in, the objects that are opened are actually random in the same scene.
* `randOpenness` specifies if the newly opened openness values should be randomized in `[0:1]` or if every openness should be exactly 1.

Note that if there is a random seed and `randOpenness == True`, each chosen openness value will also be sampled from a random seed.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5124,106 +5092,99 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
...
-                    actionFinished(false);
-                    return;
-                }
+        // syntactic sugar for open object with openness = 0.
+        public void CloseObject(
+            string objectId,
+            bool forceAction = false
+        ) {
+            OpenObject(objectId: objectId, forceAction: forceAction, openness: 0);
+        }
```
mattdeitke, line 5179: `CloseObject` is now treated as syntactic sugar for `OpenObject`, with `openness` being 0. Previously, it ran its own checks that were identical to the checks ran within `OpenObject`.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5124,106 +5092,99 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
...
+            bool forceAction
+        ) {
+            // no target object specified, so instead try and use x/y screen coordinates
+            SimObjPhysics target = null;
+            if(!ScreenToWorldTarget((float) x, (float) y, ref target, !forceAction)) {
+                // error message is set insice ScreenToWorldTarget
+                return null;
+            }
+            return target;
+        }
```
mattdeitke, line 5171: Eventually `getTargetObject` will be used for every interaction action that targets an object by either `x/y` or `objectId`. Currently, most actions repeat this code in exactly the same manner. Adding this to every action, however, will be a future PR.

Note that above this, it is overloaded to also take in an `objectId`.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5124,106 +5092,99 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
...
+        // sim object that they target.
+        private SimObjPhysics getTargetObject(
+            float x,
+            float y,
+            bool forceAction
+        ) {
+            // no target object specified, so instead try and use x/y screen coordinates
+            SimObjPhysics target = null;
+            if(!ScreenToWorldTarget((float) x, (float) y, ref target, !forceAction)) {
+                // error message is set insice ScreenToWorldTarget
```
Lucaweihs, line 5167: `incise` -> `inside`?

ai2thor/controller.py
winthos: I can't think of a case where an object is visible but the instance would not be visible. I think this should be fine to remove. I can also confirm it was a terrible pain to debug.

unity/Assets/Scripts/CanOpen_Object.cs
winthos: looks good to me

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5329,87 +5289,96 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
...
+
                 foreach (string objectId in target.GetAllSimObjectsInReceptacleTriggersByObjectID()) {
-                    SimObjPhysics toReParent = physicsSceneManager.ObjectIdToSimObjPhysics[objectId];
-                    objectIdToOldParent[toReParent.ObjectID] = toReParent.transform.parent;
-                    toReParent.transform.parent = coo.transform;
-                    toReParent.GetComponent<Rigidbody>().isKinematic = true;
+                    SimObjPhysics simObj = physicsSceneManager.ObjectIdToSimObjPhysics[objectId];
+                    objectIdToOldParent[objectId] = simObj.transform.parent;
+                    simObj.transform.parent = openableObject.transform;
+                    simObj.GetComponent<Rigidbody>().isKinematic = true;
```
Lucaweihs, line 5350: I prefer the `toReParent` name.

unity/Assets/Scripts/CanOpen_Object.cs
winthos: I think there is potential to change this behavior to instead reset to the `startOpenness` we can have an option to simply stop the animation wherever it was when `Reset` is called.

I don't think this change should be in this PR, but I can see this sort of functionality being useful when we eventually implement open/close with the Mid Level Arm. I think the behavior in that case when an arm tries to open something is to "open" it via the animation, but if the arm hits something and can no longer follow through with the opening motion, the openable object stops the animation.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5421,11 +5390,9 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
                 }
             }
 
-            if (!success) {
-                errorMessage = "Object failed to open/close successfully.";
+            if (markActionFinished) {
+                actionFinished(errorMessage.Length == 0);
             }
```
Lucaweihs, line 5395: This might be slightly dangerous if some method that calls this had already set the `errorMessage` to something.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: Looks good. Should be a pretty straightforward future PR to swap this in to the rest of the actions.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5829,125 +5727,75 @@ protected IEnumerator ToggleAndWait(CanToggleOnOff ctof)
...
-            if(action.objectId == null)
-            {
-                if(!ScreenToWorldTarget(action.x, action.y, ref target, !action.forceAction))
-                {
-                    //error message is set insice ScreenToWorldTarget
-                    actionFinished(false);
-                    return;
-                }
+        // private helper used with OpenObject commands
+        private void openObject(
```
Lucaweihs, line 5731: I don't like that this helper function controls whether or not the action is finished, what if I want to call this as part of some other method in which I want to do something after having opened the object?

Lucaweihs (commented):
Lots of great changes. I have some minor comments here and there.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: Looks good. This also helps with backwards compatibility right? I think there is a discussion to be had to completely remove CloseObject and simply have people use OpenObject with `openness` at 0, but I'm not sure which is more intuitive.

unity/Assets/Scripts/CanOpen_Object.cs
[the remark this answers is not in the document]
mattdeitke: No. This was just moved to an all-encompassing #if UNITY_EDITOR block, so that there aren't repeated UNITY_EDITOR blocks within the same context.

https://github.com/allenai/ai2thor/blob/openness/unity/Assets/Scripts/CanOpen_Object.cs#L102

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -97,529 +78,218 @@ void Start ()
...
     //         openPercentage = val;
     //         return true;
-    //     }
-
-    //     else
-    //     {
+    //     } else {
     //         return false;
     //     }
     // }
```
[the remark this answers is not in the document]
mattdeitke: I left it as a general rule not to delete any commented code, but I'm happy to remove them.

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -97,529 +78,218 @@ void Start ()
...
-                        "time", animationTime,
-                        "easetype", "linear"));
-                    }
-                }
-			}
-
+    // note: reset can interrupt the Interact() itween call because
+    // it will start a new set of tweens before onComplete is called from Interact()... it seems
+    public void Reset() {
+        if (!canReset) {
```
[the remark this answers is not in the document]
winthos: I think this the correct reading. The animating object is reset to the `startOpenness` when it hits something. The `canReset` toggle itself is so that upon multiple instances of `OnTrigger` being called, it won't constantly try and reset back and forth forever in an endless loop of trying to resolve to the "last position."

So to that end, the object is only able to reset again once it has fully reset, hence `canReset`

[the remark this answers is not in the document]
winthos: or maybe something like `hasReset` might be easier to understand?

winthos (commented):
These are some much needed changes to the open/close functionality. Good work trying to parse my original jumble of code!

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -97,529 +78,218 @@ void Start ()
...
+            Hashtable args = new Hashtable() {
+                {"islocal", true},
+                {"time", animationTime},
+                {"easetype", "linear"}
+            };
+
+            // we are on the last moving part here
+            if (i == MovingParts.Length - 1) {
+                args["onCompleteTarget"] = gameObject;
+                args["oncompleteparams"] = isOpen ? 0f : openPercentage;
```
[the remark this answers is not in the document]
mattdeitke: It doesn't look like anything ever uses it (just searching for "oncompleteparams" within the entire code base, it only appears here).

Plus, all its functionality is built within `setIsOpen`, which is called at the end:
https://github.com/allenai/ai2thor/blob/openness/unity/Assets/Scripts/CanOpen_Object.cs#L199

So, I'm going to remove it. (It was only added because it was here originally).

ai2thor/controller.py
ekolve: DanielG added this - i believe the intent of all this was just to speed up Open/CloseObject.  I think the instance_detections2d check was basically a crude visibility check (i.e. fail if trying to open the object if its behind you).  All of this should be removed and our rules should be enforced in a single place.

unity/Assets/Scripts/CanOpen_Object.cs
```diff
@@ -22,11 +20,12 @@ public class CanOpen_Object : MonoBehaviour
     public float animationTime = 0.2f;
 
     [SerializeField]
-    public float currentOpenPercentage = 1.0f; //0.0 to 1.0 - percent of openPosition the object opens. 
+    public float currentOpenPercentage = 1.0f; // 0.0 to 1.0 - percent of openPosition the object opens. 
+    private float startOpenness; // used to reset on failure
```
ekolve, line 24: I'm fine with whatever you choose, but can you make the naming consistent.  Use say "currentOpenness" and "startOpenness" or "currentOpenPercentage" and "startOpenPercentage"

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
ekolve: I agree - i would prefer that we explicitly set whether actions are successful or not.  Relying on the errorMessage string may lead to erroneous failures.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: Yep: backward compatible. Agree that since `CloseObject` is so cheap now, it would add nothing to remove it.

ekolve (changes requested):
Can you add tests for the following cases?:

* Pass openness > 1.0 and < 0.0 and assert failure
* Pass moveMagnitude instead of openness to ensure backwards compatibility

unity/Assets/Scripts/CanOpen_Object.cs
mattdeitke: Yeah, that's fair. I was just using `currentOpenPercentage` as the legacy variable name, while `startOpenness` as the new name.

unity/Assets/Scripts/CanOpen_Object.cs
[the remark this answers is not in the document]
Lucaweihs: Or perhaps something like `isCurrentlyResetting`.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: Added an option with `markActionFinished`.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: changed

unity/Assets/Scripts/CanOpen_Object.cs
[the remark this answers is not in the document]
mattdeitke: used `isCurrentlyResetting` here

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: fixed

Lucaweihs (approved):
LGTM

PR #572, opened 2021-01-13
Tiny change to the `SafelyComputeNavMeshPath` function so that the `allowedError` parameter is better handled.

Lucaweihs:


ekolve (approved):

ekolve:
lgtm

PR #578, opened 2021-01-17
Using Exceptions to short circuit action failures

mattdeitke:
## Overview

This PR allows short circuiting action failures, which may terminate the action immediately (from any function call).

We also encourage error messages to be set within the invoked exception, so

```c#
if (failure) {
    actionFinished(success: false);
    errorMessage = "Failure because of XXX.";
    return;
}
```
becomes:
```c#
if (failure) {
    throw new <...Exception>("Failure because of XXX.");
}
```
In the former case, too, notice that errorMessage would never execute if `actionFinished` was refactored to throw an exception.

## Motivation

A ton of actions are functionalized. For instance, 
* `OpenObject(x, y, forceAction)` calls `getTargetObject(x, y, forceAction)` which calls `ScreenToWorldTarget(x, y, ...)`.

Previously, if `ScreenToWorldTarget` had an error, all it would do was set the `errorMessage`, but it could not end the action right there. Instead, it would have to shakily rely on `getTargetObject` and `OpenObject` to catch that there was an error by returning null.

Often, inside the null catches, less specific error messages are overwritten, because the outer functions don't know if the inner functions wrote a better error message:
```c#
PushObject(...) {
    returnedValue = getTargetObject(...);
    if (returnedValue == null) {
        errorMessage = "maybe this happened inside of getTargetObject";
        actionFinished(success: false);
        return;
    }
    // ...
}

getTargetObject(...) {
    returnedValue = ScreenToWorldTarget(...);
    if (returnedValue == null) {
        errorMessage = "maybe this happened inside of ScreenToWorldTarget";
        return;
    }
    // ...
}

ScreenToWorldTarget(...) {
    if (failure) {
        errorMessage = "specific error message";
        return null;
    }
}
```

Now, instead of all these checks, we can short-circuit the action failure and terminate immediately, which refactors to:
```c#
PushObject(...) {
    returnedValue = getTargetObject(...);
}

getTargetObject(...) {
    returnedValue = ScreenToWorldTarget(...);
}

ScreenToWorldTarget(...) {
    if (failure) {
        throw new <...Exception>("specific error message");
    }
}
```
where the action terminates immediately when the exception is thrown.

## Avoiding Early Termination

In rare cases, actions may not actually want to terminate upon failure. Here, these actions can use try/catch statements to avoid such early stoppage. For instance:
```c#
OpenAllObjects(...) {
    for each openable object
        try {
            openObject(...)
        } catch (InvalidOperationException e) { }
}
```
which will do nothing if the `InvalidOperationException` is encountered upon calling `openObject`.

## Backwards compatibility

Everything should be fully backward compatible.

## Refactoring Note

This PR doesn't clean up every possible case where short-circuiting can be applied. Rather, just a few for now, since the functionality is just being proposed.

### DIFF
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -4630,50 +4630,50 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             // #endif
         }
 
-        //used for all actions that need a sim object target
-        //instead of objectId, use screen coordinates to raycast toward potential targets
-        //will set the target object by reference if raycast is succesful
-        public bool ScreenToWorldTarget(float x, float y, ref SimObjPhysics target, bool requireWithinViewportRange)
-        {
-            //float x = action.x;
-            y = 1.0f - y; //reverse the y so that the origin (0, 0) can be passed in as the top left of the screen
-            //cast ray from screen coordinate into world space. If it hits an object
+        // used for all actions that need a sim object target
+        // instead of objectId, use screen coordinates to raycast toward potential targets
+        // will set the target object by reference if raycast is succesful
+        public bool ScreenToWorldTarget(float x, float y, ref SimObjPhysics target, bool requireWithinViewportRange) {
+            // reverse the y so that the origin (0, 0) can be passed in as the top left of the screen
+            y = 1.0f - y;
+
+            // cast ray from screen coordinate into world space. If it hits an object
             Ray ray = m_Camera.ViewportPointToRay(new Vector3(x, y, 0.0f));
             RaycastHit hit;
-            //if something was touched, actionFinished(true) always
-            if(Physics.Raycast(ray, out hit, Mathf.Infinity, 1 << 0 | 1 << 8 | 1 << 10, QueryTriggerInteraction.Ignore))
-            {
-                if(hit.transform.GetComponent<SimObjPhysics>())
-                {
-                    //wait! First check if the point hit is withing visibility bounds (camera viewport, max distance etc)
-                    //this should basically only happen if the handDistance value is too big
-                    if(requireWithinViewportRange && !CheckIfTargetPositionIsInViewportRange(hit.point))
-                    {
-                        errorMessage = "target sim object at screen coordinate: (" + x + ", " + y + ") is not within the viewport";
-                        return false;
+
+            // if something was touched, actionFinished(true) always
+            if (Physics.Raycast(ray, out hit, Mathf.Infinity, 1 << 0 | 1 << 8 | 1 << 10, QueryTriggerInteraction.Ignore)) {
+                if (hit.transform.GetComponent<SimObjPhysics>()) {
+                    // wait! First check if the point hit is withing visibility bounds (camera viewport, max distance etc)
+                    // this should basically only happen if the handDistance value is too big
+                    if (requireWithinViewportRange && !CheckIfTargetPositionIsInViewportRange(hit.point)) {
+                        actionFinished(
+                            success: false,
+                            errorMessage: "target sim object at screen coordinate: (" + x + ", " + y + ") is not within the viewport",
+                            stopActionNow: true
+                        );
                     }
 
-                    //it is within viewport, so we are good, assign as target
+                    // it is within viewport, so we are good, assign as target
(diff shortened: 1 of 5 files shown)


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4630,58 +4630,50 @@ public void PlaceHeldObject(string objectId, bool forceAction=false, bool placeS
...
-                        errorMessage = "target sim object at screen coordinate: (" + x + ", " + y + ") is not within the viewport";
-                        return false;
+
+            // if something was touched, actionFinished(true) always
+            if (Physics.Raycast(ray, out hit, Mathf.Infinity, 1 << 0 | 1 << 8 | 1 << 10, QueryTriggerInteraction.Ignore)) {
+                if (hit.transform.GetComponent<SimObjPhysics>()) {
+                    // wait! First check if the point hit is withing visibility bounds (camera viewport, max distance etc)
+                    // this should basically only happen if the handDistance value is too big
+                    if (requireWithinViewportRange && !CheckIfTargetPositionIsInViewportRange(hit.point)) {
+                        throw new InvalidOperationException($"Target sim object at screen coordinate: ({x}, {y}) is not within the viewport");
```
mattdeitke, line 4650: InvalidOperationException is C#'s built in way of handling Invalid State Exceptions.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5119,47 +5111,41 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
...
-                bool forceAction = false
-        ) {
+        // Helper method that parses objectId parameter to return the sim object that it target.
+        // The action is halted if the objectId does not appear in the scene.
+        private SimObjPhysics getTargetObject(string objectId, bool forceAction = false) {
             // an objectId was given, so find that target in the scene if it exists
             if (!physicsSceneManager.ObjectIdToSimObjPhysics.ContainsKey(objectId)) {
-                errorMessage = "Object ID appears to be invalid.";
-                return null;
+                throw new ArgumentException($"objectId: {objectId} is not the objectId on any object in the scene!");
```
mattdeitke, line 5119: Using `ArgumentException` to mean anything that is non-null and not bound to a particular numberic range (e.g., x/y must be in [0:1]).

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5119,47 +5111,41 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
...
 
             // if object is in the scene and visible, assign it to 'target'
             SimObjPhysics target = null;
             foreach (SimObjPhysics sop in VisibleSimObjs(objectId: objectId, forceVisible: forceAction)) {
                 target = sop;
             }
+
+            // target not found!
+            if (target == null) {
+                throw new NullReferenceException("Target object not found within the specified visibility.");
```
mattdeitke, line 5130: `NullReferenceExceptions` are for exceptions where the null does not occur at the argument level, but later, when something unexpectidly results in `null`.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5119,47 +5111,41 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
...
-            bool forceAction
-        ) {
-            // no target object specified, so instead try and use x/y screen coordinates
-            SimObjPhysics target = null;
-            if(!ScreenToWorldTarget((float) x, (float) y, ref target, !forceAction)) {
-                // error message is set inside ScreenToWorldTarget
-                return null;
+        private SimObjPhysics getTargetObject(float x, float y, bool forceAction) {
+            if (x < 0 || x > 1 || y < 0 || y > 1) {
+                throw new ArgumentOutOfRangeException("x/y must be in [0:1]");
```
mattdeitke, line 5140: `ArgumentOutOfRangeException` is for numeric ranges.

ekolve:
Everything looks great, could you add one unit test that forces a TargetInvocationException from the python side in test_unity.py?

ekolve (approved):

ekolve:
lgtm

PR #581, opened 2021-01-23
Execute JSON actions from Unity's debug editor field

mattdeitke:
## Overview

Given a JSON file containing a sequence of actions, placed in `unity/Assets/Scripts/DebugScripts/`, we can now execute that sequence of actions.

For example, with a file `unity/Assets/Scripts/DebugScripts/simple.json` containing:
```json
[
  {
    "action": "MoveAhead",
    "moveMagnitude": 0.5
  },
  {
    "action": "RotateRight",
    "degrees": 30
  },
  {
    "action": "RotateRight",
    "degrees": 30
  }
]
```
running
```
run simple
```
from the debug input field will execute all 3 actions.

Equivalently, we could have also called
```
run simple.json
```
which does the same thing.

## Remarks on `reset`

This doesn't currently work with `reset`, meaning that one should change scenes manually to the scene of interest. I'm a bit hesitant to support it, since:
* From the Python side, `reset` is not an action that is executed within the `controller.step(...)` context. Rather, it has its own method: `controller.reset(...)`.
* Also from Python, `reset` can also take any initialization parameters now (along with `width` and `height`).  Thus, Python's `reset` does not match Unity's `reset`, and trying to copy all of Python's reset functionality into this `run` command will make maintance harder and possibly cause silent bugs in the future.

## Debugger for Unity

It hasn't been requested, but I am strongly against adding support for, say:
* `run simple 5`  to run only the first 5 actions.
* `run simple -5` to run all the way up to, but not through, the last 5 actions.
* `run simple 5:8`  to run actions 5 through 8.

All breakpoints can, and should, be set with a [Debugger for Unity](https://marketplace.visualstudio.com/items?itemName=Unity.unity-debug). This allows for so much better debugging support, specifically with conditional breakpoints, which can be placed in any file, and only pause if an event occurs. For instance, we may put a conditional breakpoint after an `actionFinished == false`.

If anybody needs help setting up the Debugger for Unity, let me know. I've tested it here and it works perfectly!


ekolve:
Take a look at ExecuteBatch https://github.com/allenai/ai2thor/blob/ExpRoom/unity/Assets/Scripts/DebugInputField.cs#L206

If you run the action execution in a coroutine and yield between each action waiting for the next frame then a user can use the step button in the editor to step through the actions and won't require a debugger.

mattdeitke:
> 
> 
> Take a look at ExecuteBatch https://github.com/allenai/ai2thor/blob/ExpRoom/unity/Assets/Scripts/DebugInputField.cs#L206
> 
> If you run the action execution in a coroutine and yield between each action waiting for the next frame then a user can use the step button(to the right of the play/stop buttons above the main screen) in the editor to step through the actions and won't require a debugger.

I can add separate support for this, but one thing to note is that some of these debugging actions are several hundred actions long, so going one-by-one through each of them would be extremely time consuming.

mattdeitke:
Actually, I'm not able to reproduce stepping one-by-one through each action in `ExecuteBatch` from the editor. When I try, say, `slide3`, all the actions execute right away.

a similar thing happens when using `Debug.Break()`, where after executing the first action from `ProcessControlCommand`, the `Debug.Break()` command is ignored and the entire sequence of actions execute. (Note: this is not the same as using the Debugger for Unity, which works fine here, stopping after each action with a breakpoint).

ekolve (changes requested):
A few requests - 

* please move the .json DebugScripts to ai2thor/unity/debug_scripts or ai2thor/unity/scripts/debug.  Anything in the Unity special folders (https://docs.unity3d.com/Manual/SpecialFolders.html) gets imported as an asset into the asset database.  Unity also watches for any changes to these files, so there is some overhead.  These types of files should live outside the UnityEditor owned directories.

* Instead of using a base path, look at using https://docs.unity3d.com/ScriptReference/EditorUtility.OpenFilePanel.html.  You can then provide a default starting directory for users to select a json file and if they want to load a script from outside the project, they can also do that.  This makes it more convenient if they just want to try something and they have a file living in a different location and don't want to figure out what directory they need to drop the json file in for it to get executed.  As well, this eliminates any chance of typos since they either select a json file or not.

* You are right about Debug.Break.  I used the step frame feature in conjunction with python connected to the editor.  This allowed me to step through actions in python one at a time and view the changes in the editor.  I tried several times to get the Break to trigger before a command got executed, but could not.  I imagine we could come up with some way to get around the behavior but it doesn't seem like all the complexity would improve upon just breaking with a debugger which will actually stop at a breakpoint.  

* The primary reason for using ExecuteBatch + StartCoroutine is that the current way (in the PR) that actions are executed the Unity lifecycle is being skipped for each action.  The default way that debug actions are executed are through the OnEndEdit in the text field. When the action finishes, the frame is rendered and all the steps in the lifecycle are executed (https://docs.unity3d.com/Manual/ExecutionOrder.html).  The most important set of events in the lifecycle for us is the Physics events.  If the actions are executed one after another, the loop is skipped.  A few areas where this matters are when you say pick an object up and drop it.  We have logic that waits for the object to settle by allowing one or more frames to pass and then marking the action finished.  Another is more recently we relied on Unity's built in animator to move the arm.  The way it worked was that it would only calculate the next position of the joints once per frame - so you had to let a frame pass in order for the arm to move.  You don't have to use the same ExecuteBatch function, if you want to create a new one that takes a JArray or List<DynamicServerAction> and does a similar thing that is fine with me.

ekolve (approved):

ekolve:
lgtm

PR #583, opened 2021-01-26
Better Event summarization in Event.__repr__ and Event.__str__

mattdeitke:
Currently, printing or viewing the result of an `ai2thor.server.Event` is not particularly helpful, since it only shows:
```python
<ai2thor.server.Event at {hex ID in memory}>
```
This is particularly annoying in notebooks, since one then must always immediately call:
```python
event.metadata['lastActionSuccess']
event.metadata['errorMessage']
event.metadata['actionReturn']
```
to see if anything happened.

This PR now adds this summary information to the Event, so when users print or evaluate the Event, they will see something like:
```python
<ai2thor.server.Event at 0x7f213063ed30
    .metadata['lastActionSuccess'] = True
    .metadata['errorMessage'] = ""
    .metadata['actionReturn'] = None
>
```

As an added benefit, it also shows new users that there is a `.metadata['lastActionSuccess']`, `.metadata['errorMessage']`, and `.metadata['actionReturn']`.

### DIFF
--- a/ai2thor/server.py
+++ b/ai2thor/server.py
@@ -119,7 +119,7 @@ class Event(object):
         return (
             '<ai2thor.server.Event at ' + hex(id(self)) + '\n' +
             '\t.metadata["lastActionSuccess"] = ' + self.metadata['lastActionSuccess'] + '\n' +
-            '\t.metadata["errorMessage"] = ' + self.metadata['errorMessage'] + '\n' +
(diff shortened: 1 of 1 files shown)


ehsanik (commented):
This is a very helpful change and it would make debugging in notebook or using pdb much easier and more convenient.

ekolve (approved):

PR #585, opened 2021-01-28
Fixed IK NaN issue

elimvb:
This latest IK Robot Arm Controller script update should fix all NaN issues. Please re-run the relevant slide tests to confirm this is the case.

### DIFF
--- a/unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
+++ b/unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
@@ -517,17 +517,35 @@ public class IK_Robot_Arm_Controller : MonoBehaviour
             jointMeta.rootRelativePosition = FirstJoint.InverseTransformPoint(joint.position);
 
             //local rotation currently relative to immediate parent joint
-            joint.GetChild(0).localRotation.ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
-            jointMeta.localRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            if (joint.rotation != joint.GetChild(0).rotation)
+            {
+                joint.GetChild(0).localRotation.ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
+                jointMeta.localRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            }
 
+            //edge case for where angler and parent rotations are aligned, which Quaternions have trouble resolving, so we hard-code it here
+            else
+            {
+                jointMeta.localRotation = new Vector4(1, 0, 0, 0);
+            }
+            
             //world relative rotation
             joint.GetChild(0).rotation.ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
             jointMeta.rotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
 
             //rotation relative to root joint/agent
             //root forward and agent forward are always the same
-            Quaternion.Euler(FirstJoint.InverseTransformDirection(joint.GetChild(0).eulerAngles)).ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
-            jointMeta.rootRelativeRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            if (FirstJoint.rotation != joint.GetChild(0).rotation)
+            {
+                Quaternion.Euler(FirstJoint.InverseTransformDirection(joint.GetChild(0).eulerAngles)).ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
+                jointMeta.rootRelativeRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            }
+
+            //edge case for when angler and root rotations are aligned, which Quaternions have trouble resolving, so we hard-code it here
+            else
+            {
+                jointMeta.rootRelativeRotation = new Vector4(1, 0, 0, 0);
+            }
 
             joints.Add(jointMeta);
         }


winthos (changes requested):

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -517,17 +517,35 @@ public void DropObject()
...
+            {
+                Quaternion.Euler(FirstJoint.InverseTransformDirection(joint.GetChild(0).eulerAngles)).ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
+                jointMeta.rootRelativeRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            }
+
+            //edge case for when angler and root rotations are aligned, which Quaternions have trouble resolving, so we hard-code it here
+            else
+            {
+                jointMeta.rootRelativeRotation = new Vector4(1, 0, 0, 0);
+            }
```
mattdeitke, line 548: Small style comments, can we make this:
```c#
if (FirstJoint.rotation != joint.GetChild(0).rotation) {
    // get child to grab the angler since that is what actually changes the geometry angle
    Quaternion.Euler(FirstJoint.InverseTransformDirection(joint.GetChild(0).eulerAngles)).ToAngleAxis(out angleRot, out vectorRot);
    jointMeta.rootRelativeRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
} else {
    // edge case for when angler and root rotations are aligned, which Quaternions have trouble resolving, so we hard-code it here
    jointMeta.rootRelativeRotation = new Vector4(1, 0, 0, 0);
}
```

It scares me putting so many newlines. I feel like I might accidently add something below the `if` and above the `else`.

It's also more consistent with some newer changes in #584.


(Same thing goes for the other snippet.)

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -517,17 +517,35 @@ public void DropObject()
             jointMeta.rootRelativePosition = FirstJoint.InverseTransformPoint(joint.position);
 
             //local rotation currently relative to immediate parent joint
-            joint.GetChild(0).localRotation.ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
-            jointMeta.localRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            if (joint.rotation != joint.GetChild(0).rotation)
+            {
+                joint.GetChild(0).localRotation.ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
```
mattdeitke, line 522: Just to be safe, can we add named arguments here, as in:

```c#
joint.GetChild(0).localRotation.ToAngleAxis(angle: out angleRot, axis: out vectorRot);
```
(note: angle and axis come from https://docs.unity3d.com/ScriptReference/Quaternion.ToAngleAxis.html)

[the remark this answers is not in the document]
elimvb: Easy

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
elimvb: Easy

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
elimvb: Sure. I extended that to all uses of ToAngleAxis, for consistency.

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -517,17 +517,35 @@ public void DropObject()
...
+                jointMeta.localRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            }
 
+            //edge case for where angler and parent rotations are aligned, which Quaternions have trouble resolving, so we hard-code it here
+            else
+            {
+                jointMeta.localRotation = new Vector4(1, 0, 0, 0);
```
AlvaroHG, line 529: I would change the Vector to Quaternion it is more clear, also here I am confused if you want the identity quaternion or one with the `x` component to 1? could you add comment on why this value

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -517,17 +517,35 @@ public void DropObject()
...
+            if (FirstJoint.rotation != joint.GetChild(0).rotation)
+            {
+                Quaternion.Euler(FirstJoint.InverseTransformDirection(joint.GetChild(0).eulerAngles)).ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
+                jointMeta.rootRelativeRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            }
+
+            //edge case for when angler and root rotations are aligned, which Quaternions have trouble resolving, so we hard-code it here
+            else
+            {
+                jointMeta.rootRelativeRotation = new Vector4(1, 0, 0, 0);
```
AlvaroHG, line 547: Same comment as above

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -517,17 +517,35 @@ public void DropObject()
             jointMeta.rootRelativePosition = FirstJoint.InverseTransformPoint(joint.position);
 
             //local rotation currently relative to immediate parent joint
-            joint.GetChild(0).localRotation.ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
-            jointMeta.localRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            if (joint.rotation != joint.GetChild(0).rotation)
+            {
+                joint.GetChild(0).localRotation.ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
+                jointMeta.localRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
```
AlvaroHG, line 523: If it is a Quaternion assignment use Quaternion instead of Vector4

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -517,17 +517,35 @@ public void DropObject()
...
             jointMeta.rotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
 
             //rotation relative to root joint/agent
             //root forward and agent forward are always the same
-            Quaternion.Euler(FirstJoint.InverseTransformDirection(joint.GetChild(0).eulerAngles)).ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
-            jointMeta.rootRelativeRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            if (FirstJoint.rotation != joint.GetChild(0).rotation)
+            {
+                Quaternion.Euler(FirstJoint.InverseTransformDirection(joint.GetChild(0).eulerAngles)).ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
+                jointMeta.rootRelativeRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
```
AlvaroHG, line 541: Same comment as above

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
elimvb: It's not, it's the axis of rotation, followed by the angle (magnitude).

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
elimvb: I used Vector4 because it's not a Quaternion, but the axis of rotation, followed by the angle. That's the format for these orientation outputs. You are correct about x being 1, when technically z should be 1 instead, since it is the forward vector in Unity.

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
elimvb: I used Vector4 because it's not a Quaternion, but the axis of rotation, followed by the angle. That's the format for these orientation metadata outputs. You are correct about x being 1, when technically z should be 1 instead, since it is the forward vector in Unity.

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
elimvb: I used Vector4 because it's not a Quaternion, but the axis of rotation, followed by the angle. That's the format for these orientation metadata outputs. You are correct about x being 1, when technically z should be 1 instead, since it is the forward vector in Unity.

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
AlvaroHG: localRotation is a Quaternion, if you want an axis of rotation you need to do `Quaternion.AngleAxis(0, new Vector3(1, 0, 0))` I would suggest doing that because what you are doing is setting the raw quaternion values which is represented by 4 real numbers but do not directly mean an Angle Axis rotation, `w` is the real part of the complex number (complex number extension) that a quaternion is and it being zero means a half circle rotation around some axis where `theta` is the angle around that axis and `w = cos(theta / 2)` so `theta` would be `+-π` to satisfy `w = 0` meaning a half circle rotation. localRotation . For `x = 1` in this case it does mean around a unitary vector along x because `sin(theta/2) = 1` with the values noted above, but the general formula is `localRotation.x = axis.x * sin(theta/2)` so it is unintuitive to set the values of the quaternion directly and I am not sure what you're trying to do here

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -517,17 +517,35 @@ public void DropObject()
             jointMeta.rootRelativePosition = FirstJoint.InverseTransformPoint(joint.position);
 
             //local rotation currently relative to immediate parent joint
-            joint.GetChild(0).localRotation.ToAngleAxis(out angleRot, out vectorRot);//getchild to grab the angler since that is what actually changes the geometry angle
-            jointMeta.localRotation = new Vector4(vectorRot.x, vectorRot.y, vectorRot.z, angleRot);
+            if (joint.rotation != joint.GetChild(0).rotation)
```
AlvaroHG, line 520: This is not a good way to compare floats, I don't think it would fix the issue, do an epsilon comparison on the components, or a squared vector magnitude comparison

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
AlvaroHG: it is a Quaternion and will be treated as such https://docs.unity3d.com/ScriptReference/Transform-localRotation.html

elimvb: "jointMeta.localRotation" is not the same as "<transform>.localRotation". "localRotation" is simply the name we use for the custom-class "JointMetadata()" for storing the local rotation, in angle-axis notation. I'll admit the naming convention is misleading.

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
elimvb: The misunderstanding here is in the naming conventions. The name "jointMeta.localRotation" and the transform-property "localRotation" are totally different, albeit with very similar names.

jointMeta.localRotation - A Vector4 where the first three values denote the axis of rotation, and the fourth denotes the number of degrees around that axis. This Quaternion-to-AngleAxis conversion is already done by the line of code above jointMeta.localRotaiton, although I use ToAngleAxis, since AngleAxis has been deprecated.

(transform).localRotation. - A Quaternion denoting a transform's local orientation. Again, this is converted to angle-axis notation before it ever reaches "jointMeta.localRotation".

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
elimvb: What's inherently wrong about comparing two Quaternions against each other?

mattdeitke: Just to add context: these are quaternions (not floats), and the Quaternion class overrides overrides the `==` comparison such that floating point issues are handled automatically:
https://docs.unity3d.com/540/Documentation/ScriptReference/Quaternion-operator_eq.html

> This function tests whether dot product of two quaternions is **close** to 1.0.

So, this should be fine.

Issue #586, opened 2021-01-28
Extend `NumpyAwareEncoder` to understand numpy arrays


bzinberg:
In our use case of AI2-THOR, the data payloads sometimes have array-valued fields.  Because of cross-language interop, in our codebase these are always passed around as Numpy arrays.  Currently, JSON encoding fails for these payloads, including with `NumpyAwareEncoder`.  It would be great to extend `NumpyAwareEncoder` it to understand arrays, in addition to the numpy scalars it already supports.

PR #587, opened 2021-01-28
Extend `NumpyAwareEncoder` to understand arrays

bzinberg:
Fixes https://github.com/allenai/ai2thor/issues/586, and adds a regression test that fails before the change and passes after the change.

### DIFF
--- a/ai2thor/server.py
+++ b/ai2thor/server.py
@@ -17,6 +17,8 @@ import sys
 class NumpyAwareEncoder(json.JSONEncoder):
 
     def default(self, obj):
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
         if isinstance(obj, np.generic):
             return np.asscalar(obj)
         return super(NumpyAwareEncoder, self).default(obj)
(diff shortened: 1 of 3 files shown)


ekolve:
Thanks for the PR and tests.  Are the numpy arrays in your payloads arguments to AI2Thor actions? Or do they exist there as the output from some other process/function?

bzinberg:
Thanks for the quick response @ekolve!

@ThomasSchellenbergNextCentury, @deanwetherby, do you know off hand whether the array-valued data in our use case is an argument to an AI2Thor action or something else?

ThomasSchellenbergNextCentury:
Thanks @bzinberg 

@ekolve We have forked your repo and have heavily modified it for use as the evaluation environment in the DARPA Machine Common Sense program. Rather than having a limited number of static, handmade scenes, we create scenes dynamically by passing a custom JSON property into Unity through the Initialize action that describes the layout of the room and the objects which should appear in that room. Sometimes this JSON data can contain numpy arrays.

PR #593, opened 2021-02-03
Bug fixes to and enhancements to ProcessControlCommand inside of AgentManager

mattdeitke:
Reuses the equivalent `ProcessControlCommand` from `BaseFPSAgentController`, but parameterizes the `target` object reference. 

Using the same idea, we can now call `actionFinished` and throw exceptions (#578) from `AddThirdPartyCamera` and `UpdateThirdPartyCamera` instead of the whole `actionError` and `actionSuccess` shenanigans that has been removed.

This also then fixes (and tests) many hanging bugs, such as:
```python
controller.step('AddThirdPartyCamera', ..., orthographicSize=dict(x=5, y=5))
```
where we expect `orthographicSize` to be a float.

Finally, we provide a better error message in the case that the types mismatch than what is currently provided by Json.Net.

### DIFF
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -220,6 +220,24 @@ def test_add_third_party_camera(controller):
     assert_near(camera[ThirdPartyCameraMetadata.rotation], expectedRotation, 'initial rotation should have been set')
     assert camera[ThirdPartyCameraMetadata.fieldOfView] == expectedFieldOfView, 'initial fieldOfView should have been set'
 
+    # expects position to be a Vector3, should fail!
+    event = controller.step(
+        action="AddThirdPartyCamera",
+        position=5,
+        rotation=dict(x=0, y=0, z=0)
+    )
+    assert not event.metadata['lastActionSuccess'], 'position should not allow float input!'
+
+    # orthographicSize expects float, not Vector3!
+    event = controller.step(
+        action="AddThirdPartyCamera",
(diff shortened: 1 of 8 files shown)


ai2thor/tests/test_unity.py
```diff
@@ -414,7 +432,6 @@ def test_action_dispatch_find_conflicts_stochastic(controller):
     known_conflicts = {
         'GetComponent': ['type'],
         'StopCoroutine': ['routine'],
-        'ProcessControlCommand': ['controlCommand'],
```
mattdeitke, line 417: This is the one thing I'd like to specifically call out. I don't know why this was here before, since AgentManager's `ProcessControlCommand` never had a parameter named `controlCommand` (only `msg`: https://github.com/allenai/ai2thor/blob/master/unity/Assets/Scripts/AgentManager.cs#L1055).

With it, the tests fail, and removing it passes the tests.

ekolve: This is where the conflict would have arisen from:
https://github.com/allenai/ai2thor/blob/master/unity/Assets/Scripts/BaseFPSAgentController.cs#L1453

If the test isn't finding that then something is wrong with the PR.

ekolve: i just scanned through the changes - it looks like controlCommand was changed to serverAction in your PR, so it might be fine to remove.

ekolve (approved):

PR #596, opened 2021-02-04
Updates to allow for computing the semantic hull of objects on the 2d plane.

Lucaweihs:
This PR looks large but it's mostly composed of importing the MIConvexHull library into THOR (which allows for computing convex hulls quickly).


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+        public void CameraRayCast(float x, float y) {            
+            Ray ray = m_Camera.ViewportPointToRay(new Vector3(x, y, 0.0f));
+            RaycastHit hit;
+            Physics.Raycast(
+                ray,
+                out hit,
+                Mathf.Infinity,
+                1 << 0 | 1 << 8 | 1 << 10 | 1 << 11,
+                QueryTriggerInteraction.Ignore
+            );
```
mattdeitke, line 2663: Swapped with named args:
```c#
Physics.Raycast(
     origin: ray,
     direction: out hit,
     maxDistance: Mathf.Infinity,
    layerMask: 1 << 0 | 1 << 8 | 1 << 10 | 1 << 11,
    queryTriggerInteraction: QueryTriggerInteraction.Ignore
);
```

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+            actionFinishedEmit(
+                true, 
+                m_Camera.ScreenToWorldPoint(
+                    new Vector3(col, m_Camera.pixelHeight - row, m_Camera.nearClipPlane)
+                )
+            );
+        }
+
+        public void CameraRayCast(float x, float y) {            
+            Ray ray = m_Camera.ViewportPointToRay(new Vector3(x, y, 0.0f));
```
mattdeitke, line 2655: are you sure `y` shouldn't be `1-y`, like it is in `getTargetObject`, which ensures that `y=0` is at the top and `y=1` is at the bottom?:
https://github.com/allenai/ai2thor/blob/master/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs#L4637

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+            );
+            #if UNITY_EDITOR
+            Debug.DrawLine(
+                hit.point,
+                hit.point + new Vector3(0f, 1f, 0f),
+                Color.red, 
+                100.0f
+            );
+            #endif
+            actionFinishedEmit(true, hit.point);
```
mattdeitke, line 2672: What is the use case from Python for determining the `(x, y, z)` coordinate that is hit from `(x, y)` on the last image frame? Maybe determine distance to the point that is in view(?)

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+                        new Vector3((float) p1.X, maxY, (float) p1.Y),
+                        Color.red, 100.0f
+                    );
+                }
+                #endif
+            }
+            actionFinishedEmit(true, objectIdToConvexHull);
+        }
+        
+        public void GetPixelToWorldPoint(float row, float col, float depth) {
```
mattdeitke, line 2645: Two things here: `depth` is being unused and it's not entirely clear to me what the difference between `GetPixelToWorldPoint` and `CameraRayCast`. Both appear to be returning the world coordinate corresponding to the `(x, y)` point from the last image frame.

Is the intended difference that `GetPixelToWorldPoint` looks at every layer and `CameraRayCast` only looking at layers based on the layer mask `1 << 0 | 1 << 8 | 1 << 10 | 1 << 11`?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+                float maxY = -float.PositiveInfinity;
+
+                foreach (var meshFilter in sop.GetComponentsInChildren<MeshFilter>()) {
+                    foreach (var localVertex in meshFilter.mesh.vertices) {
+                        var globalVertex = meshFilter.transform.TransformPoint(localVertex);
+                        vertices.Add(new MIConvexHull.DefaultVertex2D(globalVertex.x, globalVertex.z));
+                        maxY = Math.Max(maxY, globalVertex.y);
+                    }
+                }
+                var miconvexHull = MIConvexHull.ConvexHull.Create2D(vertices, 1e-10);
```
mattdeitke, line 2625:
```c#
var miconvexHull = MIConvexHull.ConvexHull.Create2D(vertices, 1e-10);
```
becomes something like
```c#
ConvexHullCreationResult miconvexHull = MIConvexHull.ConvexHull.Create2D(
    data: vertices,
    tolerance: 1e-10
);
```
(perhaps it's `ConvexHullCreationResult<TVertex>` instead of `ConvexHullCreationResult`:
https://github.com/DesignEngrLab/MIConvexHull/blob/241b1179e63d12e80ea689436bc37045553a53eb/MIConvexHull/ConvexHull.cs#L98)

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+            if (objectTypes != null) {
+                allowedObjectTypesSet = new HashSet<string>(objectTypes);
+            }
+
+            var objectIdToConvexHull = new Dictionary<string, List<(float x, float y)>>();
+            foreach (var sop in GameObject.FindObjectsOfType<SimObjPhysics>()) {
+                if (
+                    allowedObjectTypesSet != null 
+                    && !allowedObjectTypesSet.Contains(sop.Type.ToString())
+                ) {
```
mattdeitke, line 2612: Debatably would make this:
```c#
if (allowedObjectTypesSet != null &&
    !allowedObjectTypesSet.Contains(sop.Type.ToString())
) {
```
The C# autoformatters will ignore both cases.

mattdeitke (changes requested):
Pretty neat!

I've added a few comments to `Get2DSemanticHulls`. I think a 3d version may also seem useful, but I suppose that can be implemented pretty easily later using MIConvexHull.

I've also added a few questions about the usage and implementations of `GetPixelToWorldPoint` and `CameraRayCast`.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
             actionFinished(true);
         }
 
+        public void Get2DSemanticHulls(List<string> objectTypes = null) {
```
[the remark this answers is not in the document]
Lucaweihs: 1. `objectTypes` actually isn't required, if it is `null` then it is assumed that all object types should be used. That said, I've changed the api to
```python
        public void Get2DSemanticHulls(
            List<string> objectIds = null,
            List<string> objectTypes = null
        ) {
```
so that a user can specify ids as well.

2. The main reason to have `objectTypes` is that this function is targeting semantic information which is inherently object-type-centric. It's "easy" to handle this all on the python side but is also a bit annoying in practice since you have to do it every time you call the function (or at least I would).

3. This overload is now in.

I've also added a comment:
```python
        /*
        For each objectId, create a convex hull of the object from a top-down view.
        The convex hull will be represented as a list of (x, z) world coordinates
        such that the boundary formed by these coordinates forms the convex hull
        of these points (smallest convex region enclosing the object's points).

        If the objectIds (or objectTypes) parameter is non-null, then only objects with
        those ids (or types) will be returned.
        
        ONLY ONE OF objectIds OR objectTypes IS ALLOWED TO BE NON-NULL.

        Returns a dictionary mapping object ids to a list of (x,z) coordinates corresponding
        to the convex hull of the corresponding object.
        */
```

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+                for (int i = 0; i < pointsOnHullArray.Length; i++) {
+                    var p0 = pointsOnHullArray[i];
+                    var p1 = pointsOnHullArray[(i + 1) % pointsOnHullArray.Length];
+                    Debug.DrawLine(
+                        new Vector3((float) p0.X, maxY, (float) p0.Y),
+                        new Vector3((float) p1.X, maxY, (float) p1.Y),
+                        Color.red, 100.0f
+                    );
+                }
+                #endif
```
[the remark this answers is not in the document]
Lucaweihs: I've updated to explicit types expect in one place where it made the line outrageously long.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+                ) {
+                    continue;
+                }
+                var vertices = new List<MIConvexHull.DefaultVertex2D>();
+                float maxY = -float.PositiveInfinity;
+
+                foreach (var meshFilter in sop.GetComponentsInChildren<MeshFilter>()) {
+                    foreach (var localVertex in meshFilter.mesh.vertices) {
+                        var globalVertex = meshFilter.transform.TransformPoint(localVertex);
+                        vertices.Add(new MIConvexHull.DefaultVertex2D(globalVertex.x, globalVertex.z));
```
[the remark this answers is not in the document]

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+                Debug.Log(sop.ObjectID);
+                var pointsOnHullArray = miconvexHull.Result.ToArray();
+                for (int i = 0; i < pointsOnHullArray.Length; i++) {
+                    var p0 = pointsOnHullArray[i];
+                    var p1 = pointsOnHullArray[(i + 1) % pointsOnHullArray.Length];
+                    Debug.DrawLine(
+                        new Vector3((float) p0.X, maxY, (float) p0.Y),
+                        new Vector3((float) p1.X, maxY, (float) p1.Y),
+                        Color.red, 100.0f
+                    );
```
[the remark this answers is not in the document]

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,6 +2598,79 @@ bool includeInvisible
...
+                QueryTriggerInteraction.Ignore
+            );
+            #if UNITY_EDITOR
+            Debug.DrawLine(
+                hit.point,
+                hit.point + new Vector3(0f, 1f, 0f),
+                Color.red, 
+                100.0f
+            );
+            #endif
```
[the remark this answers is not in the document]

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: I'm using a different overload but done.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: I've decided to remove this and the `GetPixelToWorldPoint` functions. They're useful for debugging but perhaps for now we don't want them in the public api.

unity/Assets/Scripts/BaseFPSAgentController.cs

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: My style is generally that either everything is on a single line with the parens or the parens are on their own.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Action removed.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Action removed.

ekolve:
Look good, can you add the following tests:

* Get2DSemanticHulls with objectTypes
* Get2DSemanticHulls with objectIds
* Get2DSemanticHulls with objectId

This can all be done in ai2thor/tests/test_unity.py.  I am fine if you want to just take the actionReturn output and inline value or stash it in ai2thor/tests/data for each of the actions and assert that we get the same value.

Lucaweihs:
@ekolve Added the tests.

mattdeitke (approved):

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2593,10 +2594,118 @@ bool includeInvisible
...
+                #if UNITY_EDITOR
+                Debug.Log(sop.ObjectID);
+                #endif
+
+                objectIdToConvexHull[sop.ObjectID] = Get2DSemanticHull(sop.gameObject);
+            }
+            actionFinishedEmit(true, objectIdToConvexHull);
+        }
+
+        public void Get2DSemanticHulls(string objectId) {
```
mattdeitke, line 2698: Wait, real quick: would rename this one to `Get2DSemanticHull` (no plural).

Lucaweihs: I thought about this but I think I'd prefer to have it as it is as this way it's just an overload for a given action rather than having another whole action for this one behavior.

mattdeitke: This is true, for now. But, if we ever create a matching Python API, since Python does not support overloading methods, it is much more of a pain to get `Get2DSemanticHulls` working for `objectId`, `objectIds`, and `objectTypes` as input, rather than having it work for:
* `Get2DSemanticHulls`: `List[str]` (objectIds or objectTypes)
* `Get2DSemanticHulls`: `str`

Lucaweihs: Fair enough, done.

Lucaweihs:
@ekolve Tests now pass (although I had to relax the test conditions because of inconsistencies between the Linux/Mac versions).

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -266,6 +266,9 @@ public void SetAgentRadius(float agentRadius = 2.0f)
         public GameObject FindObjectInVisibleSimObjPhysics(string objectID) {
             GameObject target = null;
 
+            if (VisibleSimObjPhysics is null) {
```
ekolve, line 269: We should just call VisibleSimObjs (i can refactor this when it gets merged to main on the IsInteractableBranch) instead of calling LateUpdate().  LateUpdate is assumed to be called only once per frame by the Unity Lifecycle and if for some reason VisibleSimObjs were remove from LateUpdate, this would not work.  As well, VisibleSimObjs is only populate for the Editor and WEBGL.

ekolve (changes requested):

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
Lucaweihs: Fixed, I ended up refactoring this function a bit (and as you note, I think this function can be completely replaced once your IsInteractableBranch fixes are in).

PR #597, opened 2021-02-05
better mapView placement in iTHOR

mattdeitke:
Many scenes with ToggleMapView in iTHOR have pretty bad boundaries:
image

This now just uses the ceiling to set boundaries, if there is a ceiling structural object in the scene (i.e., iTHOR scenes). Otherwise, it sticks with the old scene boundaries (i.e., with RoboTHOR scenes).

### DIFF
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -281,7 +281,7 @@ public class AgentManager : MonoBehaviour
 	}
 
 	public void ResetSceneBounds() {
-		// Recordining initially disabled renderers and scene bounds
+		// Recording initially disabled renderers and scene bounds
 		sceneBounds = new Bounds(
 			new Vector3(float.PositiveInfinity, float.PositiveInfinity, float.PositiveInfinity),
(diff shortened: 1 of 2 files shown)


Lucaweihs:
I suppose I'm ok with doing this although it might break backwards compatability for people who assume these map views have extents equal to the scene boundaries. Is it not possible to just fix the scene bounds so that they don't result in such broken views? I imagine there must be some geometry in these scenes that's causing the problem.

mattdeitke:
Perhaps it's possible to go through each scene and check why `sceneBounds` are off, which happens in probably 50+ scenes.

But I'm not convinced that it's worth the cost right now. Moreover, since `ToggleMapView` isn't documented, we provide no real guarantees with respect to backwards compatibility. Here, `ToggleMapView`s camera position is also accessible from the camera bounds.

This is kinda just a nice hack to get visualizations looking more consistent and centered pretty much instantly with little to no changes other.

Lucaweihs (approved):
I'm fine with these changes, we should fix the scene bounds at some point though :)

PR #601, opened 2021-02-09
Better Segmentation Names

mattdeitke:
Summary:
* `renderClassImage` becomes `renderSemanticSegmentation`
* `renderObjectImage` becomes `renderInstanceSegmentation`
* `event.class_segmentation_frame` becomes `event.semantic_segmentation_frame`

Everything is fully backwards compatible, and a few tests have been added.

### On changing `class` to `segmentation`.

The word `class` is not used anywhere else, segmentation does not appear in the render part, and the phrase class segmentation is not necessarily common (Google suggests that it's referring to economic class). What the frame does is render segmentation by `objectType`, not object instance.

### On changing `object` to `instance`.

We stick with `event.instance_segmentation_frame` for naming consistency with initialization and upon accessing. It was too hard to remember which one used the name `object` and which one used the name `instance`. Moreover, the phrase instance segmentation is more commonly used across all of computer vision. This then renders segmentation by object instance.


ai2thor/controller.py
```diff
@@ -671,24 +737,44 @@ def step(self, action=None, **action_args):
...
+        # support for deprecated parameter names (old: new)
+        changed_parameter_names = {
+            "renderClassImage": "renderSemanticSegmentation",
+            "renderObjectImage": "renderInstanceSegmentation",
+        }
+        for old, new in changed_parameter_names.items():
+            if old in action:
+                warnings.warn(old + " has been renamed to " + new)
+                action[new] = action[old]
+                del action[old]
```
mattdeitke, line 761: This is about the only thing that changes in the file. Basically just remaps the old parameter names to the new ones and provides a warning that they've been renamed.

ai2thor/controller.py
```diff
@@ -671,24 +737,44 @@ def step(self, action=None, **action_args):
...
-        rotation = action.get('rotation')
+        rotation = action.get("rotation")
         if rotation is not None and type(rotation) != dict:
-            action['rotation'] = {}
-            action['rotation']['y'] = rotation
+            action["rotation"] = {}
+            action["rotation"]["y"] = rotation
+
+        # support for deprecated parameter names (old: new)
+        changed_parameter_names = {
```
ekolve, line 753: Can you change the structure of this to the following:

changed_parameter_names[<ACTION_NAME>][<OLD_KEY>] = <NEW_KEY>

presently, we have to iterate through this list on each and every action and in theory someone could add a new mapping in there for an action and it would inadvertently remap the param name for a different action.

ai2thor/controller.py
```diff
@@ -671,24 +737,44 @@ def step(self, action=None, **action_args):
...
+            "renderClassImage": "renderSemanticSegmentation",
+            "renderObjectImage": "renderInstanceSegmentation",
+        }
+        for old, new in changed_parameter_names.items():
```
ekolve, line 757: this would then become:

```
for old, new in change_parameter_names.get(action['action'], {}).items():
```
or slightly more verbose but faster:
```
if action['action'] in changed_parameter_names:
    for old, new in change_parameter_names[action['action']].items():

```

ai2thor/controller.py
```diff
@@ -671,24 +737,44 @@ def step(self, action=None, **action_args):
...
+            if old in action:
+                warnings.warn(old + " has been renamed to " + new)
```
ekolve, line 759: include the action name in the warning

ekolve (changes requested):

ekolve (approved):

ekolve:
lgtm

PR #603, opened 2021-02-10
Resolving divide by zero in compute_single_spl when path distance is 0

apoorvkh:
Setting path length ratio to 1.0 in `compute_single_spl` when path distance is 0 (causing `ZeroDivisionError`).


Lucaweihs (approved):
LGTM.

PR #604, opened 2021-02-10
Making the "Move in direction" action part of the public API.

Lucaweihs:
To the best of my understanding we currently we have no way (in the public API) to move the agent in an arbitrary given direction in the global coordinate frame. I would like to add this as I have use for it.


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -782,6 +782,10 @@ public IEnumerator checkInitializeAgentLocationAction()
             }
         }
 
+        public void MoveInDirection(Vector3 direction) {
```
mattdeitke, line 785: Perhaps this is a duplicate of MoveRelative?
https://github.com/allenai/ai2thor/blob/master/unity/Assets/Scripts/BaseFPSAgentController.cs#L1713

Moreover, allowing one to specify a non-0 `y` in `direction kinda scares me. I'm not actually sure what it would do. For better or worse, all actions in the BaseFPSAgentController assume that the agent can only move along the XZ axes.

mattdeitke (changes requested):

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Move relative is in the agent's coordinate frame, whereas move in direction is using the global coordinate frame. Happy to refactor to only allow x/z movement.

mattdeitke: Oh, I see. Seems useful then.

Yeah, I think restricting to x/z would work best.

mattdeitke: The other note: I would probably rename this to something like `MoveGlobal` to make the difference between `MoveRelative` clear.

Alternatively, since `MoveRelative` is undocumented (and thus easy to deprecate), it could just be worth adding a `coordinateSpace: string` parameter to `MoveInDirection()`, which allows for either specifying `relative` or `world`.

Lucaweihs: 👍 I have changed it `MoveGlobal` (note however that the `MoveRelative` command is a bit weird because it uses the usual video game idea of movement with very few checks as opposed to what we do). I also updated the DebugInputField.cs script which might be worth looking at.

PR #605, opened 2021-02-10
agentMode="bot" renamed to agentMode="locobot", enhance default scene, provide error warnings

mattdeitke:
#### Renames `agentMode="bot"` to `agentMode="locobot"`.

Using `agentMode="bot"` is still backwards compatible, but now deprecated. This makes it more intuitive why `agentMode="bot"` cannot move left/right, has stochastic movements, and cannot interact with objects using "magic physics", since it's modeled specifically to match the real life LoCoBot. A test has also been added for backwards compatibility.

#### Add RoboTHOR scenes to the hidden `controller.scene_names()` method, along with the ability to filter scenes by type.


ai2thor/controller.py
```diff
@@ -481,19 +482,44 @@ def __init__(
...
+                        scene = random.choice(list(robothor_scenes_in_build))
+                    else:
+                        ithor_scenes = set(self.ithor_scenes())
+                        ithor_scenes_in_build = ithor_scenes.intersection(scenes_in_build)
+                        if ithor_scenes_in_build:
+                            # Prioritize iTHOR because that's what the default agent best uses.
+                            scene = random.choice(list(ithor_scenes_in_build))
+                        else:
+                            # perhaps only using RoboTHOR or using only custom scenes
+                            scene = random.choice(list(scenes_in_build))
```
Lucaweihs, line 514: I would argue strongly for this logic to change. It should be something like:
1. If using "locobot" (or "bot"!) then use the first element of `list(sorted(robothor_scenes_in_build))`.
2. Otherwise use the first element of `list(sorted(ithor_scenes_in_build))`.
 
There is no reason to introduce any randomness or favor scene 28 (especially as 28 is often used for testing!).

ai2thor/controller.py
```diff
@@ -481,19 +482,44 @@ def __init__(
...
-                    "Initialize action failure: {}".format(
-                        event.metadata["errorMessage"]
-                    )
-                )
+
+            # older builds don't send actionReturn on Initialize
+            init_return = event.metadata['actionReturn']
+            if init_return:
+                self.server.set_init_params(init_return)
+                logging.info("Initialize return: {}".format(init_return))
```
Lucaweihs, line 522: I don't quite understand why the other logic was deleted. It seems like checking whether or not the reset succeeded (and otherwise throwing an error) seems important.

ai2thor/controller.py
```diff
@@ -532,6 +558,7 @@ def __exit__(self, *args):
         self.stop()
 
     @property
+    @lru_cache()
```
Lucaweihs, line 561: Caching is already happening in the `scenes_in_build` function through the use of the private `_scenes_in_build` attribute so I don't see the benefit of adding a lru cache?

ai2thor/controller.py
```diff
@@ -545,8 +572,8 @@ def scenes_in_build(self):
 
         return self._scenes_in_build
 
-    def reset(self, scene="FloorPlan_Train1_1", **init_params):
-        if re.match(r"^FloorPlan[0-9]+$", scene):
+    def reset(self, scene, **init_params):
```
Lucaweihs, line 575: Seems like forcing people to specify the scene here will cause backwards compatibility issues. Can we change this to `scene: str = "default"` or something and grab the first scene in the build as discussed above?

Lucaweihs (commented):
Good ideas! Some comments about the default logic but otherwise LGTM.

ai2thor/controller.py
mattdeitke: The deleted logic was moved inside of `reset`. For reference `reset` now calls Unity's `Initialize`.

There should basically never be a reason that `Initialize` fails, and if it does, something has gone badly wrong.

ai2thor/controller.py
mattdeitke: I think using `controller._scenes_in_build` is quite dangerous. For one, it is initialized to `None` and so if you call `self._scenes_in_build`, it's possible that the value is none.

Of course, there are workarounds, such as checking `if not self._scenes_in_build: <generate scenes_in_build>` each time one may ever need to call `controller.scenes_in_build`, but this is an uglier workarounds than just using `controller.scenes_in_build`, which will always produce the correct output (of the scenes in the build). We can then cache this value because it will not change.

mattdeitke: Moreover, `self.scenes_in_build` has been used all over the place, and to the best of my knowledge, `self._scenes_in_build` is basically used nowhere.

mattdeitke: Actually, nevermind. I see what you're saying here.

Lucaweihs: I think I might not have been clear. In the current `scenes_in_build` function, the first thing it does is check if `self._scenes_in_build is None` and, if it's not, returns `self._scenes_in_build`. Otherwise it computes a value for `self._scenes_in_build` and then returns it. I.e. this function _is already_ doing caching. The `@lru_cache()` call is thus redundant. If you want to remove `self._scenes_in_build` and just use `@lru_cache()` that would also be fine but, in that case, what you really want is `@lazy_property` (see how this is implemented in pytorch) rather than an `lru_cache` as a lru cache has to do a dictionary lookup.

ai2thor/controller.py
mattdeitke: Fixed.

Just a small typo, but since `ithor_scenes_in_build` and `robothor_scenes_in_build` are sets, they had to be convert to a list first before sorting, so `list(sorted(ithor_scenes_in_build))` is `sorted(list(ithor_scenes_in_build))`.

mattdeitke: I originally went with random scenes to show off the diversity, but fair point on that being unexpected.

ai2thor/controller.py
mattdeitke: Added the initial fix by just removing lru_cache for now. Potentially can slightly optimize it in the future, but probably not worth doing at the moment.

ai2thor/controller.py
mattdeitke: Similar to all other `init_params` that work by reusing the latest update if they have not been changed, I fixed this by setting `scene=None` by default, where if `scene=None`, reuse the last used scene.

Lucaweihs (approved):
LGTM.

ai2thor/controller.py
```diff
@@ -569,6 +596,7 @@ def key_sort_func(scene_name):
             )
 
         self.server.send(dict(action="Reset", sceneName=scene, sequenceId=0))
+        self.last_scene = scene
```
ekolve, line 599: what is the purpose of last_scene?

ai2thor/controller.py
```diff
@@ -588,12 +616,29 @@ def key_sort_func(scene_name):
...
+        ):
+            warnings.warn("You are using a RoboTHOR scene without using the standard LoCoBot.\n" +
+             "Did you mean to mean to set agentMode='locobot' upon initialization or within controller.reset(...)?")
+
         self.last_event = self.step(
             action="Initialize", **self.initialization_parameters
         )
 
-        self.scene = scene
+        if not self.last_event.metadata['lastActionSuccess']:
```
ekolve, line 636: you can pass raise_for_failure=True to step('Initialize') and skip the lastActionSuccess check.

ekolve (approved):

ekolve:
i had one question regarding self.last_scene, but otherwise looks fine.

ai2thor/controller.py
mattdeitke: If a scene is not passed into reset, then `self.last_scene` is used. This just makes the defaults consistent with `init_params` (see: https://github.com/allenai/ai2thor/pull/605#discussion_r574868354).

mattdeitke: Oh, wait. I suppose this is a duplicate of `self.scene`, which does the same thing.

mattdeitke: Fixed in https://github.com/allenai/ai2thor/pull/605/commits/764cd71b4e2e3540049d19806d50b5d087654737

ai2thor/controller.py
mattdeitke: Fixed in https://github.com/allenai/ai2thor/pull/605/commits/764cd71b4e2e3540049d19806d50b5d087654737

PR #609, opened 2021-02-11
Remove deprecated initialize_func

mattdeitke:
`initialize_func` does not make sense to use. Consider that it is both slower and requires utilizing and remembering deprecated functionality:
```python
controller = Controller()

def initialize_func():
    # Calling this manually is deprecated and twice as slow as calling reset.
    # It also requires remembering which Controller initialization parameters are actually meant to
    # be passed into Unity
    controller.step("Initialize", gridSize=0.35)

get_episodes_with_shortest_paths(
    controller=controller,
    episodes=...,
    initialize_func=initialize_func
)
```
Instead, since `reset` stores the parameters that were set upon initialization and passes them into Unity after each step, this whole thing is equivalent to the more natural:
```python
controller = Controller(gridSize=0.35)

get_episodes_with_shortest_paths(
    controller=controller,
    episodes=...
)
```

It doesn't look like any open source projects use this with `initialize_func` (or in general): https://github.com/search?q=get_episodes_with_shortest_paths


ekolve (approved):

ekolve:
lg

PR #610, opened 2021-02-11
Fixing TeleportFull bugs.

Lucaweihs:
`TeleportFull` was doing weird things resulting in an agent possibly not having the precise (x,z) coordinates one would expect (often off by relatively large margins). This update fixes these bugs and makes the error checking more precise.


winthos:
Could you get the last commit from my original TeleportFix branch? It cleaned up the code and removed some print statements from the debug commit that is already here

ai2thor/controller.py
```diff
@@ -769,16 +769,20 @@ def step(self, action=None, **action_args):
...
+            ]:
+                raise ValueError(self.last_event.metadata["errorMessage"])
+            elif raise_for_failure:
+                raise RuntimeError(
+                    self.last_event.metadata.get("errorMessage", f"{action} failed.")
+                )
 
-        if raise_for_failure:
-            assert self.last_event.metadata["lastActionSuccess"]
+        assert (not raise_for_failure) or self.last_event.metadata["lastActionSuccess"]
```
ekolve, line 785: We should be able to remove this since the change you added will throw the exception if necessary.

ekolve (approved):

ekolve:
looks good, merge on green

Lucaweihs:
@ekolve Just to confirm: this will cause some backwards compatibility issues (i.e. people who were using the `TeleportFull` command before will, almost certainly, be missing some inputs). In particular, for instance in my code I realize I almost never specified the `standing` parameter and this was just implicitly always taken to be `true`, you're ok with this correct?

ekolve:
@Lucaweihs If you are happy with the new restriction, I'm fine with it.  If it does become an issue for users we can make the parameters nullable and default them to the agent's current value for that param.  It does make sense to me that TeleportFull would require all parameters to put the agent into a new pose and also by requiring everything you ensure that the agent will always end up in exactly the same state.

mattdeitke:
I still don't know if I like this because it will inevitably break as we introduce new degrees of freedom to an agent.

I'm still siding with the following:
* Let's require `x`, `y`, `z` to Teleport. Add `standing`, `rotation`, and `horizon` to Teleport as optional parameters.
* Deprecate TeleportFull, since it does the same thing.

mattdeitke:
I suppose this might be alievated with https://github.com/allenai/ai2thor/pull/568#discussion_r558606511, if you use, say, GetInteractablePoses in tandom with TeleportFull.

I would then just make sure that TeleportFull works with GetInteractablePoses for each agent using the ** syntax.

We would also have to say something like `TeleportFull` is extremely strict. If we add new degrees of freedom to (say `GetIneractablePoses`), your old code may break. Use Teleport in this case.

Lucaweihs:
@mattdeitke I like your second proposal, let's have a strict `TeleportFull` and a permissive `Teleport`.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1823,76 +1849,126 @@ protected bool checkForUpDownAngleLimit(string direction, float degrees)
         }
 
         //teleport full, base version does not consider being able to hold objects
-        public virtual void TeleportFull(ServerAction action) {
-            targetTeleport = new Vector3(action.x, action.y, action.z);
+        public virtual void TeleportFull(
+            float x,
+            float y,
+            float z,
+            Vector3 rotation,
```
mattdeitke, line 1856: One other request, both for Teleport's new `rotation` and TeleportFull's existing rotation. Can we overload this from the C# side such that rotation can also take in a float on the BaseFPS and PhysicsFPS?

This is already happen in an extremely dangerous fashion from the Python side: https://github.com/allenai/ai2thor/blob/master/ai2thor/controller.py#L840 and it is quite more easier to now have to type
```python
rotation=myActualRotation
```
instead of
```python
rotation=dict(x=0, y=<myActualRotation>, z=0)
```

Lucaweihs: I tried this override and it breaks the action dispatcher unfortunately :(

Lucaweihs: It will try to do things like type casting a dictionary as a float and fail obscurely.

ekolve: Did you try creating a method like this:

```
        public void TeleportFull(
            float x,
            float y,
            float z,
            float rotation) {
   TelportFull(x, y, z, new Vector3(0.0f, rotation, 0.0f));
}
```

That should work - let me know if it does not.

The only issue I have with assuming rotation always equals rotation about the y axis is that it presumes the type of agent and how it can rotate.  A drone for example could rotate in all dimensions.

Lucaweihs: Maybe I misunderstood the error I was getting but when I tried adding a method with arguments:
```C#
public void TeleportFull(
            float x,
            float y,
            float z,
            float rotation,
            float horizon,
            bool standing,
            bool forceAction = false
) { ... }
```
it failed and complained about being unable to type cast rotation appropriately.

ekolve: What were you passing from the Python side to the step() function?

Lucaweihs: @ekolve  Oh, are we ok with removing the line that @mattdeitke pointed out? I'm not sure what the impact of this would be for existing code.

mattdeitke: Preferably don't to remove it until `TeleportFull` works with `rotation: float` for the default and locobot agents. These change in bodily rotations with respect to their yaw. I use it in a bunch of places with teleport.

PR #613, opened 2021-02-14
Teleport bugs

mattdeitke:
The new changes were incompatible with the locobot and drone's ability to teleport, since they do not have the ability to stand.

I've also reproduced the bug in the action dispatcher, where adding in `rotation: float` alongside `rotation: Vector3` tries to get the float to convert to a Vector3. This has not been fixed in this PR. I have simply commented out the new Teleport actions that would have allowed `rotation: float`.

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -837,6 +837,13 @@ class Controller(object):
 
         self.last_action = action
 
+        # dangerously converts rotation(float) to rotation(dict(x=0, y=float, z=0))
+        # this should be removed when ServerActions have been removed from Unity
+        # for all relevant actions.
+        rotation = action.get("rotation")
+        if rotation is not None and not isinstance(rotation, dict):
+            action['rotation'] = dict(y=rotation)
+
         # Support for deprecated parameter names (old: new)
         # Note that these parameters used to be applicable to ANY action.
         changed_parameter_names = {
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -862,10 +862,8 @@ class Controller(object):
         ]:
             raise ValueError(self.last_event.metadata["errorMessage"])
 
-        if raise_for_failure:
-            assert self.last_event.metadata[
-                "lastActionSuccess"
-            ], self.last_event.metadata.get("errorMessage", f"{action} failed")
+        if raise_for_failure and not self._last_event.metadata["lastActionSuccess"]:
+            raise RuntimeError(self.last_event.metadata.get("errorMessage", f"{action} failed"))
 
         return self.last_event
 
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -852,18 +852,18 @@ class Controller(object):
         self.server.send(action)
         self.last_event = self.server.receive()
 
-        if not self.last_event.metadata[
-            "lastActionSuccess"
-        ] and self.last_event.metadata["errorCode"] in [
-            "InvalidAction",
-            "MissingArguments",
-            "AmbiguousAction",
-            "InvalidArgument",
-        ]:
-            raise ValueError(self.last_event.metadata["errorMessage"])
-
-        if raise_for_failure and not self._last_event.metadata["lastActionSuccess"]:
-            raise RuntimeError(self.last_event.metadata.get("errorMessage", f"{action} failed"))
+        if not self.last_event.metadata["lastActionSuccess"]:
+            if self.last_event.metadata["errorCode"] in [
+                "InvalidAction",
+                "MissingArguments",
+                "AmbiguousAction",
+                "InvalidArgument",
+            ]:
+                raise ValueError(self.last_event.metadata["errorMessage"])
+            elif raise_for_failure:
+                raise RuntimeError(
+                    self.last_event.metadata.get("errorMessage", f"{action} failed")
+                )
 
         return self.last_event
 
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -863,7 +863,9 @@ class Controller(object):
             raise ValueError(self.last_event.metadata["errorMessage"])
 
         if raise_for_failure:
-            assert self.last_event.metadata["lastActionSuccess"]
+            assert self.last_event.metadata[
+                "lastActionSuccess"
+            ], self.last_event.metadata.get("errorMessage", f"{action} failed")
 
         return self.last_event
 
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -837,11 +837,6 @@ class Controller(object):
 
         self.last_action = action
 
-        rotation = action.get("rotation")
-        if rotation is not None and type(rotation) != dict:
-            action["rotation"] = {}
-            action["rotation"]["y"] = rotation
-
         # Support for deprecated parameter names (old: new)
         # Note that these parameters used to be applicable to ANY action.
         changed_parameter_names = {
@@ -859,20 +854,16 @@ class Controller(object):
 
         if not self.last_event.metadata[
             "lastActionSuccess"
-        ]:
-            if self.last_event.metadata["errorCode"] in [
+        ] and self.last_event.metadata["errorCode"] in [
             "InvalidAction",
             "MissingArguments",
             "AmbiguousAction",
             "InvalidArgument",
-            ]:
-                raise ValueError(self.last_event.metadata["errorMessage"])
-            elif raise_for_failure:
-                raise RuntimeError(
-                    self.last_event.metadata.get("errorMessage", f"{action} failed.")
-                )
+        ]:
+            raise ValueError(self.last_event.metadata["errorMessage"])
 
-        assert (not raise_for_failure) or self.last_event.metadata["lastActionSuccess"]
+        if raise_for_failure:
+            assert self.last_event.metadata["lastActionSuccess"]
 
         return self.last_event
 
--- a/unity/Assets/Scripts/ActionDispatcher.cs
+++ b/unity/Assets/Scripts/ActionDispatcher.cs
@@ -198,10 +198,11 @@ public static class ActionDispatcher {
                     for (int k = 0; k < minCommon; k++) {
                         if (sourceParams[k].ParameterType != targetParams[k].ParameterType) {
                             signatureMatch = false;
+                            break;
                         }
                     }
 
-                    if (sourceParams.Length != targetParams.Length) {
+                    if (signatureMatch && (sourceParams.Length != targetParams.Length)) {
                         // Some methods, like teleport full have
                         // add both required args and optional args to the end.
                         for (int i = minCommon - 1; i < sourceParams.Length; i++) {
--- a/unity/Assets/Scripts/ActionDispatcher.cs
+++ b/unity/Assets/Scripts/ActionDispatcher.cs
@@ -204,9 +204,6 @@ public static class ActionDispatcher {
                     if (sourceParams.Length != targetParams.Length) {
                         // Some methods, like teleport full have
                         // add both required args and optional args to the end.
-
-                        // TODO: Prioritize the child class one in the case of teleport,
-                        // when it's ambiguous.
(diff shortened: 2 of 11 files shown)


ai2thor/controller.py
```diff
@@ -859,20 +854,16 @@ def step(self, action=None, **action_args):
...
-            elif raise_for_failure:
-                raise RuntimeError(
-                    self.last_event.metadata.get("errorMessage", f"{action} failed.")
-                )
+        ]:
+            raise ValueError(self.last_event.metadata["errorMessage"])
 
-        assert (not raise_for_failure) or self.last_event.metadata["lastActionSuccess"]
+        if raise_for_failure:
+            assert self.last_event.metadata["lastActionSuccess"]
```
Lucaweihs, line 866: Can you add the error message?

```python
assert self.last_event.metadata["lastActionSuccess"], self.last_event.metadata.get("errorMessage", f"{action} failed.")
```

unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -167,75 +167,74 @@ public static class ActionDispatcher {
...
                     bool signatureMatch = true;
                     ParameterInfo[] targetParams = methods[j].GetParameters();
                     int minCommon = Math.Min(sourceParams.Length, targetParams.Length);
-                    for (int k = 0; k < minCommon; k++)
-                    {
-                        if (sourceParams[k].ParameterType != targetParams[k].ParameterType)
-                        {
+                    for (int k = 0; k < minCommon; k++) {
+                        if (sourceParams[k].ParameterType != targetParams[k].ParameterType) {
                             signatureMatch = false;
```
Lucaweihs, line 200: Couldn't there be a `break;` added here?

unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -167,75 +167,74 @@ public static class ActionDispatcher {
...
                     }
 
-                    if (sourceParams.Length > targetParams.Length && !sourceParams[minCommon].HasDefaultValue)
-                    {
-                        signatureMatch = false;
-                    }
-                    else if (targetParams.Length > sourceParams.Length && !targetParams[minCommon].HasDefaultValue)
-                    {
-                        signatureMatch = false;
+                    if (sourceParams.Length != targetParams.Length) {
```
Lucaweihs, line 204: Would `signatureMatch && (sourceParams.Length != targetParams.Length)` be more efficient?

unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -167,75 +167,74 @@ public static class ActionDispatcher {
...
+                        // Some methods, like teleport full have
+                        // add both required args and optional args to the end.
+                        for (int i = minCommon - 1; i < sourceParams.Length; i++) {
```
Lucaweihs, line 207: Should `i` start at `minCommon` or `minCommon-1`? It seems that the old code only checked starting at `minCommon`, was this an error?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1216,30 +1216,32 @@ public class MetadataPatch
...
+public class AgentMetadata {
     public string name;
     public Vector3 position;
     public Vector3 rotation;
     public float cameraHorizon;
-	public bool isStanding;
+
+    // TODO: this should be removed from base.
+    // some agents cannot stand (e.g., drone, locobot)
+	public bool? isStanding = null;
```
Lucaweihs, line 1227: Agreed.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1849,28 +1859,27 @@ public class TypedVariable {
...
+          property.PropertyName == "magnitude"  ||
+          property.PropertyName == "normalized")
+      ) {
          property.ShouldSerialize = instance => { return false; };
          return property;
       } else {
           return base.CreateProperty(member, memberSerialization);
       }
-
+      return base.CreateProperty(member, memberSerialization);
```
Lucaweihs, line 1882: Can this return ever occur? Seems like the above if/else clause has returns in both branches.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -1609,165 +1612,209 @@ private Vector3[] GenerateArcPoints(Vector3 startingPoint, Vector3 origin, float
...
+                oldLocalHandPosition = ItemInHand.transform.localPosition;
+                oldLocalHandRotation = ItemInHand.transform.localRotation;
+            }
 
-                //default high level hand when teleporting
+            try {
+                // default high level hand when teleporting
                 DefaultAgentHand();
+                ToggleArmColliders(arm: Arm, value: forceAction);
+                base.TeleportFull(position: position, rotation: rotation, horizon: horizon, forceAction: forceAction);
```
Lucaweihs, line 1695: Won't this call already potentially set `actionFinished(true)` and thus allow unity to emit the metadata? Perhaps you want to create a `teleportFull` method that can be called programmatically?

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -1609,165 +1612,209 @@ private Vector3[] GenerateArcPoints(Vector3 startingPoint, Vector3 origin, float
...
-                    return;
-                }
+        // has to consider both the arm and standing
+        public void TeleportFull(
+            Vector3 position,
+            Vector3 rotation,
+            float horizon,
+            bool standing,
+            bool forceAction = false
+        ) {
```
Lucaweihs, line 1677: If we're going by the mentality that a call to `TeleportFull` must always specify all agent parameters, then we need to explicitly override all `base.TeleportFull` methods that do not have a `standing` attribute and have them all throw an error.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -6806,7 +6855,7 @@ public bool objectIsCurrentlyVisible(SimObjPhysics sop, float maxDistance)
                 // Consider the case where one does not want to move on a perfect grid, and is currently moving
                 // with an offsetted set of rotations like {10, 100, 190, 280} instead of the default {0, 90, 180, 270}.
                 // This may happen if the agent starts by teleports with the rotation of 10 degrees.
-                int offset = (int) Math.Round(transform.eulerAngles.y % rotateStepDegrees);
+                int offset = (int) Math.Round(transform.localEulerAngles.y % rotateStepDegrees);
```
Lucaweihs, line 6858: Why local euler angles here?

Lucaweihs (commented):
Great fixes, added some minor comments.

unity/Assets/Scripts/ActionDispatcher.cs
mattdeitke: This is a bug fix and should use `minCommon - 1`. Consider if it checks:

`sourceParams`:

0. `position`
1. `standing`
2. `forceAction` optional

and

`targetParams`:

0. `position`
1. `forceAction` optional

Here, `minCommon=2`, so we should start checking the `sourceParams` at `standing`, since we don't have to order it like:
`sourceParams`:

0. `position`
1. `forceAction` optional
2. `standing`

unity/Assets/Scripts/AgentManager.cs
mattdeitke: Yeah, that's true here. Was using it for debugging, since it was saying "Vector3.normalized" gave a self-recurring loop.

unity/Assets/Scripts/ActionDispatcher.cs
ekolve: I'm not sure I understand the need for this change.  If we consider your first example:

0. `position`
1. `standing`
2. `forceAction` optional

vs

0. `position`
1. `forceAction` optional

The first loop that iterates up through minCommon will set signatureMatch=False, because it will iterate up to element 1 (standing vs forceAction) and these will fail the type check.  

The purpose of this bit of code is to effectively allow overriding actions that have common initial set of parameters.  So something like:

MyAction(float, float) could be overridden by MyAction(float,float, bool=True) in a subclass.  The idea is that this behavior would offer the least amount of surprise to someone editing actions across the hierarchy.

mattdeitke: The above loop was only checking for the type, so, since `standing` is a boolean and `forceAction` is a boolean, so `signatureMatch` never actually got set to `false`.

ekolve: Ahh, I see. You were getting a false positive for signature match when you shouldn't have.  Okay, then the use case that I think will break with your change is this:

source:
0. `param1` (float)
1. `param2` (float)
2. `param3` (bool=True) optional

target:
0. `param1` (float)
1. `param2` (float)

The first check will succeed since the minCommon number of params have identical types.  The new block will get entered since the number of params is different.  Within here the signatureMatch would get set to false since both source[minCommon-1] and target[minCommon-1] don't have default values, but we do want this to be considered a match since if both methods are available to be dispatched to its ambiguous about which one to pick if you pass just param1 and param2, which is why we allow this replacement if the source method was defined lower in the class hierarchy.

Its my fault for never writing a test for this (I am going to add one on master).

@Lucaweihs, what are you thoughts about 1. the existing override logic and 2. making the override logic sensitive to the names and types instead of just the types, which should make it trivial to support the case above.

ekolve: Okay, then the use case that I think will break with your change is this:

source:

0. `param1` (float)
1. `param2` (float)
2. `param3` (bool=True) optional

target:

0. `param1` (float)
1. `param2` (float)

The first check will succeed since the minCommon number of params have identical types.  The new block will get entered since the number of params is different.  Within here the signatureMatch would get set to false since both source[minCommon-1] and target[minCommon-1] don't have default values, but we do want this to be considered a match since if both methods are available to be dispatched to its ambiguous about which one to pick if you pass just param1 and param2.  In the signature match block we check whether the source and target params match in length and then throws an AmbiguousActionException.  The main purpose of all this signatureMatching is to 1. look for more specific matches in the class hierarchy (with the same number of params). 2. Look for ambiguous matches such as this one.

ekolve: I originally wrote the signatureMatch logic to try to resemble the behavior that csharp has with respect to method dispatching, but we dispatch exclusively using parameterNames.  I wonder if we just switched to the following logic whether it would be sufficient.  

a = {source.paramNames}
b = {target.paramNames}

c = a - b 

If the parameterNames in set c all have defaults, then we consider source and target a match.  I think this satisfies both of the requirements above.

mattdeitke: For now, I've just gone back to what it was, because I moved Teleport and TeleportFull to protected methods from the base.

I agree that the suggested set differences approach is the way to go.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: I have moved TeleportFull and Teleport to protected methods from the base, so this is now no longer necessary.

winthos:
looks good to me

ai2thor/controller.py
```diff
@@ -830,10 +830,12 @@ def step(self, action=None, **action_args):
...
         self.last_action = action
 
+        # dangerously converts rotation(float) to rotation(dict(x=0, y=float, z=0))
+        # this should be removed when ServerActions have been removed from Unity
+        # for all relevant actions.
         rotation = action.get("rotation")
-        if rotation is not None and type(rotation) != dict:
-            action["rotation"] = {}
-            action["rotation"]["y"] = rotation
+        if rotation is not None and not isinstance(rotation, dict):
```
ekolve, line 837: excellent! much better than type != dict

ekolve (approved):
lgtm

PR #620, opened 2021-02-17
Updates to SetObjectPoses

winthos:
-adding improved error messages returned on failed SetObjectPoses

-adding `placeStationary` flag to SetObjectPoses, allowing objects to be posed with kinematic = true, preventing additional physics resolution after repositioning

### DIFF
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -4052,7 +4052,7 @@ namespace UnityStandardAssets.Characters.FirstPerson {
         // a subsequent scene reset()
         protected IEnumerator setObjectPoses(ObjectPose[] objectPoses, bool placeStationary){
             yield return new WaitForEndOfFrame();
-            bool success = physicsSceneManager.SetObjectPoses(objectPoses, errorMessage, placeStationary);
+            bool success = physicsSceneManager.SetObjectPoses(objectPoses, out errorMessage, placeStationary);
             actionFinished(success, errorMessage);
         }
 
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -4043,17 +4043,17 @@ namespace UnityStandardAssets.Characters.FirstPerson {
                 actionFinished(false);
                 return;
             }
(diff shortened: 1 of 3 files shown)


unity/Assets/Scripts/PhysicsSceneManager.cs
```diff
@@ -292,28 +292,55 @@ public class PhysicsSceneManager : MonoBehaviour {
         RequiredObjects.Remove(sop.gameObject);
     }
 
-    public bool SetObjectPoses(ObjectPose[] objectPoses) {
+    public bool SetObjectPoses(ObjectPose[] objectPoses, string errorMessage, bool placeStationary) {
```
ekolve, line 295: Is the intent here to update modify the errorMessage parameter so that the caller receives an errorMessage on failure? In which case you need to add the `out` keyword to the parameter.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4043,17 +4043,17 @@ public void SetObjectPoses(ServerAction action)
...
         // SetObjectPoses is performed in a coroutine otherwise if
         // a frame does not pass prior to this AND the imageSynthesis
         // is enabled for say depth or normals, Unity will crash on 
         // a subsequent scene reset()
-        protected IEnumerator setObjectPoses(ObjectPose[] objectPoses){
+        protected IEnumerator setObjectPoses(ObjectPose[] objectPoses, bool placeStationary){
             yield return new WaitForEndOfFrame();
-            bool success = physicsSceneManager.SetObjectPoses(objectPoses);
-            actionFinished(success);
+            bool success = physicsSceneManager.SetObjectPoses(objectPoses, errorMessage, placeStationary);
```
ekolve, line 4055: Add `out` keyword to errorMessage (assuming you want the value updated).

ekolve (changes requested):

unity/Assets/Scripts/PhysicsSceneManager.cs
winthos: yes, that was the intention. Added the `out` keyword

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: yes, added

ekolve (approved):

Issue #483, opened 2020-09-28
Allow actions at (x, y) to return objectID on success


Tushar-N:
I'm using actions that accept x, y positions as inputs (e.g. `event = controller.step('OpenObject', x=0.2, y=0.5)`). Once it's executed, I'm trying to figure out what object was interacted with.

1. Is there any way to do it with the currently exposed methods?

2. Alternately, is there a way to just get the object that the raycast would hit? 
```
public bool ScreenToWorldTarget(float x, float y, ref SimObjPhysics target, bool requireWithinViewportRange)
```
does this and populates `target`, but this isn't an exposed method. Could this be made a callable easily?

Is there an easy way to do what I'm after? I thought I could just modify the interaction functions to return the target as a feedback parameter: e.g.`actionFinished(true) --> actionFinished(true, feedback)` like some methods that return info in `actionReturn`, but I'm finding it hard to parse recent code. PutObject has an `actionFinished(true)`, but OpenObject does not, so I got confused.

thomason-jesse:
This appears to be implemented now.
[ https://ai2thor.allenai.org/ithor/documentation/actions/object-interaction/#pickup-object ]

It can also be done using the instance segmentation mask, as seen here generalized to an array of pixel values (performs IOU between provided pixel mask and instance segmentation to select the target object)
[ https://github.com/askforalfred/alfred/blob/master/env/thor_env.py#L490 ]

thomason-jesse:
Did this feature ever pan out?

It would be possible to detect the altered object by comparing the objects' metadata before and after the action (for successfully executed actions), but that's computationally pretty tedious.

mattdeitke:
I don't necessarily think returning the `objectId` will work in every case, since some actions (i.e., `TouchThenApplyForce`) already return information to the `actionReturn`.

What I can add is a query action to the environment for the object at `(x, y)`:
```python
controller.step('GetObject', x=x, y=y, forceAction=True/False)
```
where `forceAction` specifies if visibility/interactability distance to the object should be ignored.

If an object is there, the `objectId` is returned inside of `actionReturn`. If not, the action return will be either null or empty. Then, if an `objectId` is returned, you can guarantee that calling:
```python
controller.step('OpenObject', x=x, y=y, forceAction=True/False)
```
targets an object, or can equivalently pass in the returned objectId, which is marginally faster:
```python
objectId = controller.step('GetObject', x=x, y=y, forceAction=True/False).metadata['actionReturn']
if objectId:
[...]

PR #623, opened 2021-02-19
Add GetObjectInFrame(x: float, y: float, forceAction: bool)

mattdeitke:
Basically exposes the functionality for `getTargetObject(x, y)`, as requested in #483.

Also, I plan to deprecate every single action that uses `x`, `y`, and `objectId`, in favor of people first calling:
```python
query = controller.step("GetObjectInFrame", x=x, y=y, forceAction=True)
if query:
    objectId = query.metadata['actionReturn']
    controller.step('OpenObject', objectId=objectId)
```

Note that I have overwritten `__bool__` so that it points to `event.metadata['lastActionSuccess']`.

### DIFF
--- a/ai2thor/server.py
+++ b/ai2thor/server.py
@@ -122,6 +122,9 @@ class Event(object):
 
         self.events = [self]  # Ensure we have a similar API to MultiAgentEvent
 
+    def __bool__(self):
+        return self.metadata["lastActionSuccess"]
+
     def __repr__(self):
         """Summarizes the results from an Event."""
         action_return = str(self.metadata["actionReturn"])
--- a/ai2thor/server.py
+++ b/ai2thor/server.py
@@ -33,6 +33,9 @@ class MultiAgentEvent(object):
         self.third_party_camera_frames = []
         # XXX add methods for depth,sem_seg
 
+    def __bool__(self):
+        return bool(self._active_event)
+
     @property
     def cv2img(self):
         return self._active_event.cv2img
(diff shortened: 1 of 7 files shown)


winthos:
Looks good to me. I assume this syntax will be easier to use rather than passing in object targeting information into each action? That's my only concern, which way is easier to read and use.

mattdeitke:
I don't necessarily know if its easier to read, but I still think it improves a few things better:
1. Making actions more modular. Instead of having a single action that does 2 things, break it down into easier pieces.
2. Will remove the need for overloading these actions with many parameters if there is a python api, since python does not support overloading.

ai2thor/server.py
```diff
@@ -122,6 +122,9 @@ def __init__(self, metadata):
 
         self.events = [self]  # Ensure we have a similar API to MultiAgentEvent
 
+    def __bool__(self):
```
ekolve, line 125: You should add a __bool__ method into MultiAgentEvent that delegates to self._active_event.__bool__

ekolve (changes requested):

ekolve:
lgtm - i'm fine with the Windows changes as well.  I requested one small change WRT to the new __bool__ overload.  

In terms of documentation, you should consider waiting until the next pip release is made since its currently awaiting validation for arm functionality.  As well, the new __bool__ overload should be documented.

ai2thor/server.py
mattdeitke: added: https://github.com/allenai/ai2thor/pull/623/commits/2cb3ca45f0721b13003ff3f13cdc7c5b7cbc2877

ekolve (approved):

Issue #540, opened 2020-12-08
Feature request: PutObject with (x, y) should influence target location


thomason-jesse:
With the new PutObject interface that allows specifying a normalized (x, y) position on the screen where the receptacle object is, it would be awesome if that (x, y) position were further propagated to select a resting position for the pickupable object on the receptacle object. 

My understanding is that currently a series of candidate positions are constructed on the receptacle object where the pickupable object can fit, and that these are sorted randomly before one is selected as the target location for the pickupable object. Instead of a random sort, with the (x, y) specification, these could be ordered by proximity to the specified (x, y) so that the pickupable object goes close to the location specified by the API call.

This feature would reduce friction with the API call for PutObject, especially in cases where the receptacle object is large enough that placement happens at the corner of / off screen (e.g., L shaped counters, floors, large tables).

winthos:
This is definitely some functionality that can be added in. I will make a note to add this to the list of features on the to-do list!

thomason-jesse:
I noticed this didn't make it into 2.5.3. Any chance for it on a soon-ish commit?

aishwaryap:
A related request. Would it possible to have an exposed method that converts an (x, y) on the egocentric image to an (x, y, z) in scene space using `raycast`. If we had this information, we could use a combination of `GetSpawnCoordinatesAboveReceptacle` and `PlaceObjectAtPoint` to perform the kind of placement requested in the main message. In addition, we can do things like find objects "near" rather than at the clicked location (helpful for small objects like pencils).

thomason-jesse:
@winthos  I've put in a pull request that resolves this issue.
https://github.com/allenai/ai2thor/pull/627

ekolve:
PR #684 resolves this issues.

PR #627, opened 2021-02-21
adds heuristic to PlaceObject close to RaycastHit location 

thomason-jesse:
For PlaceObject(x, y) calls, the placed object should be put close to the RaycastHit location, rather than at a random free spawn point on the receptacle (the current code). This pull request implements the distance from the RaycastHit as the heuristic for choosing an object spawn point on the receptacle, backing off to random as before if PlaceObject(objectId) is used.


thomason-jesse:
Would resolve this issue:

https://github.com/allenai/ai2thor/issues/540

thomason-jesse:
Another reason to add functionality like this:

When using a combination of GetSpawnCoordinatesAboveReceptacle / GetCoordinateFromRaycast / PlaceObjectAtPoint, placement can still fail frequently since PlaceObjectAtPoint requires invoking the physics engine, where the `placeStationary` ensures objects don't move around.

winthos:
This functionality has been added as of commit 6557027c0a16a9daf8d81ad16e0515c44e8622c6

Issue #618, opened 2021-02-16
reachablePositions in event.metadata is empty (2.7.0)


lajanugen:
I recently upgraded to 2.7.0 and have some issues with the reachable positions variable in event metadata - it seems to be empty. Issues can be reproduced as follows,

```
import ai2thor.controller
controller = ai2thor.controller.Controller()
e = controller.step('GetReachablePositions')
assert e.metadata['reachablePositions'] == []
```

mattdeitke:
Hey, please use `actionReturn` instead of `reachablePositions` for now. Using `reachablePositions` will probably throw an exception soon.

ekolve:
With PR #628 `reachablePositions` is remapped to `actionReturn`. Closing issue as it appears to be resolved.

PR #628, opened 2021-02-22
Metadata Objects and Aliases

mattdeitke:
## Summary

Allows us to edit the metadata from the Python side, which makes it possible to deprecate and rename things, without breaking backwards (and sideways -- with respect to a Unity build not using it's intended Python API) compatibility.

This PR also removes `reachablePositions` from the Unity metadata and aliases `.metadata["reachablePositions"]` to `actionReturn` if `GetReachablePositions` has been called (otherwise `[]`). This addresses https://github.com/allenai/ai2thor/issues/618.

## Future `visible` Alias Note

In the future, I plan to alias `visible` in the object metadata to `isInteractable`, since `visible` doesn't actually mean `visible` in the frame, rather, is both visible in the frame and is within the initialized `visibilityDistance`. This should be done at the same time that `visibilityDistance` gets aliased to something like `interactableDistance`.

This is not, however, as immediate of a priority, since everything still works fine with `visible` and `visibilityDistance`.


ai2thor/tests/test_unity.py
```diff
@@ -203,10 +201,10 @@ def test_fast_emit(controller):
...
     assert event.metadata["actionReturn"] is None
     assert event_fast_emit.metadata["actionReturn"] == "foo"
-    assert id(event.metadata["objects"]) == id(event_fast_emit.metadata["objects"])
-    assert id(event.metadata["objects"]) != id(event_no_fast_emit.metadata["objects"])
-    assert id(event_no_fast_emit_2.metadata["objects"]) != id(
-        event_no_fast_emit.metadata["objects"]
+    assert id(event._metadata["objects"]) == id(event_fast_emit._metadata["objects"])
+    assert id(event._metadata["objects"]) != id(event_no_fast_emit._metadata["objects"])
+    assert id(event_no_fast_emit_2._metadata["objects"]) != id(
+        event_no_fast_emit._metadata["objects"]
```
mattdeitke, line 207: This change was made because `.metadata` now creates a new object from `._metadata` after it is called for the first time from an event. So, the `id` of the newly created `.metadata` object of course changes.

ekolve:
Can you try running the following benchmark?
```python

import msgpack
import time
import json
import gc
import ai2thor.controller
from ai2thor.server import MetadataDict, MetadataList
import io

gc.disable()

c = ai2thor.controller.Controller()

metadata = c.last_event.metadata
for o in metadata['objects']:
    o['isInteractable'] = o['visible']

class ObjectMetadata(dict):

    def __init__(self, inner_dict):
        self.inner_dict = inner_dict

    def __getitem__(self, key):
        return self.inner_dict[key]

class Metadata(dict):

    def __init__(self, inner_dict):
        self.inner_dict = inner_dict

    def __getitem__(self, key):
        return self.inner_dict[key]


raw_data = msgpack.dumps(metadata)

def bench_baseline():
    d = msgpack.loads(raw_data, raw=False)

def bench_metadata_simple():
    d = Metadata(msgpack.loads(raw_data, raw=False))
    d['objects'] = [ObjectMetadata(o) for o in d['objects']]
    for i in range(len(d['objects'])):
        o = d['objects'][i]
        o['isInteractable']

def bench_metadata():
    d = msgpack.loads(raw_data, raw=False)
    value = MetadataDict(
        child_metadata=d,
        key_sequence=tuple(),
        cached_key_sequences=dict()
    )
    for i in range(len(value['objects'])):
        o = value['objects'][i]
        o['isInteractable']


def bench(method, count ):
    s = time.time()
    for i in range(count):
        method()
    e = time.time() - s
    print("[%s] AVG %s" % (method.__name__, e/count))

def bench_unpacker():
    data = io.BytesIO(raw_data)
    unpacker = msgpack.Unpacker(data, raw=False)
    map_count = unpacker.read_map_header()
    f = dict()
    for _ in range(map_count):
        k = unpacker.unpack()
        if k == 'objects':
            object_header = unpacker.read_array_header()
            f['objects'] = []
            for _ in range(object_header):
                o = ObjectMetadata(unpacker.unpack())
                f['objects'].append(o)
        else:
            f[k] = unpacker.unpack()
    for i in range(len(f['objects'])):
        o = f['objects'][i]
        o['isInteractable']

bench(bench_unpacker, 1000)
bench(bench_metadata, 1000)
bench(bench_baseline, 1000)
bench(bench_metadata_simple, 1000)
```

ekolve:
Based on the above results I would recommend simply wrapping each object in metadata['objects'] and also wrapping metadata and then within __getitem__ of either ObjectMetadata or Metadata perform any aliasing or deprecation.  The difference between wrapping each time vs. just loading without wrapping was around 20 - 30 microseconds.

mattdeitke:
Hmm.. Will check it out. But, I don't like the deprecated keys still appear in the metadata. For instance, `visible` appears when you print out `event.metadata['objects']`. I would prefer to completely hide it upon deprecation (and have it be silently aliased to `isInteractable` for backwards compatibility purposes). It would be quite confusing if somebody calls `print(event.metadata['objects'])`, sees `visible` is a key, then expects it to mean `visible` in the frame.

ai2thor/server.py
```diff
@@ -85,15 +85,39 @@ def unique_rows(arr, return_index=False, return_inverse=False):
...
+            warnings.warn(
+                'The key event.metadata["reachablePositions"] is deprecated and has been remapped to event.metadata["actionReturn"].'
+            )
+            x = "actionReturn"
+        elif x == "reachablePositions":
+            raise IndexError(
+                "You are trying to access event.metadata['reachablePositions'] without first "
+                + "calling controller.step(action='GetReachablePositions'). Also, "
+                + "the key 'reachablePositions' is deprecated in favor of event.metadata['actionReturn']."
+            )
```
mattdeitke, line 108: Note: I am intentionally throwing an error if they try to access `reachablePositions` without first calling `action="GetReachablePositions"`.

ai2thor/tests/data/metadata-schema.json
```diff
@@ -521,9 +521,6 @@
         "colorBounds": {
             "type": "null"
         },
-        "reachablePositions": {
-            "type": "array"
-        },
```
mattdeitke, line 526: `"reachablePositions"` now no longer appears as a key in the metadata. However, it will silently alias to `actionReturn` if "GetReachablePositions" is the last action.

ai2thor/tests/test_unity.py
```diff
@@ -843,6 +836,27 @@ def test_get_scenes_in_build(controller):
...
+
+    assert "reachablePositions" not in event.metadata.keys()
+    event = controller.step("Pass")
+    try:
+        event.metadata["reachablePositions"]
+        assert (
+            False
+        ), "reachablePositions shouldn't be available without calling action='GetReachablePositions'."
+    except:
+        pass
```
mattdeitke, line 857: There are 3 tests happening here, each of which is intentional:
1. Upon calling `GetReachablePositions`, calling `event.metadata["reachablePositions"]` should silently alias to `actionReturn` (which contains the reachable positions).
2. `reachablePositions` should never appear as a key in `event.metadata` because it is now deprecated (in favor of using `actionReturn` directly). However, calling `event.metadata["reachablePositions"]` will now be sideways and backwards compatible.
3. If a user does not call `GetReachablePositions` and tries to access `event.metadata["reachablePositions"]`, instead of the old way that returned an empty list, an `IndexError` is thrown. This now removes that silent bug.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -429,18 +428,18 @@ protected void HideAllAgentRenderers()
             return reachablePos;
         }
 
-        public void GetReachablePositions(int maxStepCount = 0) {
-            if(maxStepCount != 0) {
-                reachablePositions = getReachablePositions(1.0f, maxStepCount);
+        public void GetReachablePositions(int? maxStepCount = null) {
+            Vector3[] reachablePositions;
+            if (maxStepCount != null) {
```
Lucaweihs, line 433: When using nullable types, I prefer doing something like `maxStepCount.HasValue` rather than`maxStepCount != null` as really a nullable value is never really "equal" to null, it just fails to have a value.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -429,18 +428,18 @@ protected void HideAllAgentRenderers()
...
+                reachablePositions = getReachablePositions(maxStepCount: (int) maxStepCount);
```
Lucaweihs, line 434: Here I'd also prefer `maxStepCount.Value`.

Lucaweihs (commented):
Minor comments about nullable values, otherwise LGTM.

unity/Assets/Scripts/BaseFPSAgentController.cs
mattdeitke: Fixed in https://github.com/allenai/ai2thor/pull/628/commits/f4f78b083cfbda37c743642548cacc5c72606492.

unity/Assets/Scripts/BaseFPSAgentController.cs
mattdeitke: Fixed in https://github.com/allenai/ai2thor/pull/628/commits/f4f78b083cfbda37c743642548cacc5c72606492.

ai2thor/server.py
```diff
@@ -85,15 +85,39 @@ def unique_rows(arr, return_index=False, return_inverse=False):
...
+class MetadataWrapper(dict):
+    def __init__(self, raw_metadata: dict):
+        super().__init__(raw_metadata)
+        self._raw_metadata = raw_metadata
+
+    def __getitem__(self, x):
+        # alias deprecated functionality
+        if (
+            x == "reachablePositions"
+            and self._raw_metadata["lastAction"] == "GetReachablePositions"
```
ekolve, line 97: Doesn't GetSceneBounds also depend on reachablePositions?

ekolve (commented):
lgtm, just one comment about GetSceneBounds

ai2thor/server.py
mattdeitke: No. It looks like it just encapsulates the bounds of all Renderer's (i.e., object meshes that exist in the scene):
https://github.com/allenai/ai2thor/blob/main/unity/Assets/Scripts/AgentManager.cs#L370

image

mattdeitke: Fixed in https://github.com/allenai/ai2thor/pull/628/commits/184c7e8ddfa82a17f84cc10d260098df65b94544

PR #634, opened 2021-02-23
Extending the "GetReachablePositions" method to also allow for using local (agent-orientation-based) directions.

Lucaweihs:
`GetReachablePositions` returns positions relative the global coordinate grid. This is a bit constraining if I'd like my agent to be able to start at a rotation that is not in `[0, 90, 180, 270]`. This PR provides a few small changes to allow for using the left/right directions from the agent's transform instead of the world coordinates when a `directionsRelativeAgent` boolean is passed as `true` to a `GetReachablePositions` call.


ai2thor/tests/test_unity.py
```diff
@@ -1003,6 +1003,27 @@ def test_get_interactable_poses(controller):
     ), "GetInteractablePoses with large maxDistance is off!"
 
 
+@pytest.mark.parametrize("controller", [wsgi_controller, fifo_controller])
+def test_get_reachable_positions(controller):
+    controller.reset("FloorPlan28")
+
+    event = controller.step("GetReachablePositions")
+    assert len(event.metadata["actionReturn"]) == 116
```
mattdeitke, line 1011: From experience with `GetInteractablePoses`, this is probably going to break soon based on slight changes to Unity and non-determinism issues. I would each of these something like:

```python
event = controller.step("GetReachablePositions")
positions = event.metadata["actionReturn"]
assert 100 < len(positions) < 125
```
just to give it a bit more breathing room.

Then for `directionsRelativeAgent=True`, also assert that it has fewer positions than when `directionsRelativeAgent=False`.

mattdeitke (changes requested):

ai2thor/tests/test_unity.py
Lucaweihs: 👍 as you predicted, this non-determinism problem happened immediately (failed test with a small mismatch, `111!=112`). I've gone ahead and improved the tests so that there is a range.

roozbehm: Why can't we make GetInteractablePoses deterministic? Everything is static at the start. What makes it non-deterministic?

Lucaweihs:
@mattdeitke / @ekolve any more thoughts or can I merge this?

mattdeitke (approved):
LGTM. One note I think we might have discussed earlier, the name `GetReachablePositions` doesn't really fit anymore, since it doesn't work with `rotateStepDegrees` and can't, because if rotate step degrees is, say, 70 degrees, then there may be an infinite number of reachable positions.

Not in this PR, but a future one, I think it's perhaps worth changing (in a backwards + sideways compatible way) to be `GetGridPositions`.

ai2thor/tests/test_unity.py
```diff
@@ -1279,4 +1278,32 @@ def test_get_coordinate_from_raycast(controller):
     assert_near(
         query.metadata["actionReturn"],
         {'x': -0.5968407392501831, 'y': 1.5759981870651245, 'z': -1.0484200716018677}
-    )
\ No newline at end of file
+    )
+
+
+@pytest.mark.parametrize("controller", fifo_wsgi)
+def test_get_reachable_positions(controller):
```
ekolve, line 1285: It looks like a test with this method name already exists.  Can you rename the new one?

ekolve (changes requested):
lgtm, one comment about the duplicate test name.

ai2thor/tests/test_unity.py
Lucaweihs: Done

PR #640, opened 2021-03-03
using ContractResolver to handle Vector3 in DynamicServerAction

ekolve:


mattdeitke (approved):
It looks like this solved my issue too, which was just running `telefull` from master's current debug input field :)

PR #641, opened 2021-03-03
Add GetCoordinateAtPixel query action

mattdeitke:
For a given (x, y) point in the last frame, it returns the global coordinate of the object that is first hit by a RayCast going through that (x, y) point.

For instance, let `x=0.25` and `y=0.5` in the image below. The actionReturn would be the first global 3D point hit by a RayCast at `x=0.25` and `y=0.5`, or `x=-0.5968, y=1.5759, z=-1.0484` as read in the editor, in this case.
image

Requested in https://github.com/allenai/ai2thor/issues/483#issuecomment-786962805 and previously proposed by @Lucaweihs (but it ended up not being added to master because his only use for it was debugging).

### DIFF
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -1164,5 +1164,5 @@ def test_get_coordinate_at_pixel(controller):
     query = controller.step("GetCoordinateAtPixel", x=0.25, y=0.5)
     assert_near(
         query.metadata["actionReturn"],
-        {"x": -0.596839666, "y": 1.57599819, "z": -0.668121457},
+        {'x': -0.5968407392501831, 'y': 1.5759981870651245, 'z': -1.0484200716018677}
     )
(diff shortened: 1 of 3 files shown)


winthos:
Looks good to me

winthos (approved):

Lucaweihs:
What's the domain of the x/y input coordinates, is it [0:1]? If so I disagree with the name `GetCoordinateAtPixel` as we're not specifying pixels but screen space coordinates.

mattdeitke:
Yes, [0:1]. I don't have a huge preference for the name. I agree it's a bit weird because "AtPixel" is not technically referring to a pixel. Perhaps `GetCoordinateFromCast` or `GetCoordinateInFrame` or something related.. although I don't love these.

thomason-jesse:
This would unblock me for a number of things and would enable an API-only workaround for
https://github.com/allenai/ai2thor/pull/627
in combination with `GetSpawnCoordinatesAboveReceptacle` and `PlaceObjectAtPoint`

thomason-jesse:
Thanks for the suggestion; Luca. I can confirm the PR works well for me 👍

PR #643, opened 2021-03-05
Allow Agent Manager actions (e.g., Initialize, Reset, AddThirdPartyCamera) to be called by "run" in the debug input field

mattdeitke:
This should not change anything other than allowing improvements to the debug input.

There is also now a way to print out initialization and step commands from Colab, for each `step` command and `reset` command:
https://colab.research.google.com/drive/10iptbuuukc3bMCE5hzl4SvJGO3jIXITP?authuser=2#scrollTo=Nazs8kUi1nGR

(requested by @elimvb)


winthos (approved):
looks good to go

PR #648, opened 2021-03-10
fix toward ScreenToWorldTarget()

winthos:
**Bug Fixes**

- When casting against the default layer (0<<1) in ScreenToWorldTarget, there are cases where a collider on an object like Cabinets and Drawers are on the Default layer, causing the raycast to be blocked from certain angles, resulting in a false "Object not found!" error even though the sim object is within the unobstructed path of the ray. Adding a check to check the collider hit by the ray and all parents of that object to see if it is a sim object in order to account for this.

- Fixed forceAction = True actually doing something when targeting objects via (x,y) screen coordinates. Previously, whether forceAction was true or not did nothing, and targeted objects by the screen to world ray could have returned null due to distance checks. Now, if forceAction =True, any object hit will be returned regardless of distance to the agent. This means we can now do things like screencast to an object in the viewport that is across the room and out of reach, but still pick it up. This brings the (x,y) targeting behavior in line with passing target by objectId if forceAction is true.

- Fixed logic error where (x,y) targeting did not use the same `isVisible` logic as passing by objectId. Originally, when an object was targeted via a screen to world raycast, it did not check if an object was visible according to the logic of visibility points. This could lead to weird cases where, if passing a target by objectId, the action would succeed, but hitting the object at the "right" spot with (x,y) would cause this same action to fail. This was because the old logic only checked for if the point hit by the raycast was within visibility distance or not, it did not account for at least some part of the object being visible. Because the visibility logic is "if any one visibilty point is within range to the agent and not obstructed, the object is visible" this meant that the old logic could hit a point on a potential target object that was outside the visible range, but if that object was big enough (think a long table) then the closer half of that object would be visible to the agent anyway. Now, this check correctly considers the entire visibility logic and not just `maxVisibleDistance`. 

**New Features**

- Updates to the `GetObjectInFrame` action. Error messages have been improved. If no sim object was hit at the coordinate, it now returns `no sim objects found at ({x}, {y})` as both the `errorMessage` and the `actionReturn`. If an object is found, the objectId is returned in `actionReturn`. Additionally, there is now a `checkVisible` param that can be passed into `GetObjectInFrame`. The default behavior of `GetObjectInFrame` only checks if the object hit at that raycast coordinate is within the viewport, but just because an object is in the viewport does not mean it is `isVisible = true` to the agent. Now, if you pass in `checkVisible = true` to `GetObjectInFrame` it will perform an additional check based on the same logic of object visibility using vis points. While `checkVisible = true`, if a valid target is hit but it is not currently visible to the agent, an `errorMessage` and `actionReturn` of `target hit ({target.objectID}) at ({x}, {y}) is not currently Visible to Agent` will be set and returned.


mattdeitke:
I'm a bit concerned about this quick fix for the following reason: Suppose the raycast hits an object with `SimObjPhysics` as a component. Won't `GetComponentInParent` now only look in the parent for the component, and not itself?

Instead, why not just removing the "Default" layer as an option from the raycast's layermask, and force it to hit a "SimObjVisible" or "PlaceableSurface" layerMask? Wouldn't that solve the problem of a Default mask getting hit by mistake, causing one to have to look in the parent?

winthos:
Note that the GetComponentInParent also checks the game object itself and then recursively checks parents. So the behavior is the same except it also checks in the parents for a potential sim object component https://docs.unity3d.com/ScriptReference/GameObject.GetComponentInParent.html

mattdeitke:
FYI, the checks are now failing when the exhaustive test is included. There look to still be disparities between instance masks and ScreenToWorldTarget.

winthos:
I've made some updates to show the more involved changes on this PR to the original comment.

Also @mattdeitke should we just remove the pixel perfect test case for now, or do we want to update it with the 2-3 pixel buffer that we talked about? Currently the build is still failing because of this test.

mattdeitke:
I will update the pixel test today.

mattdeitke:
Ok. I have given a large buffer (i.e., generously chopping off 5 pixels from the edge of each object's instance segmentation mask) and it looks like it is still failing, but now at a different spot (i.e.,: `GetObjectInFrame` says GarbageCan, while instance masks say Floor):
```
Failed at (0.7866666666666666, 0.92) for Floor|+00.00|+00.00|+00.00 with agent at: {'name': 'agent', 'position': {'x': -1.5, 'y': 0.9009982347488403, 'z': -1.5}, 'rotation': {'x': -0.0, 'y': 180.0, 'z': 0.0}, 'cameraHorizon': -0.0, 'isStanding': True, 'inHighFrictionArea': False}
```

winthos:
Hmm interesting, I will investigate further and post updates here

PR #653, opened 2021-03-12
#webgl-deploy

mattdeitke:
Updates to enhance the demo.

Allows for:
* All scenes to be rendered in the demo
* Removes the default loading page (I override it with a custom loading page)
* Modifies/removes some keyboard commands so that they are controllable solely through javascript. Examples include ``` and `Escape`, which used to be how one started the demo. However, there are problems with the `Escape` keyword, since it's used by the browser to allow a user to show it's cursor (after we've hidden it, and we cannot override this browser command).

### DIFF
--- a/unity/ProjectSettings/ProjectSettings.asset
+++ b/unity/ProjectSettings/ProjectSettings.asset
@@ -13,7 +13,7 @@ PlayerSettings:
   useOnDemandResources: 0
   accelerometerFrequency: 60
   companyName: Allen Institute for Artificial Intelligence
-  productName: AI2-THOR
+  productName: AI2-Thor
   defaultCursor: {fileID: 0}
   cursorHotspot: {x: 0, y: 0}
   m_SplashScreenBackgroundColor: {r: 0.13725491, g: 0.12156863, b: 0.1254902, a: 1}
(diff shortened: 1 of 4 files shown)


unity/ProjectSettings/ProjectSettings.asset
```diff
@@ -13,15 +13,15 @@ PlayerSettings:
   useOnDemandResources: 0
   accelerometerFrequency: 60
   companyName: Allen Institute for Artificial Intelligence
-  productName: AI2-Thor
+  productName: AI2-THOR
```
ekolve, line 16: Can you revert this change for now? It will affect the executable name generated by the build which we assume to be 'AI2-Thor'.

ekolve (changes requested):

unity/ProjectSettings/ProjectSettings.asset
mattdeitke: Hmm... Perhaps, although this change was intentional. AI2-Thor is an incorrect spelling that doesn't look that good (should be either AI2-THOR or ai2thor).

That said, it doesn't affect the demo in any way, so I'm fine reverting it for now.

ekolve: I agree that the spelling should be corrected.  I think it got set very early in the life of the project but never changed.  Altering the spelling complicates things a bit since we assume the binary name to be 'AI2-Thor' currently.  This really only affects OSXIntel64 and can update how we look for the executable path to make it handle name changes properly.  Can you create an issue and assign it to me to support this?

PR #656, opened 2021-03-15
🎉 Material Swapping!

mattdeitke:
### Summary

- Adds new `RandomizeMaterials` and `ResetMaterials` actions. `RandomizeMaterials` only randomizes the materials within each object category. So, an Apple, for instance, won't ever have the texture of `Lettuce`, `Bread`, or an `Orange`. Calling `ResetMaterials` then provides the ability to go back to the original colors in the scene. _Note:_ Upon calling reset, `ResetMaterials` is not called by default.

- Adds a significant amount of additional materials to the collection of materials that can be randomized. Most notably, wall materials are now included, along with many other sim objects.

### Example Usage

```python
controller = Controller()

controller.step(action="RandomizeMaterials")
# {...}
controller.step(action="ResetMaterials")
```

### Remarks

1. I have tested this significantly and am truly blown away by how incredible the results are. Each shuffling pretty much looks like an entirely new room!

2. In a future PR, I think it's definitely worth scaling this functionality up by using materials from object datasets, just to increase domain randomization even more. Nicely, there is next to no extra work to add new material (whereas adding a new SimObject is much more work), so I can see this expanding pretty easily.

3. I plan to add this as a button to the [demo](https://ai2thor.allenai.org/demo), so one can easily see the results in high resolution every room, without having to code up anything.

### Backwards Compatibility

Since many materials of existing objects have been added, the old `RandomizeColors` action will now be able to randomize the colors of these materials too. Here, materials that were not present inside of `ColorChanger.cs` previously simply never changed in color. Hence, `RandomizeColors` will slightly break in terms of backwards compatibility.

I do not see this as an issue, at all. More of a bug fix. Moreover, since `RandomizeColors` is stochastic action by nature (i.e., not even a random seed for colors can be passed in), any current use of the action was _already_ stochastic.


unity/Assets/Scripts/ColorChanger.cs
```diff
@@ -1,84 +1,214 @@
...
-    public Material[] woodMaterials;
-
-    Material[] targetMaterials;
-    Material[] backgroundMaterials;
-    Material[] furnitureMaterials;
-    Material[] quickMaterials;
-
-    Material[] allMaterials;
-    Color[] origColors;
+    public Material[] alarmClockMaterials,
```
AlvaroHG, line 8: Materials were dragged here and stored in the prefab. For a future iteration we could create a json dictionary from object class to list of materials, and load them dynamically with asset `Resources.Load` or `AssetDatabase.LoadAssetAtPath`. Changing this text file would be easier and visible in PRs.

mattdeitke: This is true. Although the materials are stored in pretty random file locations, where metal materials alone might be spread out between, say. 20 different folders and dozens of different paths.

So there would have to be a major big refactor/cleanup before that would work and I'm not sure if there's an immediate benefit to this approach speed wise.

mattdeitke:
Making a draft until 2 things are ready:

1. Val/Test room textures are not used in the randomize process.
2. Legal approves/denies the use of adding large scale (CC0 licensed) texture datasets to the set of all textures. This greatly increases the amount of available textures.

PengkunLiu:
Hello, Mattdeitke! Sorry to disturb you. I would like to ask the RandomizeColors is available on AI2THOR 2.1.0? Because I can not find the relevant code in the codes of the controller part. But I could find the relevant codes in BaseFPSAgentController.cs.

mattdeitke:
> I would like to ask the RandomizeColors is available on AI2THOR 2.1.0?

Hi @PengkunLiu, material swapping is not yet available in AI2-THOR's latest version. I think I should have time to get it ready and merged to the latest version by next week. But, it is not currently available in AI2-THOR version 2.1.0. It may be possible to add it to an older version like that, although I haven't looked into it yet.

PengkunLiu:
> > I would like to ask the RandomizeColors is available on AI2THOR 2.1.0?
> 
> Hi @PengkunLiu, material swapping is not yet available in AI2-THOR's latest version. I think I should have time to get it ready and merged to the latest version by next week. But, it is not currently available in AI2-THOR version 2.1.0. It may be possible to add it to an older version like that, although I haven't looked into it yet.
> 
> _Note: This is a new feature._

Thank you very much for your quick reply.  
How about event = controller.step('ChangeColorOfMaterials') or event = controller.step('RandomizeColors')  could work? I still can not find them in the codes of the controller part.  Looking forward to hearing from you.

mattdeitke:
>  How about event = controller.step('ChangeColorOfMaterials') or event = controller.step('RandomizeColors') could work? I still can not find them in the codes of the controller part. Looking forward to hearing from you.

Unfortunately, these were both introduced after AI2-THOR version v2.1.0, so they are not in that version.

PengkunLiu:
> > How about event = controller.step('ChangeColorOfMaterials') or event = controller.step('RandomizeColors') could work? I still can not find them in the codes of the controller part. Looking forward to hearing from you.
> 
> Unfortunately, these were both introduced after AI2-THOR version v2.1.0, so they are not in that version.

Thank you very much for your explanation.

mattdeitke:
FYI: Planning on opening up a separate PR that adds external materials, since I want to check how this affects the memory, before adding them all.

However, PR should be ready to go. I'm still adding a few tests and documentation, but don't expect much to change.

mattdeitke (commented):
Adding some comments to speed up the reviewing process.

I've tested it quite a bit, and added in some new tests. Let me know if there are any questions. Would like to add it to the demo this week.

unity/Assets/Scripts/ColorChanger.cs
```diff
@@ -1,84 +1,287 @@
...
 
     public void Start() {
-        targetMaterials = alarmClockMaterials.Concat(appleMaterials).Concat(basketballMaterials).Concat(bowlMaterials).Concat(garbageBinMaterials).Concat(houseplantMaterials).Concat(pillowMaterials).Concat(sprayBottleMaterials).ToArray();
-        backgroundMaterials = bedMaterials.Concat(boxMaterials).Concat(cellphoneMaterials).Concat(cupMaterials).Concat(floorLampMaterials).Concat(penPencilMaterials).Concat(plateMaterials).Concat(potMaterials).Concat(statueMaterials).Concat(watchMaterials).ToArray();
-        furnitureMaterials = armchairMaterials.Concat(bedMaterials).Concat(chairMaterials).Concat(coffeeTableMaterials).Concat(deskMaterials).Concat(diningTableMaterials).Concat(dresserMaterials).Concat(officeChairMaterials).Concat(shelvingUnitMaterials).Concat(sideTableMaterials).Concat(sofaMaterials).ToArray();
-        quickMaterials = ceramicMaterials.Concat(fabricMaterials).Concat(glassMaterials).Concat(lightMaterials).Concat(metalMaterials).Concat(miscMaterials).Concat(plasticMaterials).Concat(woodMaterials).ToArray();
+        materials = new Dictionary<string, Material[]> {
+            ["AlarmClock"] = alarmClockMaterials,
```
ekolve, line 92: Can all of these strings be changed to SimObjType enum? Instead of `"AlarmClock"` use `SimObjType.AlarmClock`.

unity/Assets/Scripts/ColorChanger.cs
```diff
@@ -1,84 +1,287 @@
...
+    }
+
+    private void shuffleMaterials(List<Material> materialGroup) {
+        for (int n = materialGroup.Count - 1; n >= 0; n--) {
+            int i = Random.Range(0, n + 1);
+            swapMaterial(materialGroup[i], materialGroup[n]);
+        }
+    }
+
+    public int RandomizeMaterials(
```
ekolve, line 206: Can you add the option to pass in a randomSeed so that configurations can be repeated? For all the other random actions we offer the ability to pass in a seed for this reason.

ekolve (changes requested):

unity/Assets/Scripts/ColorChanger.cs
Lucaweihs: Yes, I don't think you actually need the `shuffleMaterials` function. At some point in the past I added the `ListExtensions.cs` file which adds a couple of different `.Shuffle_` methods (in-place shuffling) with various forms of seeding to the `List` class.

mattdeitke: @Lucaweihs, I don't think this shuffling method would work, because the swaps aren't just changing the order in the List/array, they're instead changing the underlying properties with:
```c#
    private void swapMaterial(Material mat1, Material mat2) {
        if (mat1.HasProperty("_MainTex") && mat2.HasProperty("_MainTex")) {
            Texture tempTexture = mat1.mainTexture;
            mat1.mainTexture = mat2.mainTexture;
            mat2.mainTexture = tempTexture;
        }

        if (mat1.HasProperty("_Color") && mat2.HasProperty("_Color")) {
            Color tempColor = mat1.color;
            mat1.color = mat2.color;
            mat2.color = tempColor;
        }
    }
```

mattdeitke: > Can you add the option to pass in a randomSeed so that configurations can be repeated?

I suppose. Although we won't be able to maintain the random seed in a backwards compatible way, since new materials will presumably be continuously added. So, this backwards seed can only be guaranteed to be reproducible when using the same build.

unity/Assets/Scripts/ColorChanger.cs
mattdeitke: Probably not well.. There are many materials that are not object types. For instance, `Grunge`, `Paper`, `Plastic`, `Metal`, `Glass` are not sim object types.

unity/Assets/Scripts/ColorChanger.cs
winthos: I think this is ok, we can just add a warning that seeds should be expected to be used within a version.

unity/Assets/Scripts/ColorChanger.cs
winthos: Yeah, since not every object with a material that may be swapped is a sim object, I think the way Matt has it set currently is the way to go. It is a bit more tedious if we ever need to update this to include new objects or materials, but it shouldn’t be too bad

winthos (approved):
This looks good to go from my perspective.

unity/Assets/Scripts/ColorChanger.cs
mattdeitke: Copying over a message I sent to @ekolve for completeness. Feel free to chime in.
TL;DR: I don't think random seed should be set inside of any action parameters. It should only be set globally.

----------------

So I have a problem with randomSeed at the moment. Suppose I have a loop:

```python
controller = Controller()
for episode in range(int(1e8)):
    controller.step(action="RandomizeMaterials", randomSeed=42)
    # Do something...
```

this has me expecting the materials will change once, and then never change after that because the seed is reset each time? (Here, from C#, you'd set Random.Seed = 42).That is how randomSeed is used in other actions, but this is very unnatural and basically introducing a silent bug, as one would expect calling RandomizeMaterials many times inside of this loop would constantly change the materials differently.I think there should instead be something like:

```python
controller = Controller(randomSeed=42)
for episode in range(int(1e8)):
    controller.step(action="RandomizeMaterials")
```

or

```python
controller = Controller()
controller.randomSeed = 42
for episode in range(int(1e8)):
    controller.step(action="RandomizeMaterials")
```

which is done in other libraries, which sets the random seed for everything up front, and then samples from that random seed later. But that might be beyond the scope of this PR.

ekolve: Setting a global random seed would lead to a lot of confusion about to reproduce a configuration.  If the seed is set once on startup, then all the actions that depend on the RNG must be called in the *exact* same order otherwise the configuration will be different.  Whats complicates things more is suppose that you don't want to call something like 'PutObject' (which has randomization) for a particular scene setup, then you basically won't be able to have the same seed for a subsequent call to another action that the RNG depends on.  As well, if the seed is set once for the Player, then all training, scene loading, etc. must occur in exactly the same order and each action must make the exact same number of calls to the RNG or a subsequent's action won't receive the same sequence of random numbers.

mattdeitke: After further discussion, we are moving this to a future PR.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -776,6 +773,196 @@ public Quaternion TargetRotation
...
+                    }
+                    if (!useValMaterials.HasValue) {
+                        useValMaterials = false;
+                    }
+                    if (!useTestMaterials.HasValue) {
+                        useTestMaterials = false;
+                    }
+                    if (!useExternalMaterials.HasValue) {
+                        useExternalMaterials = true;
+                    }
```
Lucaweihs, line 901: These statements can be simplified:
```c#
useTrainMaterials = useTrainMaterials.GetValueOrDefault(true);
useValMaterials = useValMaterials.GetValueOrDefault(false);
useTestMaterials = useTestMaterials.GetValueOrDefault(false);
useExternalMaterials = useExternalMaterials.GetValueOrDefault(true);
```
and similarly for those below.

unity/Assets/Scripts/SimUtil.cs
```diff
@@ -540,7 +541,115 @@ public static class SimUtil {
...
+            ["kitchen"] = new HashSet<Material>(),
+            ["livingRoom"] = new HashSet<Material>(),
+            ["bedroom"] = new HashSet<Material>(),
+            ["bathroom"] = new HashSet<Material>(),
+            ["robothor"] = new HashSet<Material>(),
+        };
+
+        // iTHOR scenes
+        foreach (int sceneType in new int[] {0, 200, 300, 400}) {
+            for (int i = 1; i <= 30; i++) {
```
Lucaweihs, line 575: Is there a way to parameterize this `30` value in case we add more scenes in the future?

ai2thor/tests/test_unity.py
```diff
@@ -1620,6 +1623,122 @@ def test_unsupported_manipulathor(controller):
...
+    )
+    assert controller.step(action="RandomizeMaterials", fromRoomTypes=["RoboTHOR"])
+    assert (
+        controller.last_event.metadata["actionReturn"]["totalMaterialsConsidered"]
+        == 320
+    )
+
+    controller.step(action="ResetMaterials")
+
+
```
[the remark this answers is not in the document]
Lucaweihs: It looks like this seeding is being pushed back to a future PR, can you add a TODO here for the above test?

Lucaweihs (commented):
Minor comments but otherwise LGTM.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -776,6 +773,196 @@ public Quaternion TargetRotation
...
+                if (fromRoomTypes != null) {
+                    if (!chosenRoomTypes.Contains("robothor")) {
+                        throw new ArgumentException(
+                            $"fromRoomTypes must include \"RoboTHOR\" inside of a RoboTHOR scene: {scene}. You gave: {fromRoomTypes.ToString()}."
+                        );
+                    }
+                }
+            }
+
+            switch (sceneType) {
```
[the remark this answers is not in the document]
mattdeitke: Yes. Switch statements are much better optimized than alternative representations when compiled to assembly.

unity/Assets/Scripts/SimUtil.cs
mattdeitke: Hmm.. probably not well. I suppose one could maybe look at the list of scenes, but suppose these scenes are in "unlisted/development only" mode (like 500 level hallways in the current set of scenes). Here, one wouldn't want to automatically include these.

ai2thor/tests/test_unity.py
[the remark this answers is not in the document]
mattdeitke: Fixed in https://github.com/allenai/ai2thor/pull/656/commits/6abc97c96cf31cd0536f2a50ba3c234e3955d322

unity/Assets/Scripts/BaseFPSAgentController.cs
mattdeitke: Fixed in https://github.com/allenai/ai2thor/pull/656/commits/4fb4558e21393ad3e557d7a9101a677ce8bf58bb

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -776,6 +773,196 @@ public Quaternion TargetRotation
...
+                            break;
+                        case 4:
+                            if (!chosenRoomTypes.Contains("bathroom")) {
+                                throw new ArgumentException(
+                                    $"fromRoomTypes must include \"Bathroom\" inside of a Bathroom scene: {scene}. You gave: {fromRoomTypes.ToString()}."
+                                );
+                            }
+                            break;
+                    }
+                }
```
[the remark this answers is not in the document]
mattdeitke: Fixed in https://github.com/allenai/ai2thor/pull/656/commits/777819ee7bd4c92b5aa23549b6bb7fabe56b8b14. I suppose that's fine for now, though perhaps makes things a bit more difficult if the scenes undergo a naming convention change.

PR #658, opened 2021-03-15
removing UNITY_EDITOR condition for StartCoroutine(EmitFrame) to make…

ekolve:
… it easier to connect Python to the Editor; added break in while look in the event Editor can't connect


winthos (approved):
looks ready to go!

PR #661, opened 2021-03-17
batch actions

ekolve:
This PR adds support for the following syntax:

```python

with controller.batch():
    controller.step('InitialRandomSpawn')
    controller.step('TeleportFull', position=...)
    controller.step('OpenObject', objectId=...)
```

All the steps will get executed very quickly without full metadata generation or image capture thanks to the fastActionEmit.  Roundtrip times will be on the order of 50-100 microseconds per step().  

A few things that I would like to discuss - I currently am not permitting nesting of batches - i.e. you can't do this:

```python
def mything(controller):
    with controller.batch():
        controller.step()

    controller.step() # oops, i'm might still be in batch mode

with controller.batch():
    controller.step()
    mything()

```

This is consistent with the behavior of https://docs.python.org/3/library/warnings.html#warnings.catch_warnings.  I think this is the right initial choice as its much easier to relax this in the future and will lead to less confusion about when block exits that you are no longer in a batch block.

The second issue I want to discuss is what metadata should be available to the user.  We will have actionReturn, errorMessage, and lastActionSuccess, but all the other fields could be potentially stale. For example, you could TeleportFull, followed by RotateRight, now the objects marked as visible are no longer valid and the agent's position is no longer valid.  We have a few options:

1. Strip on the python side the volatile keys (objects, agent, rgb frame, etc) regardless of the action that is getting executed.  This will lead to exceptions on the user's side when they attempt to access these fields, but they will never get surprised by using stale data. They will still be able to access the actionReturn, lastActionSuccess and errorMessage fields. 
2. Remove the metadata altogether. Set `controller.last_event = None` since  the frame is also stale and return None from any step() called within the batch() block.  This is the most conservative, but prevents the user from accessing the actionReturn value which will be useful depending on the action.
3. A variant of option 1, strip anything (even non-volatile keys) in the event/metadata that we don't explicitly know to be updated.  So you would get an Event with metadata and the only thing on the metadata would be actionReturn, errorMessage, lastActionSuccess. 
4. Do nothing - document that any step() within the batch() renders the event stale and let user's work with the data if they so choose.


Lucaweihs:
Fantastic addition to THOR. I don't mind disallowing nested batching, this seems very reasonable to me. Before diving into the the various metadata alternatives one question: why will the agent position be stale? Don't we need to run physics updates and what not to be able to do things like a `"RotateRight"` or `"MoveAhead"` anyway?

ekolve:
> Fantastic addition to THOR. I don't mind disallowing nested batching, this seems very reasonable to me. Before diving into the the various metadata alternatives one question: why will the agent position be stale? Don't we need to run physics updates and what not to be able to do things like a `"RotateRight"` or `"MoveAhead"` anyway?

The short answer is that I am using the fastActionEmit logic to achieve the speed, which means that object/agent metadata isn't retrieved after each step.  fastActionEmit was originally intended for actions like GetReachablePositions where agent/object metadata doesn't change between invocations of the action. 

In this case, there is no reason we couldn't grab the updated agent/object metadata, rgb/depth frame, etc.  It really comes down to how fast you want this to be.  If we grab all the same event data we grab without the batch() context manager the speed benefits will quickly evaporate and the performance will be no different from what we have today.

Lucaweihs:
Gotcha, how difficult would it be to allow the user to specify the fields to return when initializing the context manager? I suspect that it would be useful to know things like the agent position and whether or not the action was successful but that it would almost never make sense to get the rgb/depth frames. It would be cool, for instance, if I could write my own version of `GetReachablePositions` using this batch context and get within an order of magnitude of the speed of the C# version.

ekolve:
`GetReachablePositions` already uses fastActionEmit, so its about as fast as you can get (in terms of eliminating metadata data IO/generation), but that isn't really what you are asking.  I don't think it would be too hard to support specifying which fields you want returned, but I think that may end up making things a bit complicated.  The fields that impact performance the most are RGB/Depth and object metadata.  Object metadata can be limited with SetObjectFilter, but I'm not sure if thats an option here since access to the object metadata may be useful.

The other concern I have about this approach in general is that physics doesn't get run since it all gets executed in a tight loop during WaitForEndOfFrame.  We could call physics after every action, but that may be something we should be doing independent of batch() actions.

One really simple alternative to this version of batch(), would be that the context manager simply passes `renderImage=False` when inside the block.  This would still provide a decent speed improvement and keep the metadata consistent.  A list of objectIds could also be passed to batch() that we use with SetObjectFilter to further improve speed.  This would never be as fast as the fastActionEmit version, but metadata would be consistent and it would wrap a few of the techniques already in use for improving performance.

Lucaweihs:
Closing for now, may come back to this feature in the future.

PR #664, opened 2021-03-18
WebGL demo fixes

AlvaroHG:
* Fixes move speed, frame independent and consistent across editor and WebGL
* Fixes throw action
* Fixes scroll actions, WebGL was silently crashing because of the use of dynamic
* Fixes highlight shader states for faraway and nearby objects


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5011,10 +5018,12 @@ private IEnumerator checkDropHandObjectActionFast(SimObjPhysics currentHandSimOb
 
             GameObject go = ItemInHand;
             DropHandObject(action);
-            if (this.lastActionSuccess) {
+            // Force is not applied because action success from DropHandObject starts a coroutine that waits for the object to be stationary
+            // to return lastActionSuccess == true that is not what we want for throwing an object, review why this was that way
+            //if (this.lastActionSuccess) {
```
AlvaroHG, line 5023: @winthos just your input for this

winthos: Yeah I think we need to determine what counts as the action successfully completing. Is it when the object leaves the hand and is no longer being held, or is it when it comes to rest after having been to thrown?

I’m not sure which makes the most sense

AlvaroHG:
Merging into non-main branch

PR #668, opened 2021-03-19
💡 Randomized Lighting!

mattdeitke:
Introduces domain randomized lighting, with several parameters to control the extent with which lighting is randomized.

> Note that lighting is completely reset upon calling `controller.reset(scene)`.

## Example Usage

```python
controller = Controller()
controller.step(action="RandomizeLighting")
```

## RoboTHOR Examples

RoboTHOR has the most visible changes, particularly because there is no visible skybox in the scenes.

 

 

 

 

 

## iTHOR Examples

The egocentric images from iTHOR still end up changes the pixels quite a bit, even if its not as visible of a change to us.


Lucaweihs:
Very cool update. One main thought: would it be possible to have a bit more granularity to allow for randomly varying different aspects of the lighting? I.e., currently it seems that color randomization is sort of a "yes/no" but it would be quite cool to have it be a little more like the [ColorJitter](https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.ColorJitter) function from `torchvision` which lets the delta in brightness/hue/etc separately.

mattdeitke:
@Lucaweihs Added support ranged hue, saturation, and brightness values.

Similar to how PyTorch's color jitter works with both `float` and `tuple(float, float)`, I think it would be nice to also be able to do something similar, where `brightness = X`, corresponding to `brightness = [max(0, 1 -X), 1 + 0.5]`, along with `hue=Y, saturation=Z`, corresponding to `hue=(Y, Y), saturation=(Z, Z)` but I haven't found a good way to do this that:

- (a) works with the action dispatcher
- (b) isn't reassigned on the Python side -- which makes debugging annoying when C# and Python mismatch
- (c) can be written in such a way that passing calling the method without argument names is dis-ambiguous (e.g., one cannot have `RandomizeColors(float brightness, float hue, (float, float)? saturation)` while also having `RandomizeColors(float saturation, float hue, (float, float)? brightness)`, since calling `RandomizeColors(0, 1)` is then ambiguous to C#.

In C# 10.0, they are expected to support union types, so maybe then might be the best time.

winthos (approved):
looks good to go to me

Lucaweihs (approved):
LGTM

elimvb (approved):
LGTM

PR #669, opened 2021-03-20
Fixed getReachablePositions error message bug

elimvb:
I added an additional `continue` check to a foreach loop, so that a point on the floorGrid that had already failed a check for agent-reachability could break out of its iteration, before it could potentially return a false-positive error message at check #2 saying it had completed check #1 but failed #2.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -388,18 +388,21 @@ namespace UnityStandardAssets.Characters.FirstPerson
                             continue;
                         }
 
+                        //Check 2: Is p + d within the bounds of the scene?
                         bool inBounds = agentManager.SceneBounds.Contains(newPosition);
 
-                        //Check 2: Is p + d within the bounds of the scene?
                         if (errorMessage == "" && !inBounds) {
                             errorMessage = "In " +
                                 UnityEngine.SceneManagement.SceneManager.GetActiveScene().name +
                                 ", position " + newPosition.ToString() +
                                 " can be reached via capsule cast but is beyond the scene bounds.";
+                            
+                            shouldEnqueue = false;
+                            continue;
                         }
 
                         //Check 3: Can the carried object fit into p + d's four cardinal orientations?
-                        shouldEnqueue = inBounds && (
+                        shouldEnqueue = (
                             handObjectCanFitInPosition(newPosition, 0.0f) ||
                             handObjectCanFitInPosition(newPosition, 90.0f) ||
                             handObjectCanFitInPosition(newPosition, 180.0f) ||
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -340,19 +340,34 @@ namespace UnityStandardAssets.Characters.FirstPerson
             HashSet<Vector3> seenPoints = new HashSet<Vector3>();
             int layerMask = 1 << 8;
             int stepsTaken = 0;
+
+            //Visualize SceneBounds
+            //GameObject boundsVis = GameObject.CreatePrimitive(PrimitiveType.Cube);
+            //boundsVis.transform.position = agentManager.SceneBounds.center;
+            //boundsVis.transform.localScale = 2 * agentManager.SceneBounds.extents;
+
+            //Traverse entire room in every direction with every step, stopping at any points that repeat and continuing on from ones that don't yet
             while (pointsQueue.Count != 0) {
                 stepsTaken += 1;
                 Vector3 p = pointsQueue.Dequeue();
+                //Check if p is a point already reached from a different direction
                 if (!goodPoints.Contains(p)) {
+                    //Add p to HashSet of reachable positions
                     goodPoints.Add(p);
+                    //Initialize hashset ("list") of objects that agent is currently colliding with (but don't epxect it to have anything yet)
                     HashSet<Collider> objectsAlreadyColliding = new HashSet<Collider>(objectsCollidingWithAgent());
+                    //From p, move in each cardinal direction by d (gridSize * gridMultiplier)
                     foreach (Vector3 d in directions) {
                         Vector3 newPosition = p + d * gridSize * gridMultiplier;
+                        //If p + d is a point already reached from a different direction, move on to the next direction from p
                         if (seenPoints.Contains(newPosition)) {
                             continue;
                         }
+
+                        //Since p + d is a new point, run the checks to see if it is reachable by the agent
                         seenPoints.Add(newPosition);
 
+                        //Check 1: Does agent hit any non-agent, non-floor collider between p and p + d (other than what it was already colliding with on Awake())?
                         RaycastHit[] hits = capsuleCastAllForAgent(
                             cc,
                             sw,
@@ -372,7 +387,15 @@ namespace UnityStandardAssets.Characters.FirstPerson
                                 break;
                             }
                         }
+
+                        //If CapsuleCast check failed, move on to next direction from p (otherwise Check 2's error message could return a false-positive)
+                        if(shouldEnqueue == false) {
+                            continue;
+                        }
+
                         bool inBounds = agentManager.SceneBounds.Contains(newPosition);
+
+                        //Check 2: Is p + d within the bounds of the scene?
                         if (errorMessage == "" && !inBounds) {
                             errorMessage = "In " +
                                 UnityEngine.SceneManagement.SceneManager.GetActiveScene().name +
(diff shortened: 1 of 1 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -372,23 +382,38 @@ protected void HideAllAgentRenderers()
                                 break;
                             }
                         }
+
+                        //If CapsuleCast check failed, move on to next direction from p (otherwise Check 2's error message could return a false-positive)
+                        if(shouldEnqueue == false) {
```
mattdeitke, line 387: Two things here:

1. Can you make this conditional check `!shouldEnqueue` instead of `== false`?
2. Can you add a space after `if`? (This is also a C# styling convention.)

So it goes from
```c#
if(shouldEnqueue == false) {
```
to
```c#
if (!shouldEnqueue) {
```

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -340,19 +340,29 @@ protected void HideAllAgentRenderers()
...
             while (pointsQueue.Count != 0) {
                 stepsTaken += 1;
                 Vector3 p = pointsQueue.Dequeue();
+                //Check if p is a point already reached from a different direction
                 if (!goodPoints.Contains(p)) {
+                    //Add p to HashSet of reachable positions
                     goodPoints.Add(p);
+                    //Initialize hashset ("list") of objects that agent is currently colliding with (but don't epxect it to have anything yet)
                     HashSet<Collider> objectsAlreadyColliding = new HashSet<Collider>(objectsCollidingWithAgent());
+                    //From p, move in each cardinal direction by d (gridSize * gridMultiplier)
```
mattdeitke, line 354: Another C# style convention, can you add a space above each comment, if it is not the start of a new indentation group? So this block goes from:
```c#
                Vector3 p = pointsQueue.Dequeue();
                //Check if p is a point already reached from a different direction
                if (!goodPoints.Contains(p)) {
                    //Add p to HashSet of reachable positions
                    goodPoints.Add(p);
                    //Initialize hashset ("list") of objects that agent is currently colliding with (but don't epxect it to have anything yet)
                    HashSet<Collider> objectsAlreadyColliding = new HashSet<Collider>(objectsCollidingWithAgent());
                    //From p, move in each cardinal direction by d (gridSize * gridMultiplier)
                    foreach (Vector3 d in directions) {
                        Vector3 newPosition = p + d * gridSize * gridMultiplier;
                        //If p + d is a point already reached from a different direction, move on to the next direction from p
                        if (seenPoints.Contains(newPosition)) {
                            continue;
                        }
```
to:
```c#
                Vector3 p = pointsQueue.Dequeue();

                // Check if p is a point already reached from a different direction
                if (!goodPoints.Contains(p)) {
                    // Add p to HashSet of reachable positions
                    goodPoints.Add(p);

                    // Initialize hashset ("list") of objects that agent is currently colliding with (but don't epxect it to have anything yet)
                    HashSet<Collider> objectsAlreadyColliding = new HashSet<Collider>(objectsCollidingWithAgent());

                    // From p, move in each cardinal direction by d (gridSize * gridMultiplier)
                    foreach (Vector3 d in directions) {
                        Vector3 newPosition = p + d * gridSize * gridMultiplier;

                        // If p + d is a point already reached from a different direction, move on to the next direction from p
                        if (seenPoints.Contains(newPosition)) {
                            continue;
                        }
```
Notice that `// Add p to HashSet of reachable positions` does not have a newline above it because it is starting a new indented further than the previousline.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -340,19 +340,29 @@ protected void HideAllAgentRenderers()
             HashSet<Vector3> seenPoints = new HashSet<Vector3>();
             int layerMask = 1 << 8;
             int stepsTaken = 0;
+
+            //Traverse entire room in every direction with every step, stopping at any points that repeat and continuing on from ones that don't yet
```
mattdeitke, line 344: For each of the comments, can you add a single space after the backslashes //?

So this goes from:
```c#
//Traverse entire room in every direction...
```
to:
```c#
// Traverse entire room in every direction...
```
This is a C# style convention and makes things look a bit less cluttered.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -340,19 +340,29 @@ protected void HideAllAgentRenderers()
...
                         Vector3 newPosition = p + d * gridSize * gridMultiplier;
+                        //If p + d is a point already reached from a different direction, move on to the next direction from p
                         if (seenPoints.Contains(newPosition)) {
                             continue;
                         }
+
+                        //Since p + d is a new point, run the checks to see if it is reachable by the agent
                         seenPoints.Add(newPosition);
 
+                        //Check 1: Does agent hit any non-agent, non-floor collider between p and p + d (other than what it was already colliding with on Awake())?
```
mattdeitke, line 365: Can you break some of these long comments into multiple lines, to make it a bit easier to read on all devices/editors? So it becomes something like:
```c#
// Check 1: Does agent hit any non-agent, non-floor collider between
// p and p + d (other than what it was already colliding with on Awake())?
```

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -691,7 +717,7 @@ public IEnumerator checkInitializeAgentLocationAction()
             float[] xs = new float[] { grid_x1, grid_x1 + gridSize };
             float[] zs = new float[] { grid_z1, grid_z1 + gridSize };
             List<Vector3> validMovements = new List<Vector3>();
-
+
```
mattdeitke, line 720: Would remove the whitespace here

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -372,23 +382,38 @@ protected void HideAllAgentRenderers()
...
+                        }
+
+                        //Check 2: Is p + d within the bounds of the scene?
                         bool inBounds = agentManager.SceneBounds.Contains(newPosition);
+
                         if (errorMessage == "" && !inBounds) {
                             errorMessage = "In " +
                                 UnityEngine.SceneManagement.SceneManager.GetActiveScene().name +
                                 ", position " + newPosition.ToString() +
                                 " can be reached via capsule cast but is beyond the scene bounds.";
```
mattdeitke, line 398: Why is an error message getting set on this check, but none of the other checks?

Moreover, I'm not convinced that this check is even working properly. If this check is ever true (i.e., `!inBounds`), the `errorMessage` will get set, and based on this line in the `GetReachablePositions` action, no positions will actually ever be returned!
https://github.com/allenai/ai2thor/blob/grid_size_debugging/unity/Assets/Scripts/BaseFPSAgentController.cs#L465

My question is then, are you sure `inBounds` is not always set to true, which might be the case if `agentManager.SceneBounds` contains an infinite space of values.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -372,23 +382,38 @@ protected void HideAllAgentRenderers()
...
                         }
 
-                        shouldEnqueue = shouldEnqueue && inBounds && (
+                        //Check 3: Can the carried object fit into p + d's four cardinal orientations?
+                        shouldEnqueue = (
                             handObjectCanFitInPosition(newPosition, 0.0f) ||
                             handObjectCanFitInPosition(newPosition, 90.0f) ||
                             handObjectCanFitInPosition(newPosition, 180.0f) ||
                             handObjectCanFitInPosition(newPosition, 270.0f)
                         );
```
mattdeitke, line 410: If `shouldEnqueue = false` is set here, then `shouldEnqueue = (...` is set in the next block, setting `shouldEnqueue = false` wouldn't have ended up doing anything, since its value is being immediately overwritten. Perhaps you're missing another

```c#
if (!shouldEnqueue) {
    continue;
}
```
statement.

mattdeitke (changes requested):
:+1:  It looks like it solves the issue that used to print something, upon initialization, like:
```
In FloorPlan10_physics, position (6.0, 0.9, -1.3) can be reached via capsule cast but is beyond the scene bounds
```
and solves the `GetReachablePositions` returning `None` issue when the grid size is large. Here are the outputs before and after:
https://colab.research.google.com/drive/18UZLAO5BOi-PrKAAaK1RFuu4qHeUmK5L?usp=sharing

Left a few questions and a few things C# style comments.

unity/Assets/Scripts/BaseFPSAgentController.cs
elimvb: Done

unity/Assets/Scripts/BaseFPSAgentController.cs
elimvb: Done

unity/Assets/Scripts/BaseFPSAgentController.cs
elimvb: Done

unity/Assets/Scripts/BaseFPSAgentController.cs
elimvb: More to this than I realized. Done.

unity/Assets/Scripts/BaseFPSAgentController.cs
elimvb: No, that's not necessary either, since `shouldEnqueue` is redeclared and reinitialized after every loop iteration anyways. I kept it for clarity, but I suppose I ought to remove it.

unity/Assets/Scripts/BaseFPSAgentController.cs
elimvb: Done.

unity/Assets/Scripts/BaseFPSAgentController.cs
elimvb: I think that failed return was the point when Luca added this error-check to the logic: to raise the issue.

As for `inBounds` state, that is definitely working as intended. `SceneBounds` has a helper function committed to checking that it isn't infinite, and encompasses the scene so that when p + d reaches outside of it, it's immediately disqualified from being reachable. That being said, this check should hypothetically never need to be used as long as all of our rooms are watertight with collision, which they are now.

Lucaweihs:
Ah, [unfortunately I have a PR that extends the `GetReachablePositions` function](https://github.com/allenai/ai2thor/pull/669/files#diff-9f8d6622ca2af04d8f42b7574a97147d5ddfb07e19077bda19deb59da36cd970R353-R360). Can you maybe add these fixes to that PR? I think my PR might fix some of the issues you've solved as well.

PR #672, opened 2021-03-23
cached axisAlignedBox fix

winthos:
axis aligned bounding boxes were not being updated correctly due to being cached.

This is a quick fix to ensure that axis aligned boxes are correctly updated with each emit.


Lucaweihs:
Questions / comments:

1. What is the FPS impact of this change?
2. There is a typo in `axisAlignedBoundigBox`, it should be `axisAlignedBoundingBox` (this isn't an error that you introduced but good to fix anyway).
3. Don't we also have to move `cachedObjectOrientedBoundingBox` above the position/rotation if statement?

winthos:
Typo is fixed.

Also, I originally thought the object oriented box could be safely cached, but I found some cases (book open vs closed) where the object position doesn't change but would still require the box to be updated, so all of this caching logic has been removed for now.

As for FPS impact, I'll need to recruit someone to help me benchmark the fps difference. I don't expect it to be too much.

AlvaroHG:
It's seems the action-per-second rate is 12% slower in average across all rooms with the move, look, rotate actions. Ran it with the benchmark task in a linux server, here are the output files with more detail.

[benchmark_linux_master.txt (c18d6c062b27a0c097b4ac180e70071003d8a088)](https://github.com/allenai/ai2thor/files/6208752/benchmark_linux_master.txt) runs at 90 actions-per-second
[benchmark_linux_pr.txt (70b31d7fdf4621e17928c6fb97e621ca4bae611e)](https://github.com/allenai/ai2thor/files/6208754/benchmark_linux_pr.txt) runs at 80 actions-per-second

ekolve:
Would it be possible to generate a cache key for this based off the properties in the SimObj? Such as the state of the object (IsOpen, IsBroken, etc.)? We should be able to incorporate all that into a key and if the key changes we regenerate the boxes.

winthos:
I think a cache key like that could work. Do you think that should be a future PR or should we delay this one until I make a cache key for it?

PR #673, opened 2021-03-25
Bounding box sub object update

elimvb:
Quick update to BoundingBox_Update, so getAxisAlignedBoundingBox pulls from SimObjPhysics collider array instead of grabbing all child colliders beneath transform, to avoid grabbing "receptacled" colliders as well.

Also, the coffee machines had no collision arrays, so I added them.


winthos (approved):
looks good

PR #682, opened 2021-03-31
Unity test framework

ekolve:
This adds support for Unity Test Framework to our project.  The majority of the changes are related to assembly definition files and moving code underneath the Scripts directory.


tasks.py
```diff
@@ -3282,3 +3283,84 @@ def generate_pypi_index(context):
...
 
+@task
+def ci_test_utf(context):
+    s3 = boto3.resource("s3")
+    results_path, results_logfile = test_utf(context)
+    
+    for l in [results_path, results_logfile]:
+        key = "builds/" + os.path.basename(l)
+        with open(l) as f:
+            s3.Object('ai2-prior-erick-test', key).put(
```
AlvaroHG, line 3294: what is this path? is some other system reading this specific location?

tasks.py
```diff
@@ -3282,3 +3283,84 @@ def generate_pypi_index(context):
...
+import json
+import os
+
+def test_testresults_exist():
+    test_results_path = "{test_results_path}" 
+    assert os.path.isfile("{test_results_path}"), "TestResults at: {test_results_path}  do not exist"
+
+"""
+    )
+    for class_name, test_cases in class_tests.items():
```
AlvaroHG, line 3339: awsome!

unity/Assets/Priority Queue/PriorityQueue.asmdef
```diff
@@ -0,0 +1,3 @@
+﻿{
+	"name": "PriorityQueue"
```
AlvaroHG, line 2: why were these assembly definitions generated?

AlvaroHG:
LGTM, just a couple of questions

tasks.py
```diff
@@ -3282,3 +3283,84 @@ def generate_pypi_index(context):
...
+
+@task
+def test_utf(context):
+    project_path = os.path.join(os.getcwd(), 'unity')
+    commit_id = git_commit_id()
+    test_results_path = os.path.join(project_path, 'utf_testResults-%s.xml' % commit_id)
+    logfile_path = os.path.join(os.getcwd(), 'thor-testResults-%s.log' % commit_id)
+
+    command = (
+        "%s -runTests -testResults %s -logFile %s -testPlatform PlayMode -projectpath %s "
```
AlvaroHG, line 3307: we should probably pass width and height here, for consistent visibility tests, like we do when normally running THOR from python

ekolve: We aren't able to pass width/height on the command line or change it using `Screen.SetResolution`, but I did update the ProjectSettings to set the default width/height to 300/300.

PR #684, opened 2021-04-03
Add Heuristic to PutObject close to (x,y) RaycastHit on Target Receptacle

winthos:
This is a rework of the feature requested in #627 however includes accommodation for edits made in #648 which changed a lot in the `screenToWorldTarget` function.

Passing in `putNearXY=True` to the `PutObject` action will now use the distance from the RaycastHit as the heuristic for choosing an object spawn point on the receptacle.


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4476,35 +4476,55 @@ protected void SetStateOfObjectsThatDontHaveAnimationTime(SetObjectStates SetObj
...
+                forceAction: forceAction,
+                placeStationary: placeStationary,
+                randomSeed: randomSeed);
         }
 
         //if you are holding an object, place it on a valid Receptacle 
         //used for placing objects on receptacles without enclosed restrictions (drawers, cabinets, etc)
         //only checks if the object can be placed on top of the target receptacle
-        public void PlaceHeldObject(float x, float y, bool forceAction=false, bool placeStationary=true, int randomSeed = 0, float z = 0.0f){
+        public void PlaceHeldObject(float x, float y, bool forceAction=false, bool placeStationary=true, int randomSeed = 0, float maxDistance = 0.0f, bool putNearXY = false){
```
Lucaweihs, line 4500: As changing `z` to `maxDistance` is an backwards incompatible API change we should only be making this at the next larger release (e.g. 2.8).

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4521,10 +4541,39 @@ public void PlaceHeldObject(string objectId, bool forceAction=false, bool placeS
...
+                randomSeed: randomSeed, 
+                maxDistance: maxDistance);
         }
 
-        private void placeHeldObject(SimObjPhysics targetReceptacle, bool forceAction, bool placeStationary, int randomSeed, float z) {
+        private void placeHeldObject(            
+            SimObjPhysics targetReceptacle, 
+            bool forceAction, bool placeStationary, 
+            int randomSeed, 
+            float maxDistance) {
```
Lucaweihs, line 4556: Save comment as above for `maxDistance`.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4521,10 +4541,39 @@ public void PlaceHeldObject(string objectId, bool forceAction=false, bool placeS
...
+                placeStationary: placeStationary, 
+                randomSeed: randomSeed, 
+                maxDistance: maxDistance,
+                putNearXY: false,
+                hit: hit);
+            }
+
+        private void placeHeldObject(
+            SimObjPhysics targetReceptacle, 
+            bool forceAction, bool placeStationary,
```
Lucaweihs, line 4572: Add new line after `bool forceAction,`

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4638,20 +4687,35 @@ public void PlaceHeldObject(string objectId, bool forceAction=false, bool placeS
             InstantiatePrefabTest script = physicsSceneManager.GetComponent<InstantiatePrefabTest>();
             //set degreeIncrement to 90 for placing held objects to check for vertical angles
             List<ReceptacleSpawnPoint> spawnPoints = targetReceptacle.ReturnMySpawnPoints(onlyPointsCloseToAgent);
-            if (randomSeed != 0) {
-                List<ReceptacleSpawnPoint> randomizedSpawnPoints = new List<ReceptacleSpawnPoint>();
-                float maxDistance = z;
-                if (maxDistance == 0.0f) {
+            if (randomSeed != 0 || putNearXY) {
+                List<KeyValuePair<ReceptacleSpawnPoint, float>> distSpawnPoints = new List<KeyValuePair<ReceptacleSpawnPoint, float>>();
+                if (Mathf.Approximately(maxDistance, 0.0f)) {
```
Lucaweihs, line 4692: Using `Mathf.Approximately` doesn't seem correct here, we're trying to check if the user passed an _exact_ 0 value. That said, I would prefer this was refactored so that `maxDistance` was of type `float?` (so we set it equal to `maxVisibleDistance` if it's null) and we threw and error if anyone passed a value of `maxDistance==0`.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4638,20 +4687,35 @@ public void PlaceHeldObject(string objectId, bool forceAction=false, bool placeS
...
                     maxDistance = maxVisibleDistance;
                 }
                 foreach (ReceptacleSpawnPoint sp in spawnPoints) {
                     Vector3 tmp = new Vector3(transform.position.x, sp.Point.y, transform.position.z);
                     if (Vector3.Distance(sp.Point, tmp) < maxDistance) {
-                        randomizedSpawnPoints.Add(sp);
+                        float dist = 0; 
+                        if (putNearXY) {
+                            dist = -Vector3.Distance(sp.Point, hit.point);  // Will be used to sort from greatest negative distance (closest) to least (farthest) from Raycast Hit
```
Lucaweihs, line 4700: This seems backwards to me? Sorting, by default, goes from smallest to largest, this will make it go from largest to smallest.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4638,20 +4687,35 @@ public void PlaceHeldObject(string objectId, bool forceAction=false, bool placeS
...
+                        }
+                        distSpawnPoints.Add(new KeyValuePair<ReceptacleSpawnPoint, float>(sp, dist));
                     }
                 }
-                randomizedSpawnPoints.Shuffle_(randomSeed);
-                spawnPoints = randomizedSpawnPoints;
+
+                //actually sort by distance closest to raycast hit if needed here, otherwise leave random
+                if(putNearXY) {
+                    distSpawnPoints.Sort((x, y) => (y.Value.CompareTo(x.Value)));
```
Lucaweihs, line 4708: Ah, here's why the above didn't cause an error: you've effectively reversed things twice. If you remove the negative above and change this line to
```
distSpawnPoints.Sort((x, y) => (x.Value.CompareTo(y.Value)));
```
then things should work.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -4638,20 +4687,35 @@ public void PlaceHeldObject(string objectId, bool forceAction=false, bool placeS
...
+                }
+
+                if(!putNearXY) {
```
Lucaweihs, line 4711: This seems like it could be an `else {` ?

Lucaweihs (changes requested):
Minor comments here and there. My main worry is that this introduces a non-backwards-compatible API change by changing `z` to `maxDistance`. We should revert this until we're ready to make a 2.8 release.

ekolve:
Could you add a UnitTest for this that tests with/without passing in the `putNearXY` boolean and asserting the location that object was placed?

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: This shouldn't be an issue for backwards compatibility. The `maxDistance`/`z` value is only internally used by the `PlaceHeldObject` function. `PlaceHeldObject` is onlyu called by `PutObject` which never exposed the  `maxDistance`/`z` to the python controller.

The original versions of `PutObject` were:

```       
        //Target what receptacle to put a held object via (x, y) screen coordinates
        public void PutObject(float x, float y, bool forceAction=false, bool placeStationary=true){
            PlaceHeldObject(x, y, forceAction, placeStationary);
        }

        //Target what receptacle to put a held object via objectId directly
        public void PutObject(string objectId, bool forceAction=false, bool placeStationary=true){
            PlaceHeldObject(objectId, forceAction, placeStationary);
        }
 ```
 
The new versions included in this update still don't expose or change the default values of `maxDistance`/`z` , but they do add new optional parameters of `randomSeed` and `putNearXY`

```        
        public void PutObject(float x, float y, bool forceAction=false, bool placeStationary=true, int 
        randomSeed=0, bool putNearXY=false){
            PlaceHeldObject(
                x: x, 
                y: y,
                forceAction: forceAction, 
                placeStationary: placeStationary,
                randomSeed: randomSeed,
                putNearXY: putNearXY);
        }

        public void PutObject(string objectId, bool forceAction=false, bool placeStationary=true, 
        int randomSeed=0){
            PlaceHeldObject(                
                objectId: objectId,
                forceAction: forceAction,
                placeStationary: placeStationary,
                randomSeed: randomSeed);
        }
```

File: unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs (hunk 3 above)
Comment (winthos) [reply unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs]: done

File: unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs (hunk 4 above)
Comment (winthos) [reply unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs]: this has now been replaced so `maxDistance` uses a nullable type.

File: unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs (hunk 6 above)
Comment (winthos) [reply unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs]: this has now been changed as recommended so its no longer going backwards and then undoing it for no reason haha

File: unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs (hunk 7 above)
Comment (winthos) [reply unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs]: this has been switched to an else statement

File: unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs (hunk 5 above)
Comment (winthos) [reply unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs]: this has been resolved in comment below.

**Comment by winthos** [comment]
unit test asserting the correct positions when using/not using `PutNearXY` has been added

File: unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs (hunk 1 above)
Comment (Lucaweihs) [reply unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs]: I'm not sure I understand, `PlaceHeldObject` is exposed to the Python API and so anyone who used this action with the `z` argument would now find that this doesn't work. Or am I wrong?

File: unity/Assets/UnitTests/TestPutObject.cs [hunk 8]
```diff
@@ -0,0 +1,62 @@
...
+    public class TestPutObject : TestBase
+    {
+        [UnityTest]
+        public IEnumerator TestPutObject_PutNearXY_True()
+        {
+            yield return ExecuteDebugAction("init");
+            yield return ExecuteDebugAction("rr");
+            yield return ExecuteDebugAction("rr");
+            yield return ExecuteDebugAction("ld");
+            yield return ExecuteDebugAction("mr");
```
Lucaweihs, line 20: In our group chat about these C# tests we'd talked about not using the debug controller as it effectively creates a "meta" API and will make people nervous to change `DebugInputField.cs`. Is there a way to use the actions by passing a json-formatted string of the python-side act more directly (e.g. similarly as is done when calling `run simple.json` in the input field)? Pinging @ekolve / @mattdeitke in case they know.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: Yes, it is exposed but the documented one that should be used directly from the Python API is [PutObject](https://ai2thor.allenai.org/ithor/documentation/interactive-physics/#sub-put-object)

`PutObject` has not changed and never took `z`. Although I'm not sure if there are older projects that instead use `PlaceHeldObject` directly, which may be the concern here.

Lucaweihs: Yes that's my concern. Generally I'm ok with such changes so long as they throw an error, can you create a python-side test that checks that taking a `PlaceHeldObject` breaks when called with `z` but succeeds when called with `maxDistance` (all other parameters the same)?

winthos: yes I'll try this. I think because of the new action dispatcher, passing in a param that is not used will throw some sort of exception? I'll update here once I try it out.

ekolve:
wrt to @Lucaweihs comment about not using the DebugInputField actions,  the following function `ExecuteAction` can be added to TestBase.  

```python
public class TestBase {
        public IEnumerator ExecuteAction(Dictionary<string, object> action) {
            var debugInputField = GameObject.FindObjectOfType<DebugInputField>();
            while(debugInputField.PhysicsController.IsProcessing) {
            }
            debugInputField.PhysicsController.ProcessControlCommand(action);
        }
}
```

Lucaweihs (approved):
LGTM (assuming that the final build passes all the tests of course).

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: ok! so after some back and forth, if someone happens to be using `PlaceHeldObject` and passes in `z`, a missing argument exception now gets thrown:

```Action failed with error message 'action: PlaceHeldObject is missing the following arguments: maxDistance```

There is still a potential issue where if one were to pass in both `maxDistance` and `z` it would succeed with potentially silent unintended behavior, but we need to have a larger conversation about accounting for this later. As both Eric and Alvaro brought up, we currently don't want to have a forced exception check for extraneous parameters.

To facilitate this, I've made `maxDistance` a required parameter on `PlaceHeldObject` and the defaulting of `maxDistance` to equal `maxVisibleDistance` set on initialize is now done within `PutObject`. So `PutObject` behaves exactly the same as before even though `maxDistance` is no longer a nullable type.

Additionally added in a bunch of unit tests to help check for all combinations of this.

PR #687, opened 2021-04-05
Add third-party plugin entry

ekolve:


mattdeitke:
Do you think this should be in the top-level `/README.md`, or do you think it should be in a sub-directory's README, like `/unity/README.md`? I'd maybe prefer the later because this isn't necessarily information that anybody using the Python API alone will ever need.

ekolve:
I'd be fine with either (top-level or unity/).  The only concern I would have is that someone would not think to look for a README beneath that folder and we already have a block `Unity Development`, so I don't think its adding too much noise to the page.

README.md
```diff
@@ -46,6 +46,12 @@ controller = ai2thor.controller.Controller(
...
...der and right-click in the window that displays the contents of the folder and selecte Create -> Assembly Definition.  This will create a new file in the folder - rename the Assembly definition to match the name of the plugin you are adding (e.g. 'Priority Queue' or 'iTween').  This allows the Assembly definition to be easily found during creation of the reference.  To reference the plugin, a reference must be created in the Assembly Definition file located at `unity/Assets/Scripts/AI2ThorBase`.  Locate the `AI2ThorBase` Assembly Definition file under `unity/Assets/Scripts` within the Project tab.  Click on the file and locate the section titled "Assembly Definition References".  Click the `+` sign to add a new entry, then click on the circle to the right of the newly created entry.  That will bring up a menu with all the Assembly definitions in the project.  Select the name of the new plugin/Assembly Definition that was just created.  Scroll to the bottom and click the "Apply" button.  The plugin should now be available to use.  The AI2ThorBase Assembly definition and the new plugin folder will need to be commited. Additional information about Assembly definitions can be found at:
```
mattdeitke, line 51: Looks good. One small thing, I thought we were going away from using AI2Thor (in favor of AI2-THOR)? Perhaps just change `unity/Assets/Scripts/AI2ThorBase` to `unity/Assets/Scripts/Assembly`.

ekolve: Thats a good point - i renamed the assembly on a different branch and will merge that in to main and updated the docs.

mattdeitke (approved):

Issue #651, opened 2021-03-11
Action to select metadata keys that are required


Tushar-N:
A lot of metadata is sent over at each controller step, but not all this metadata is required.

1.  Is it possible to set the scope of metadata returned by the server on the fly (as a controller action).
2. Will this make things faster because less data needs to be sent over, or is this not a speed bottleneck (compared to sending over frame/segmentation tensors etc)?

I could have sworn I saw something about this in an issue/pull request, but I can't find it now. Sorry if this is a duplicate.

------------

For example, if I don't care about the agent, or scene bounds, or specific properties of objects.
```
controller.step(
  "UpdateRequiredMetadata",
  meta_keys={
    'agent': False, 
    'sceneBounds': False,
    'objects':{'axisAlignedBoundingBox': False, 'mass': False}
  }
)
```

or if I ONLY care about specific object types? 

```
controller.step(
  "UpdateRequiredMetadata",
  obj_types={
    'Painting': False, 
    'DeskLamp': False,
  }
)
```

but later down the line, I change my mind and I do want info about one of them
```
controller.step(
  "UpdateRequiredMetadata",
  obj_types={
    'DeskLamp': True,
  }
)
```

ekolve:
Yes, there is an action that you can run:

Pass a list of objectIds to the following action that you are interested in.  This will greatly reduce the amount of metadata generated.
```python
controller.step('SetObjectFilter', objectIds=[...])
```


If you want to remove the filter, call the following action.
```python
controller.step('ResetObjectFilter')
```

Tushar-N:
Perfect, looks like I missed this function. Thanks!

Tushar-N:
Reopening this just to say that it would be nice to have some extra functionality for this class of actions. For example:
- setting object filters based on objectType, rather than Id. Right now if I limit the scope to Knife and Lettuce, but then I slice the lettuce, I lose the ability to have the new sliced objects in the metadata.
- More fine-grained control over metadata keys (like I described in the original post)

Tushar-N:
Also wanted to ask: If I limit objects in metadata, this also limits what objects I can interact with. For example this does not work, even though `query.metadata["actionReturn"]` contains the correct objectId.
```
# while standing right in front of a fridge
controller.step('SetObjectFilter', objectIds=['Lettuce|+00.22|+01.19|00.00'])
query = controller.step(action='GetObjectInFrame', x=0.5, y=0.5)
if query:
[...]

PR #693, opened 2021-04-08
refactor SimObjPhysics.isInteractable into BaseFPSAgent.IsInteractabl…

ekolve:
The primary purpose of the PR is to address issue #651, which has to do with when a user sets an ObjectFilter, but later attempts to call say `OpenObject` on an object not listed in the filter.  This will currently fail as the visibility check for the object won't succeed since the object gets filtered out.  This PR limits the filter created with SetObjectFilter to just the metadata.  As well, I refactored the `VisibleSimObjs` methods and `getTargetObject` method to specify the targetObjectId when performing the visibility check.  This will dramatically speed up these calls as the number of objects that are checked are just limited to the targetObject.  

On the SimObjPhysics class there is a field `isInteractable`.  This gets updated during visibility checks, but gets used during actions such as OpenObject/CloseObject.  I have done away with this field in favor of a method on the Agent `IsInteractable`.  Storing whether an object is interactable on the object itself can lead to errors due to the fact that it is essentially a cached value from the last time the visibility check was run as well in multi-agent configurations this value may end up representing a different agent's visibility.  I renamed the field to `debugIsInteractable` as well as the field `isVisible` to `debugIsVisible` as both of these were used for Editor based debugging.

The major side-effect of this change will be that the filter won't be applied as strictly as it was before.  For example, the action  `OpenAllVisibleObjects` will now actually open all the visible objects whereas with a filter it would only open the visible objects also present in the filter.   There aren't many actions like this and I don't think this change will lead to more confusion (likely less since actions now behave as people would expect) and we should get a small performance boost in the places we call getTargetObject/VisibleSimObjs (with an objectId).


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,30 +2603,39 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
         }
 
         public SimObjPhysics[] VisibleSimObjs(string objectId, bool forceVisible = false) {
-            ServerAction action = new ServerAction();
-            action.objectId = objectId;
-            action.forceVisible = forceVisible;
-            return VisibleSimObjs(action);
+            SimObjPhysics sop = getSimObjectFromTypeOrId(objectType: null, objectId: objectId);
```
Lucaweihs, line 2606: This function's behavior is super strange to me given its name and inputs. Looking at how it's used in the code base I would vote for removing this completely and replacing it with something like:
```
public SimObjPhysics GetSimObjFromId(string objectId, bool excludeIfNotVisible)
```

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,30 +2603,39 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
...
+                return simObjs.ToArray();
+            } else {
+                return GetAllVisibleSimObjPhysics(m_Camera, maxVisibleDistance, simObjs);
+            }
         }
 
 
         public SimObjPhysics[] VisibleSimObjs(ServerAction action) {
             List<SimObjPhysics> simObjs = new List<SimObjPhysics>();
+            SimObjPhysics[] visibleSimObjs = null;
```
Lucaweihs, line 2621: As we're refactoring things anyway can we make it so that `VisibleSimObjs(ServerAction action)` does not take a `ServerAction` and instead explicitly declares its parameters?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2597,30 +2603,39 @@ protected void updateAllAgentCollidersForVisibilityCheck(bool enableColliders)
...
+            if (!string.IsNullOrEmpty(action.objectId))
+            {
+                // if an objectId was passed in we can limit the check to just the target object
+                // in theory we could return from here, but its possible that a conflicting 
+                // objectType is passed in which would then filter out a found object
```
Lucaweihs, line 2626: I would vote for removing this behavior. People should instead call the hypothetical `GetSimObjFromId(objectId, excludeIfNotVisible)` function if they want this.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2641,12 +2656,13 @@ public SimObjPhysics[] VisibleSimObjs(bool forceVisible)
                 return GetAllVisibleSimObjPhysics(m_Camera, maxVisibleDistance);
             }
         }
-        protected SimObjPhysics[] GetAllVisibleSimObjPhysics(Camera agentCamera, float maxDistance)
+        protected SimObjPhysics[] GetAllVisibleSimObjPhysics(Camera agentCamera, float maxDistance, IEnumerable<SimObjPhysics> simObjs = null)
```
Lucaweihs, line 2659: Can we rename `simObjs` so that it's more informative for including these means?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2655,17 +2671,13 @@ protected SimObjPhysics[] GetAllVisibleSimObjPhysics(Camera agentCamera, float m
         // is within range of the maxVisibleDistance, but obscurred only within this
         // range and is visibile outside of the range, it will get reported as invisible
         // by the new scheme, but visible in the current scheme.
-        private SimObjPhysics[] GetAllVisibleSimObjPhysicsDistance(Camera agentCamera, float maxDistance)
+        private SimObjPhysics[] GetAllVisibleSimObjPhysicsDistance(Camera agentCamera, float maxDistance, IEnumerable<SimObjPhysics> simObjs)
```
Lucaweihs, line 2674: Same comment as above.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2677,21 +2689,21 @@ private SimObjPhysics[] GetAllVisibleSimObjPhysicsDistance(Camera agentCamera, f
             return visible.ToArray();
         }
 
-        private SimObjPhysics[] GetAllVisibleSimObjPhysicsCollider(Camera agentCamera, float maxDistance)
+        private SimObjPhysics[] GetAllVisibleSimObjPhysicsCollider(Camera agentCamera, float maxDistance, IEnumerable<SimObjPhysics> simObjs)
```
Lucaweihs, line 2692: Same comment as above.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -2111,8 +2111,8 @@ public void DirectionalPush(ServerAction action)
                 return;
             }
 
-            if (!action.forceAction && target.isInteractable == false) {
-                print(target.isInteractable);
+            if (!action.forceAction && !IsInteractable(target)) {
+                print(IsInteractable(target));
```
Lucaweihs, line 2115: Should this print statement be gated by `# if UNITY_EDITOR`?

Lucaweihs (changes requested):
Very happy to see these changes! I have left some comments, mainly I think the Unity-side API around visibility should be refactored as it is getting quite wonky with methods having a lot of hidden functionality.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: done, actually created two methods.  One that is just a method to get a SimObj by objectId and another that also does the interactable check with forceVisible boolean.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: Done, i removed this method altogether and refactored all uses to `getInteractableSimObjectFromId`

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -3285,7 +3243,36 @@ bool includeInvisible
...
+            if (!physicsSceneManager.ObjectIdToSimObjPhysics.ContainsKey(objectId)) {
+                errorMessage = "Cannot find sim object with id '" + objectId + "'";
+                return null;
+            }
+           
+            SimObjPhysics sop = physicsSceneManager.ObjectIdToSimObjPhysics[objectId];
+            if (sop == null) {
+                errorMessage = "Object with id '" + objectId+ "' is null";
+                return null;
+            }
```
Lucaweihs, line 3272: Is this possible? Somehow I feel like this should be a critical failure of some sort as it suggests the `ObjectIdToSimObjPhysics` dictionary is broken somehow.

Lucaweihs (approved):
Added small question but it's not a blocker so I've went ahead and approved.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: Good catch. Removed.

ekolve: Good catch. Removed.

winthos (approved):
lgtm

PR #696, opened 2021-04-12
Can_Open script and openable asset updates for dynamic bounding boxes

elimvb:
Pretty self-explanatory, I removed all references to state-change-based bounding boxes, since they're not needed anymore with dynamically-updated BBs.


winthos (approved):
looks good to go

PR #697, opened 2021-04-13
ManipulaTHOR UX Ehancements

mattdeitke:
NOTE: No functionality has changed here, just reorganizing some things and making breaking naming syntactic changes.

#### Action Names

1. `MoveMidLevelArm` -> `MoveArm`
2. `MoveMidLevelArmHeight` -> `MoveArmBase`
3. `PickUpMidLevelHand` -> `PickupObject`
4. `DropMidLevelHand` -> `ReleaseObject`
5. `SetMidLevelHandRadius` -> `SetHandSphereRadius` (matches what's reported in the metadata)
6. `MoveContinuous` -> `MoveAgent`
7. `RotateContinuous` -> `RotateAgent`

#### Metadata

1. `["arm"]["PickupableObjectsInsideHandsphere"]` -> `["arm"]["PickupableObjects"]`

#### Parameter Names

1. `action="PickupObject"`. `objectIds` -> `objectIdCandidates` -- since the objects only get picked up iff they are within the hand's sphere (i.e., there is no guarantee).
2. `action="Move"`. `direction: dict(x, y, z)` has been broken down into 2 parameters, `ahead: float` (amount in meters to move ahead) and `right: float` (amount in meters to move right). Note hat `ahead` and `right` can be negative, which supports moving  `back` and `left`, respectively.

## Developer Notes

All the arm actions are now in `ArmAgentController.cs`. This is necessary because `PickupObject` behaves differently between iTHOR and ManipulaTHOR. It also allows for actions like `MoveAhead` and `RotateRight` to fail in ManipulaTHOR, and actions like `Move` to fail in iTHOR. I have also matched all the static fields on the sidebar with what is in PhysicsRemoteFPSAgentController, which ArmAgentController inherits from.

 

## Tests

Tests have been added to make sure that `ArmAgentController` works as expected, but they have not been added for each arm action, as that functionality has not changed.

## Future Notes

I have made many actions that are iTHOR default agent specific, but don't work in ManipulaTHOR, unsupported. The full list can be found [here](https://github.com/allenai/ai2thor/blob/manipulathor-ux/unity/Assets/Scripts/ArmAgentController.cs#L135). Notable entries: `Move{Ahead, ...}`, `Rotate{Right, ...}`, `PickupObject(x, y)`, `PickupObject(objectId)`.

In the future, `Move{Ahead, ...}` and `Rotate{Right, ...}` should be supported, and simply shorthands for calling `Rotate`. However, this would require a bit of a refactor, where all `Move{Ahead, ...}` and `Rotate{Right, ...}` actions are not written using `ServerAction`.


mattdeitke:
FYI: This is ready for review and decently well tested 😄

ehsanik (commented):
Can we set the default values for the following variables to match what we use for the arm?

'disableRendering': True,
    'restrictMovement': False,
    'waitForFixedUpdate': False,
    'returnToStart': True,
    'speed': 1,

unity/Assets/Scripts/ContinuousMovement.cs
```diff
@@ -6,215 +6,207 @@
...
-                (t, direction) => t.localPosition + direction * unitsPerSecond * fixedDeltaTime
             );
         }
-        else {
 
-            return moveClosure(
-                (t) => t.position,
-                (t, pos) => t.position = pos,
-                (t, direction) => t.position + direction * unitsPerSecond * fixedDeltaTime
+        public static IEnumerator move(
```
ehsanik, line 53: can we rename move to moveAgent and rotate to rotateAgent to distinguish them from moving and rotating the arm?

unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -0,0 +1,285 @@
...
+                );
+            } else {
+                StartCoroutine(rotate);
+            }
+        }
+
+        // currently not finished action. New logic needs to account for the
+        // hierarchy of rigidbodies of each arm joint and how to detect collision
+        // between a given arm joint an other arm joints.
+        public void RotateMidLevelHand(ServerAction action) {
```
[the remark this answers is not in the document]
mattdeitke: I talked with @winthos about this one. He basically said let's wait until it's complete to add it.

unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -0,0 +1,285 @@
...
+        public void ReleaseObject() {
+            IK_Robot_Arm_Controller arm = getArm();
+            arm.DropObject();
+
+            // TODO: only return after object(s) dropped have finished moving
+            // currently this will return the frame the object is released
+            actionFinished(true);
+        }
+
+        public void WhatObjectsCanHandPickUp() {
```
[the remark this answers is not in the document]
mattdeitke: It's now removed in https://github.com/allenai/ai2thor/pull/697/commits/d725d288924be1b2478401bda91fb8c677d71efc.

unity/Assets/Scripts/ArmAgentController.cs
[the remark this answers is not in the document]
winthos: Yeah, currently this “works” but does not propagate collision checks within the arm joints themselves, so you would be able to rotate held objects inside the arm geometry which could cause issues.

unity/Assets/Scripts/ContinuousMovement.cs
winthos: These are all internal logic functions so we don’t necessarily need a name change for these. I’ll leave the final decision up to @mattdeitke

winthos (approved):
I think it’s about ready to go. Much cleaner than before.

unity/Assets/Scripts/ContinuousMovement.cs
mattdeitke: Fixed (in the public API via. ArmAgentController.cs) in https://github.com/allenai/ai2thor/pull/697/commits/cb0241ccae8330002b38dc2ec0b40d234ce75c98.

mattdeitke:
Default variables are now set based on https://github.com/allenai/ai2thor/pull/697#pullrequestreview-634810327.

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -91,21 +102,8 @@ void Start()
     //     StopMotionOnContact = b;
     // }
 
-    //debug for gizmo draw
-#if UNITY_EDITOR
```
ekolve, line 95: Did you mean to remove this?

mattdeitke: It's not removed. Just moved to the bottom of the file inside of the other
```
#if UNITY_EDITOR
```
block.

https://github.com/allenai/ai2thor/blob/manipulathor-ux/unity/Assets/Scripts/IK_Robot_Arm_Controller.cs#L617

ekolve (approved):

ehsanik (approved):

PR #698, opened 2021-04-13
RoboTHOR Procedural Wall Placement

mattdeitke:
I want to merge this before it's 1000x harder to do, if autoformatting gets applied.

All the functionality is ready and works, though I'm keeping it undocumented for now.

What this does is procedurally generate walls, ceilings, and floors in a maze like manner within RoboTHOR.


winthos (approved):
Lgtm

AlvaroHG:
This looks like functionality we may need down the line, not sure we should merge to main. We should also sync to unify what I have for iTHOR and testing, to unify how things are generated. We will reuse this just refactor it in the ProceduralGeneration library. For now I'd say keep in it's branch

mattdeitke:
Would heavily prefer against holding off on merging this indefinitely. From experience, if something is kept as an unmaintained branch, it will never be merged because merging in the future will be 1000x harder due to merge conflicts.

Moreover, this does not affect any existing functionality. It only affects users if they try to use `scene="robothor_generative"` .

PR #701, opened 2021-04-15
reformatting python code with black

ekolve:
I have reformatted all the Python code with black.  Please have a look to see if anything looks incorrect.


mattdeitke (approved):

tasks.py
```diff
@@ -545,7 +549,7 @@ def let_through(self, node):
             return self.construct_mapping(node)
 
     YamlUnity3dTag.add_constructor(
-        u"tag:unity3d.com,2011:47", YamlUnity3dTag.let_through
+        "tag:unity3d.com,2011:47", YamlUnity3dTag.let_through
```
Lucaweihs, line 552: I'm not sure what the `u` prefix does here, I assume removing this is fine?

Lucaweihs (commented):
I went through things quite quickly but looks pretty much fine. I left one small comment. Another thought mostly unrelated to this PR: there seems to be some inconsistency in how we perform string interpolation (sometimes with `%s`, sometimes with `.format`, sometimes with f-strings), eventually it would be great to use f-strings everywhere.

mattdeitke:
Worth noting: f-strings would break all backwards + sideways compatibility before Python 3.6. So we would have to remove support for Python 2.7 and Python 3.5.

It appears f-strings largely only appear in the tests at the moment, which make ai2thor still use-able in Python 2.7 and 3.5.

tasks.py
ekolve: That is a relic from Python 2.7 support.  In Python2.7 prefixing a string with `u` creates a unicode string.  This isn't necessary with Python 3.

ekolve:
Regarding f-strings, @mattdeitke is correct.  We would break compatibility with < Python 3.6, which may or may not affect a lot of people.  Pytorch continues to support Python 3.5.  Python 3.5 is the default Python 3 version for Ubuntu 16.04, though this can be upgraded it may deter some users.  If f-strings or other Python 3.6 are compelling enough of a feature we should update setup.py to remove < Python 3.5 compatibility so that users are aware of the lack of compatibility.  Otherwise, I think we should limit f-string usage to just tests.

Lucaweihs:
Good point regarding f-strings. Do we still actually support python 2.7? I think it's safe to deprecate this.

Lucaweihs (approved):

mattdeitke:
Also remove it from the README: https://github.com/allenai/ai2thor#requirements

ekolve:
> Good point regarding f-strings. Do we still actually support python 2.7? I think it's safe to deprecate this.

Y, we get import errors with Python 2.7.  Something to note, we have the following code in the ai2thor/controller.py
`self.last_event.metadata.get("errorMessage", f"{action} failed"`
Which means that importing that module will result in an error with Python 3.5.  We haven't received any issues indicating users are getting errors with this, though users may not be reporting.

PR #703, opened 2021-04-16
adding fix for ImageSynthesis enabled and resolution change

### LINKED ISSUE #477: Simulator freeze when 'ChangeResolution' called after initializing with renderObjectImage=True
Minimum to reproduce in ipython:

`In [1]: import ai2thor`
`In [2]: ai2thor.__version__`
`Out[2]: '2.4.10'`
`In [3]: from ai2thor.controller import Controller`
`In [4]: c = Controller(renderObjectImage=True)`
`Initialize return: {'cameraNearPlane': 0.1, 'cameraFarPlane': 20.0}`
`In [5]: c.step(action="ChangeResolution", x=500, y=500)`

The final call ([5]) never returns.

ekolve:
Fixes #477


ai2thor/tests/test_unity.py
```diff
@@ -1004,6 +1005,32 @@ def test_get_reachable_positions(controller):
     except:
         pass
 
+# Test for Issue: 477
```
Lucaweihs, line 1008: Should probably run black on this file.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -766,6 +766,44 @@ private void Update()
...
+		this.resetAllImageSynthesis();
+		this.primaryAgent.actionFinished(true);
+	}
+
+	public void ChangeResolution(float x, float y) {
+		int height = Convert.ToInt32(y);
+		int width = Convert.ToInt32(x);
+		Screen.SetResolution(width, height, false);
+		StartCoroutine(WaitOnResolutionChange(width, height));
+	}
```
Lucaweihs, line 806: With the new action dispatcher, can we make this `ChangeResolution(int x, int y)`? I believe they were floats only because `ServerAction.x` and `ServerAction.y` are.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -766,6 +766,44 @@ private void Update()
...
+		{
+			resetImageSynthesis(camera);
+		}
+	}
+	
+	public IEnumerator WaitOnResolutionChange(int width, int height) {
+		while (Screen.width != width || Screen.height != height) {
+			yield return null;
+		}
+		this.resetAllImageSynthesis();
```
Lucaweihs, line 797: Is it possible that `resetAllImageSynthesis` causes a change in which colors represent which objects for the instance/semantic segmentation frames? I remember the code for doing this assignment was a bit complex.

Lucaweihs (commented):
Minor comments but otherwise LGTM.

ai2thor/tests/test_unity.py
ekolve: That comment actually passed black, but i merged in main to get all the other reformatted code.

unity/Assets/Scripts/AgentManager.cs
ekolve: Good catch. Changed.

unity/Assets/Scripts/AgentManager.cs
ekolve: It isn't possible (at least in this case) for the colors to change since we take the MD5 of the ObjectId of the SimObject, which can't change during the ChangeResolution action.  Just to protect us a bit in the future I added assertions in test_unity.py that compare the pre and post color_to_object_id dictionaries on the event.

Lucaweihs: 👍  I suppose I was more worried that someone would call an action which reassigns all object ids without reassigning the colors (does this ever happen?), grabs the mapping of colors->objects, and then `ChangeResolution` (which would then update all the colors). I suppose this isn't really an issue with this PR but it would be nice if we had a method for assigning colors that was more invariant to these types of issues (as the object id seems to be more of an "object id" nowadays).

ekolve: I'm not sure if thats possible since the the ObjectId is assigned when the scene loads based off the position and even if the object moves, the objectId remains the same.  But your point is valid, since even the slightest change in the scene to the object results in a new ID and a new color.

PR #705, opened 2021-04-19
Look/Rotate fix and updates

winthos:
Fix to bugged behavior when defaulting the agent hand position and rotation after a LookUp/Down action.

Additionally adds a check for the overlap box cast at the end of a rotate check so that it correctly ignores any held object colliders and the agent's collider.


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -437,13 +439,16 @@ public override void RotateRight(ServerAction action)
...
+                base.RotateRight(action);
+
                 //only default hand if not manually Interacting with things
                 if(!action.manualInteract)
                 {
                     DefaultAgentHand();
                 }
 
-                base.RotateRight(action);
+                return;
```
mattdeitke, line 451: Can you remove these `return` statements? It doesn't look like anything comes after them.

PR #709, opened 2021-04-20
webgl unpause physics

AlvaroHG:
* moves pause, unpause and advancephysics to PhysicSceneManager
* unpauses when webgl controller is set

Need to test.

FPS controller in Editor pickup action still broken, we need to fix general unpausing for certain modes. Perhaps an Initialize flag that is set to "pause" by default if that is the desired behavior? All client code that calls initialize would have to change though, so I think "pausing" should be the explicit flag and "unpaused" the default at the current THOR stage.


unity/Assets/Scripts/PhysicsSceneManager.cs
```diff
@@ -817,6 +818,55 @@ System.Random rng
         }
     }
 
+    //pause physics autosimulation! Automatic physics simulation can be resumed using the UnpausePhysicsAutoSim() action.
+    //additionally, auto simulation will automatically resume from the LateUpdate() check on AgentManager.cs - if the scene has come to rest, physics autosimulation will resume
```
AlvaroHG, line 822: Comment seems outdated?

ekolve:
lgtm

ekolve (approved):

AlvaroHG:
Added webgl-demo-fixes that were not on main

PR #714, opened 2021-04-20
Possible dispatcher invalid arg fix

AlvaroHG:
Making dispatcher fail when an argument that does not exist as a parameter for a particular action, is passed to Dispatch.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -1771,6 +1771,11 @@ namespace UnityStandardAssets.Characters.FirstPerson
             {
                 ActionDispatcher.Dispatch(target: target, dynamicServerAction: controlCommand);
             }
+            catch (InvalidArgumentsException e) {
+                errorMessage = $"action: ${controlCommand.action} called with invalid arguments:" + string.Join(",", e.ArgumentNames.ToArray());
+                errorCode = ServerActionErrorCode.InvalidArgument;
+                actionFinished(false);
+            }
             catch (ToObjectArgumentActionException e)
             {
                 Dictionary<string, string> typeMap = new Dictionary<string, string>{
--- a/unity/Assets/Scripts/ActionDispatcher.cs
+++ b/unity/Assets/Scripts/ActionDispatcher.cs
@@ -2,6 +2,7 @@
 using System.Reflection;
 using System.Collections.Generic;
 using System;
+using System.Linq;
 using UnityEngine;
 using Newtonsoft.Json.Linq;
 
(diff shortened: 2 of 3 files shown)


ekolve:
Can you add a test(s) to ai2thor/test/test_unity.py that includes extraneous parameters to one or more functions and asserts the correct exception is thrown?

ekolve:
I really like this change. I think it has the potential to save a lot of user's time debugging potentially mis-dispatched calls at the expense of resolving some step() calls that include extraneous params.  I want to make sure everyone is on board with this change and we should also do our best to publicize the breaking change when its released.

mattdeitke:
👍  Definitely a good change! I had previously just assumed that this was ready happening.

I haven't tested the implementation, so maybe this is already the case, but I assume (and think it should) fail loudly (i.e., throw an exception and not just have `lastActionSuccess=False`) from the Python side if this error occurs.

AlvaroHG:
adding the tests, and yes it would be one of those breaking changes

ekolve:
Would it be possible to merge your change to a branch off of the allenai/ai2thor repo? We don't have a way to whitelist forks for travis-ci to build, so we won't know whether the tests are all passing prior to merging.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1691,7 +1691,12 @@ public Quaternion TargetRotation
 
             try {
                 ActionDispatcher.Dispatch(target: target, dynamicServerAction: controlCommand);
+            } catch (InvalidArgumentsException e) {
+                errorMessage = $"action: ${controlCommand.action} called with invalid arguments:" + string.Join(",", e.ArgumentNames.ToArray());
+                errorCode = ServerActionErrorCode.InvalidArgument;
+                actionFinished(false);
```
Lucaweihs, line 1697: Can we enrich this error message a bit by adding a list of all the ways the given action can be called? I.e. I would like to see an error message that returns something like:
```
You have attempted to use the "PushObject" action with invalid extraneous arguments.
Expected arguments: ...
Your arguments: ...
Invalid arguments: ...

Valid ways to invoke the for "PushObject" action include:
# 1 - PushObject(objectId, ...)
# 2 - PushObject(objectId, ...)
# 3 - PushObject(objectId, ...)
```

Lucaweihs (commented):
Big fan of this change, one small comment regarding increasing the amount of information presented in the error message.

unity/Assets/Scripts/BaseFPSAgentController.cs
AlvaroHG: that is a good idea let me see how much restructuring it needs to get all the overwrites, since at that point in the code we had selected a particular, best matching, overwrite

AlvaroHG:
Updated version in repo PR #717

PR #716, opened 2021-04-21
csharp_prefer_braces / brace fixes

ekolve:
Adding [csharp_prefer_braces](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0011#csharp_prefer_braces) to editorconfig as an error.  This will cause violations to appear when a linter is run as well as within IDEs.  Resharper codecleanup was used to fix all the existing violations of the rule.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,9 @@ root = true
 [*]
 indent_style = space
 
+[unity/Assets/Scripts/**.cs]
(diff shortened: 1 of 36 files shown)


mattdeitke (approved):
Ooh, yeah. That's quite nice.

Lucaweihs (approved):
This makes me very happy.

PR #717, opened 2021-04-22
Dispatcher Invalid Argument Fix

AlvaroHG:
Does what #714 from a non-forked branch

Adds some restructuring for testing "step" both in unity and python from the same entrypoint.


AlvaroHG:
@Lucaweihs Example output of failure errorMessage:
```
Action: "PutObject" called with invalid argument: z
Expected arguments: Single x, Single y, Boolean forceAction = False, Boolean placeStationary = True, Int32 randomSeed = 0, Boolean putNearXY = False
Your arguments: x, y, z, forceAction, placeStationary
Valid ways to call "PutObject" action:
	Void PutObject(Single x, Single y, Boolean forceAction = False, Boolean placeStationary = True, Int32 randomSeed = 0, Boolean putNearXY = False)
	Void PutObject(String objectId, Boolean forceAction = False, Boolean placeStationary = True, Int32 randomSeed = 0)'.
```
Decided to put the invalid argument first as it may be the first thing people read and what triggered the error.

(The your arguments section seems a bit redundant to me, I had that out initially, but I can imagine the use when having many calls, note that types are not known for arguments at this point so it's just the barebones names)

I can add an action endpoint to return these method signatures the `Valid ways to call "..."` since it may be useful to users and for future automated documentation, future api generation.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1151,34 +1134,64 @@ public class OptionalVector3 {
...
+	// Extra helper functions
+	protected string LoadStringVariable(string variable, string name)
+	{
+		string envVarName = ENVIRONMENT_PREFIX + name.ToUpper();
+		string envVarValue = Environment.GetEnvironmentVariable(envVarName);
+		return envVarValue == null ? variable : envVarValue;
+	}
+
+	protected int LoadIntVariable(int variable, string name)
+	{
```
ekolve, line 1176: Can you check your formatter settings? The first brace shouldn't be on a new line if .editorconfig exists in the root.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1538,7 +1551,7 @@ public class DynamicServerAction {
 
     public int sequenceId {
         get {
-            return (int)this.jObject["sequenceId"];
+            return this.GetValue("sequenceId", 0);
```
ekolve, line 1554: This should be reverted.  We can't default the sequenceId as it should always originate from the Python side.  If the sequenceId is missing, then something is really wrong.  It would be better to just check to see if it exists and throw an error if it doesn't.

ekolve (changes requested):

unity/Assets/Scripts/AgentManager.cs
AlvaroHG: Fixed

unity/Assets/Scripts/AgentManager.cs
AlvaroHG: Fixed as conversation we had

ai2thor/tests/test_unity.py
```diff
@@ -1598,3 +1598,22 @@ def test_unsupported_manipulathor(controller):
...
+            y=0.0,
+            z=1.0,
+            forceAction=False,
+            placeStationary=True
+        )
+        print("Error code {0}".format(event.metadata["errorCode"]))
+        assert not event.metadata["lastActionSuccess"], "Extra parameter 'z' in action"
+        assert event.metadata[
+            "errorMessage"
+        ], "errorMessage with invalid argument"
```
Lucaweihs, line 1619: I'm a bit confused about the semantics of `pytest.raises(ValueError)`. Here we should expect that this call results in a `ValueError` being raised, correct? Shouldn't this mean that the asserts won't be hit?

Lucaweihs (approved):
I did not read the implementation details closely but I'm very happy with the changes from a usability standpoint. I had one minor question w.r.t. the tests but I'll approve preemptively as I don't really expect it to be a real issue.

ekolve (approved):

ai2thor/tests/test_unity.py
AlvaroHG: Good point, this meant to be escaped left, now fixed

AlvaroHG:
merging!

PR #727, opened 2021-04-26
local-build-test task / local-build with scenes

ekolve:
Typically when I want to run the tests or just experiment with a few scenes in a Unity Player I will modify Build.cs to include just the scenes I care about.  This PR modifies the `local-build` task to accept a scene(s) argument and introduces a new task `local-build-test` that just builds the player with the TEST_SCENE (currently FloorPlan28). The following is now possible

```python
invoke local-build -s FloorPlan28
invoke local-build -s FloorPlan28_physics -s FloorPlan1
```

### DIFF
--- a/unity/Assets/Editor/Build.cs
+++ b/unity/Assets/Editor/Build.cs
@@ -12,7 +12,7 @@ public class Build {
 #else
 		var buildTarget = BuildTarget.StandaloneOSXIntel64;
 #endif
-        build(GetBuildName(), GetAllScenePaths(), buildTarget);
+        build(GetBuildName(), buildTarget);
     }
 
     static string GetBuildName() {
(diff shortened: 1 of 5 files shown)


unity/Assets/Editor/Build.cs
```diff
@@ -12,43 +12,55 @@ public class Build {
...
         if (defines != "") {
             var targetGroup = BuildPipeline.GetBuildTargetGroup(target);
             PlayerSettings.SetScriptingDefineSymbolsForGroup(targetGroup, GetDefineSymbolsFromEnv());
         }
+
+        List<string> scenes = GetScenes();
+        foreach (string scene in scenes) {
+            Debug.Log("Adding Scene " + scene);
+        }
         BuildPipeline.BuildPlayer(scenes.ToArray(), buildName, target, BuildOptions.StrictMode | BuildOptions.UncompressedAssetBundle);
```
Lucaweihs, line 41: I've found that if you don't change any of the scene files (i.e. when you change only scripts) replacing
```
BuildOptions.StrictMode | BuildOptions.UncompressedAssetBundle
```
with
```
BuildOptions.BuildScriptsOnly | BuildOptions.StrictMode | BuildOptions.UncompressedAssetBundle
```
can result in some non-trivial speedups on sequential builds. Not a super fair request but can we add in an option to `local_build` to only build scripts like this? I.e. ``--build_scripts_only t`` or similar?

Lucaweihs (approved):
Great update, I made one "feature request" comment but understand that this is perhaps somewhat beyond the scope of this PR.

mattdeitke (approved):
Just tested. Works for me :)

Definitely much nicer than editing `Build.cs` each time.

unity/Assets/Editor/Build.cs
ekolve: Thats a good suggestion. I'll handle that as a separate PR.

PR #730, opened 2021-04-27
Throw ValueError on invalid Controller(commit_id)

ekolve:


Lucaweihs (approved):
LGTM. If we're in the process of adding more informative errors, can we add some more information / links to the `startx.py` script for when the x_display is invalid or cannot be found? [I also wrote a script to return all open x displays](https://github.com/allenai/allenact/blob/main/allenact_plugins/ithor_plugin/ithor_util.py#L61) in AllenAct which might be useful to return a more informative error message, e.g. something like:
```
$ c = Controller(x_display="notvalid")
---------------------------------------------------------------------------
RuntimeError: Could not find an open `x_display` of name "notvalid", your currently open x-displays are ["0.0", "0.1"]. Please use one of these or, if there are none, use our startx.py script (link) to start x-displays on all of your GPUs. You will require sudo privileges to start new x-displays.
```

PR #735, opened 2021-04-30
Refactoring Invalid ActionDispacher

ekolve:
Refactor invalid ActionDispatcher params logic to allow renderInstanceSegmentation to be both an action param and per step param.

### DIFF
--- a/unity/Assets/Scripts/ActionDispatcher.cs
+++ b/unity/Assets/Scripts/ActionDispatcher.cs
@@ -322,7 +322,7 @@ public static class ActionDispatcher {
             var paramDict = methodParams.ToDictionary(param => param.Name, param => param);
             var invalidArgs = dynamicServerAction
                 .Keys()
-                .Where(argName => !paramDict.ContainsKey(argName) && argName != "action")
+                .Where(argName => !paramDict.ContainsKey(argName) && !DynamicServerAction.AllowedExtraneousParameters.Contains(argName))
                 .ToList();
             if (invalidArgs.Count > 0) {
                 Func<ParameterInfo, string> paramToString =
--- a/unity/Assets/Scripts/ActionDispatcher.cs
+++ b/unity/Assets/Scripts/ActionDispatcher.cs
@@ -321,8 +321,8 @@ public static class ActionDispatcher {
         } else {
             var paramDict = methodParams.ToDictionary(param => param.Name, param => param);
             var invalidArgs = dynamicServerAction
-                .Keys()
-                .Where(argName => !paramDict.ContainsKey(argName) && !DynamicServerAction.AllowedExtraneousParameters.Contains(argName))
+                .ArgumentKeys()
+                .Where(argName => !paramDict.ContainsKey(argName))
                 .ToList();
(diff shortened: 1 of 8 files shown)


unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -322,7 +322,7 @@ public static class ActionDispatcher {
             var paramDict = methodParams.ToDictionary(param => param.Name, param => param);
             var invalidArgs = dynamicServerAction
                 .Keys()
-                .Where(argName => !paramDict.ContainsKey(argName) && argName != "action")
+                .Where(argName => !paramDict.ContainsKey(argName) && !DynamicServerAction.AllowedExtraneousParameters.Contains(argName))
```
AlvaroHG, line 325: I like that everything is under `DynamicServerAction.AllowedExtraneousParameters` the only thing is that I feel dispatch should be called with the method `arguments` only, so by this point we shouldn't be filtering. If we did this we also need to change line `342` which passes the arguments for the exception to print the message now the message will look: `you called with arguments: 'x', 'y', 'z', 'action', 'sequenceId'`

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1122,21 +1113,19 @@ public class OptionalVector3 {
...
-        var args = new DynamicServerAction(
-            new JObject(
-                controlCommand.jObject.Properties().Where(p => !injectedArguments.Contains(p.Name))
-            )
-        );
-
         if (agentManagerActions.Contains(controlCommand.action)) {
             // let's look in this class for the action
-            this.activeAgent().ProcessControlCommand(controlCommand: args, target: this);
+            this.activeAgent().ProcessControlCommand(controlCommand: controlCommand, target: this);
```
AlvaroHG, line 1118: Now that `action` is included, I would do the filtering in `BaseFPSAgent`, `ProcessControlCommand` right before the dispatch, and create a new args object. Like make sure that Dispatch is called always with the right arguments, I think that will simplify logic and reduce future bugs.

AlvaroHG:
One suggestion

unity/Assets/Scripts/ActionDispatcher.cs
ekolve: The problem is that we have an argument `renderInstanceSegmentation` that can be used as a global and action argument.  So we have to defer filtering.

ekolve: Your point about the exception message is good - i'll refactor to take that into account.

unity/Assets/Scripts/ActionDispatcher.cs
```diff
@@ -339,7 +339,7 @@ public static class ActionDispatcher {
                 );
 
                 throw new InvalidArgumentsException(
-                    dynamicServerAction.Keys().Where(argName => argName != "action"),
+                    dynamicServerAction.ArgumentKeys(),
```
AlvaroHG, line 342: great!

AlvaroHG:
LGTM

PR #739, opened 2021-05-03
local-build with BuildOptions.BuildScriptsOnly 

ekolve:
Adding option to local-build to include BuildScriptsOnly when running a build.  The following is now possible:

```bash
invoke local-build --scripts-only
```

I found that in order for the build run execute with any speedup `BuildOptions.Development` also needed to be included with the options.


Lucaweihs (approved):
LGTM! I wish there was a way for unity to detect whether or not the scenes had changed from the last build, this would make it possible to set this automatically when not changed.

PR #740, opened 2021-05-04
Use python-xlib for X selecting and checking X server

ekolve:
This PR removes xdpyinfo and uses python-xlib to validate (display is running and has the GLX extension loaded).  If no x_display is passed in and a DISPLAY is not present in the environment one will be selected from the list of valid_displays.

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -952,24 +952,83 @@ class Controller(object):
         # print("Viewer: http://%s:%s/viewer" % (host, port))
         command = self.unity_command(width, height, headless=self.headless)
         makedirs(self.log_dir)
-        self.server.unity_proc = proc = subprocess.Popen(command, env=env, stdout=open(os.path.join(self.log_dir, 'unity.log'), "a"))
+        self.server.unity_proc = proc = subprocess.Popen(
+            command, env=env, stdout=open(os.path.join(self.log_dir, "unity.log"), "a")
+        )
         self.unity_pid = proc.pid
         atexit.register(lambda: proc.poll() is None and proc.kill())
 
-    def check_x_display(self, x_display):
-        with open(os.devnull, "w") as dn:
-            # copying the environment so that we pickup
-            # XAUTHORITY values
-            env = os.environ.copy()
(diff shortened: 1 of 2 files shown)


ai2thor/controller.py
```diff
@@ -952,25 +952,84 @@ def _start_unity_thread(self, env, width, height, server_params, image_name):
...
+            return ":0.0"
+
+    def _valid_displays_message(self):
+        valid_displays = self._valid_x_displays()
+        if valid_displays:
+            return "The following valid displays were found %s" % (
+                ", ".join(valid_displays)
+            )
+        else:
+            return ""
```
Lucaweihs, line 1003: Instead of "" I think this should be "We couldn't find any open x displays.".

ai2thor/controller.py
```diff
@@ -952,25 +952,84 @@ def _start_unity_thread(self, env, width, height, server_params, image_name):
...
+                raise ValueError(
+                    "Display %s does not have the GLX extension loaded.  GLX is required by Unity3D. %s"
+                    % (x_display, self._valid_displays_message())
                 )
 
+        except (Xlib.error.DisplayNameError, Xlib.error.DisplayConnectionError) as e:
+            raise ValueError(
+                "Invalid display: %s - %s. %s"
+                % (x_display, e, self._valid_displays_message())
+            )
```
Lucaweihs, line 1031: Can we add a little more information to this error message? In particular I think it would be great if it told people what the problem was (i.e. either "we could not find a valid x_display" or "the given x_display was bad") and gave them some information about how to trouble shoot / fix this (i.e. "If no X-display is currently running on your machine, please run our `startx.py` (link) script as `sudo python startx.py &` which will start X-servers on each of your GPUs. Note that you will need administrator privileges to start these X-servers, if you do not have such privileges please contact your server administrator.".

Lucaweihs (commented):
Very nice, I think this will save people from a lot of issues when they first start using THOR! I've added some minor comments that I think might improve things even further.

ekolve:
This PR is replaced by PR #774.

PR #741, opened 2021-05-05
Incorporate More Materials

mattdeitke:
The first of a few material sets I'm looking at incorporating.. Still needs to be cleaned up before it's ready to review.

Empirically it makes `RandomizeMaterials` look slightly less realistic in some cases, but the variance with these materials is ginormous.

TODOs:
- Check how much larger these files make the build.
- Assign OfficeCeiling, Marble, and Concrete to different material types.
- Review the existing material assignment decisions, some empirically appear to be a bit off.

NOTE: These changes solely affect `RandomizeMaterials` and nothing else. Users have the ability to opt out of using external materials, if they want to.

Ping me offline for more.


Lucaweihs:
Closing this for now, may come back to it in the future.

PR #754, opened 2021-05-15
Remove lights from randomizable materials

mattdeitke:
Solves a visual bug that only happens in a few scenes and was discovered after adding the initial `RandomizeMaterials` to the demo. Since some light rays are transparent, and others, apparently, are not, randomizing a transparent light ray to a solid light ray looks... bad.

This PR simply removes lighting materials from materials that may be randomized.

image

image


Lucaweihs (approved):
LGTM.

ekolve (approved):

PR #757, opened 2021-05-17
[Demo] Slow down throw speed

mattdeitke:
Currently, throwing an object is uncontrollably fast. I'm not sure if 400f is low enough, since I can't test it until it is live. But, it seems like a reasonable starting point to try.


winthos (approved):
looks good to me

PR #758, opened 2021-05-17
When setting object poses, reset the parents of all moved objects to the top-level object.

Lucaweihs:
The `SetObjectPoses` method fails to reset the object hierarchy when moving objects. This can lead to weird results where an object across the room will be the child of a drawer across the room. This can break the semantic hull code as it assumes that all meshes that are within an object in the hierarchy correspond to that object. The solution here is simply to unparent objects when running `SetObjectPoses` (making them children of the top level "Objects" object). I have also run black/C# formatting so there are changes corresponding to this.

### DIFF
--- a/ai2thor/_quality_settings.py
+++ b/ai2thor/_quality_settings.py
@@ -1,11 +1,13 @@
 # GENERATED FILE - DO NOT EDIT
-DEFAULT_QUALITY = 'Ultra'
-QUALITY_SETTINGS = {'DONOTUSE': 0,
- 'High': 5,
- 'High WebGL': 8,
- 'Low': 2,
- 'Medium': 3,
- 'MediumCloseFitShadows': 4,
- 'Ultra': 7,
- 'Very High': 6,
- 'Very Low': 1}
+DEFAULT_QUALITY = "Ultra"
+QUALITY_SETTINGS = {
+    "DONOTUSE": 0,
+    "High": 5,
+    "High WebGL": 8,
+    "Low": 2,
+    "Medium": 3,
+    "MediumCloseFitShadows": 4,
+    "Ultra": 7,
+    "Very High": 6,
+    "Very Low": 1,
+}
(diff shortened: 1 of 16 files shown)


winthos:
One thing i'm a little concerned about is if the placeStationary flag is used for SetObjectPoses, ideally we'd want any receptacles to be the parent of the child object because in the case of something like placing a fork in a drawer, if placeStationary = TRUE then all the rigidbodies will have kinematic true set. This means, if the drawer were to open and the fork isn't a child of the drawer, the fork will float in place where it was. This isn't an issue when placeStationary is false, as physics resolution will move the fork with the drawer.

This leads to perhaps a larger discussion of how we want to resolve mismatched initial states in terms of "stationary" mode and "physics" mode. We have many actions like `InitialRandomSpawn`, `SetObjectPoses` and even the simple `PutObject` that have the option to place an object at a position and either let physics resolve its final resting position and orientation, or to freeze the position upon placement by making it "stationary."

The way that we currently freeze objects is by setting their rigibody to `kinematic = True`, and here is where we have the problem if mixing a potentially stationary object with subsequent actions that might expect more mobility either through physics resolution or animations like opening/closing a drawer.

These paradigms are fundamentally at odds with each other, and the flow chart of how the object behavior is expected to be based on if it started stationary vs started physics determined gets more muddled especially with things like semantic hull which assume all child mesh renderers belong to the object in question, which is not the case when putting objects in receptacles in a way that makes them parented.

Lucaweihs:
Totally agree @winthos that we need more consistency with how we handle "kinematic" objects, especially when teleporting objects around the scene. I think people generally don't even want things to be truly kinematic as it's quite strange to force an object to stay in a single position no matter what. Really I think people want something similar to what I did for rearrangement: after placing an object we just don't want it to move until it's acted on, this is accomplished pretty well by manually putting the object's rigidbody to sleep. I know this solution doesn't work well in all cases but maybe this can be what "placeStationary" does when not in the proposed "kinematic" mode.

This all said: I think this PR is much better than the alternative (where an object may stay parented to drawer even after it has been teleported across the room) and so don't think the above discussion should delay merging (after NeurIPS of course).

winthos (approved):
looks good, and we will have a bigger chat about stationary/not stationary modes sometime later

PR #759, opened 2021-05-17
adding default values to near/far clipping planes

winthos:
In regards to 3rd party cameras, this fixes an issue where light and shadow resolutions were dynamically scaled based on the clipping plane values, so defaulting this to the same clipping planes as the agent camera for parity

### DIFF
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -420,8 +420,8 @@ public class AgentManager : MonoBehaviour {
         string skyboxColor,
         bool? orthographic,
         float? orthographicSize,
-        float? nearClippingPlane = null,
-        float? farClippingPlane = null
+        float? nearClippingPlane,
+        float? farClippingPlane
     ) {
         if (orthographic != true && orthographicSize != null) {
             throw new InvalidOperationException(
@@ -489,8 +489,8 @@ public class AgentManager : MonoBehaviour {
         string skyboxColor = null,
         bool orthographic = false,
         float? orthographicSize = null,
-        float? nearClippingPlane = null,
-        float? farClippingPlane = null
+        float nearClippingPlane = 0.1f,
+        float farClippingPlane = 20.0f
     ) {
         // adds error if fieldOfView is out of bounds
         assertFovInBounds(fov: fieldOfView);
@@ -551,7 +551,9 @@ public class AgentManager : MonoBehaviour {
         float? fieldOfView = null,
         string skyboxColor = null,
         bool? orthographic = null,
-        float? orthographicSize = null
+        float? orthographicSize = null,
+        float? nearClippingPlane = null,
+        float? farClippingPlane = null
     ) {
         // adds error if fieldOfView is out of bounds
         if (fieldOfView != null) {
@@ -582,7 +584,9 @@ public class AgentManager : MonoBehaviour {
             fieldOfView: fieldOfView == null ? thirdPartyCamera.fieldOfView : (float)fieldOfView,
             skyboxColor: skyboxColor,
             orthographic: orthographic,
-            orthographicSize: orthographicSize
+            orthographicSize: orthographicSize,
+            nearClippingPlane: nearClippingPlane,
+            farClippingPlane: farClippingPlane
         );
     }
 
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -451,8 +451,18 @@ public class AgentManager : MonoBehaviour {
             camera.nearClipPlane = (float)nearClippingPlane;
         }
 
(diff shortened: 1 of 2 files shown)


unity/Assets/Scripts/AgentManager.cs
```diff
@@ -471,7 +488,9 @@ private enum serverTypes { WSGI, FIFO };
         float fieldOfView = DEFAULT_FOV,
         string skyboxColor = null,
         bool orthographic = false,
-        float? orthographicSize = null
+        float? orthographicSize = null,
+        float? nearClippingPlane = null,
+        float? farClippingPlane = null
```
mattdeitke, line 493: It looks like these do not need to be optional `float?`, since they are being parsed as:
```c#
        if (farClippingPlane == null) {
            camera.farClipPlane = 20.0f;
        } else {
            camera.farClipPlane = (float) farClippingPlane;
        }
```
can:

- `float? farClippingPlane = null` become `float farClippingPlane = 20.0f`
- `float? nearClippingPlane = null` become `float nearClippingPlane = 0.1f`

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -419,7 +419,9 @@ private enum serverTypes { WSGI, FIFO };
         float fieldOfView,
         string skyboxColor,
         bool? orthographic,
-        float? orthographicSize
+        float? orthographicSize,
+        float? nearClippingPlane = null,
+        float? farClippingPlane = null
```
mattdeitke, line 424: Can
```c#
        float? nearClippingPlane = null,
        float? farClippingPlane = null
```
not have defaults for the private method?

Perhaps this is just a personal style preference, but since this is a private method, I generally like how every single call to it must specify all parameters.

Here, if new parameters are added, like `nearClippingPlane` and `farClippingPlane`, they have to explicitly be passed in at every place the private method is called, and if they're not, a compiler warning will appear.

In short, I think this helps ensure that public parameters are always passed to the private method.

mattdeitke (changes requested):

unity/Assets/Scripts/AgentManager.cs
winthos: ok check the new commit, made changes

unity/Assets/Scripts/AgentManager.cs
winthos: sounds good, changes in

mattdeitke (approved):
Looks good. The only other thing I'd say is to autoformat it, but I'm not sure if there's going to be a more automatic process for that.

ekolve:
It should be possible to format the block using VSCode if you highlight and select Format Selection.  VSCode will respect the .editorconfig that is checked in (or at least it should).

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -442,6 +444,21 @@ private enum serverTypes { WSGI, FIFO };
...
+            camera.nearClipPlane = 0.1f;
+        }
+
+        else { camera.nearClipPlane = (float) nearClippingPlane;}
+
+        if(farClippingPlane == null) {
+            camera.farClipPlane = 20.0f;
+        }
+
+        else { camera.farClipPlane = (float) farClippingPlane;}
```
mattdeitke, line 460: Wait, one other bug (appears twice, also for `nearClipPlane`)... If these are `null` (from UpdateThirdPartyCameras), then they should be keeping their current values.

Can
```c#
        if (farClippingPlane == null) {
            camera.farClipPlane = 20.0f;
        } else {
            camera.farClipPlane = (float) farClippingPlane;
        }
```

then be changed to
```c#
        if (farClippingPlane != null) {
            camera.farClipPlane = (float) farClippingPlane;
        }
```

winthos: gotchya, changed with latest commit. Also formatted all changes (Format Selection works flawlessly)

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -471,7 +484,9 @@ private enum serverTypes { WSGI, FIFO };
         float fieldOfView = DEFAULT_FOV,
         string skyboxColor = null,
         bool orthographic = false,
-        float? orthographicSize = null
+        float? orthographicSize = null,
+        float nearClippingPlane = 0.1f,
+        float farClippingPlane = 20.0f
```
ekolve, line 489: Instead of hardcoding these values can we instead grab them from this.primaryAgent.m_Camera.nearClipPlan/farClipPlane within the method if the passed in value is null?

ekolve (changes requested):

unity/Assets/Scripts/AgentManager.cs
winthos: updated with the latest commit

ekolve (approved):

PR #761, opened 2021-05-18
Save state of gameObject's layer for all colliders during oobb generation

ekolve:
Prior to moving gameObjects to the NonInteractive layer, save their state.  This fixes an issue where a component that has two colliders gets restored incorrectly to the NonInteractive layer since the second iteration through the loop causes the gameObject to get restored to the NonInteractive layer.


winthos:
Lgtm

winthos (approved):

ekolve:
I'm going to hold off on merging this since it affects everything until after May 26th.

PR #762, opened 2021-05-18
CS/PY invoke format tasks

ekolve:
This PR adds three new tasks to format python and csharp code

```bash
invoke format # format all code
invoke format-py # format python code using black
invoke format-cs # format csharp code using dotnet-format
```

For `format-cs`, a dotnet runtime is installed into the `.dotnet` directory beneath the checked out repo and dotnet-format is installed as a local tool.


Lucaweihs (commented):
Looks like there were two LGTM alerts but otherwise seems great.

mattdeitke (approved):
LGTM

PR #763, opened 2021-05-19
ai2thor-xorg

ekolve:
This PR adds the script formerly known as `startx.py` as `ai2thor-startx` and also packages it as a script which gets added to  a user's PATH during pip installation.  This allows for the following commands from the shell:

```bash

sudo ai2thor-xorg start [optional display]
sudo ai2thor-xorg stop
sudo ai2thor-xorg print-config
```


ekolve:
The command `ai2thor-startx start` is a little odd since `start` appears twice.  Another possible name for this might be `ai2thor-xorg`.

Lucaweihs (approved):
LGTM. I think I would prefer `ai2thor-xorg` but I'm fine with either.

ekolve:
sg - i prefer ai2thor-xorg as well.

PR #765, opened 2021-05-20
Clears RandomizeColors on reset()

mattdeitke:
Matches the same `reset` functionality present in `RandomizeMaterials`.

While this breaks backwards compatibility, I'd consider it more of a bug fix, since calling `RandomizeColors` -> `reset()` -> `ResetColors` currently cannot reset the colors to their original ones. Moreover, this action is seldom used at this point and, I don't think, has even been used in a project yet.


Lucaweihs (approved):
LGTM. I also consider this a bug fix and thus have no complaints about merging now.

ekolve (approved):
LGTM - one question, why is it only possible to resetMaterials or resetColors, but not both?

mattdeitke:
> one question, why is it only possible to resetMaterials or resetColors, but not both?

`resetMaterials` both resets the materials _and_ resets the colors.

PR #768, opened 2021-05-24
DropHandObject -> DropHeldObject

mattdeitke:
Renames `DropHandObject` to `DropHeldObject`. In iTHOR, its unclear what the "hand" is.

Everything is backwards and sideways compatible. That is; users can still pass in `DropHandObject` to older and newer versions, and it will still work.

### DIFF
--- a/ai2thor/tests/test_unity.py
+++ b/ai2thor/tests/test_unity.py
@@ -1786,7 +1786,7 @@ def test_invalid_arguments(controller):
 
 @pytest.mark.parametrize("controller", fifo_wsgi)
 def test_drop_object(controller):
-    for action in ["DropObject", "DropHandObject"]:
+    for action in ["DropHeldObject", "DropHandObject"]:
         assert not controller.last_event.metadata["inventoryObjects"]
         controller.step(
             action="PickupObject",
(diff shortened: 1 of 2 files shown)


winthos (approved):
looks good to go

mattdeitke:
https://github.com/allenai/ai2thor/pull/768/commits/9fa68c90263fd0ff6bfd351572718a4056a13780 Just added a test that ensures backwards compatibility for `DropHandObject` and that `DropObject` is working.

Lucaweihs (commented):
LGTM. One annoying request: if you're messing with this action anyway, can you upgrade it so that it uses the new action dispatch api?

mattdeitke:
Probably. I tried doing this kinda thing in bulk to most of the actions back some time ago
https://github.com/allenai/ai2thor/pull/584

although my finding is that there are just so many actions, and so many actions that call other actions, that it is just so easy to break something, so I've kinda avoided it.

mattdeitke:
@Lucaweihs updated and slightly cleaned up in:
https://github.com/allenai/ai2thor/pull/768/commits/58bd7b5e2ad1000b3d01aa71c7ea7de71fe0c909

Lucaweihs (approved):
LGTM

PR #770, opened 2021-05-24
Move hand

mattdeitke:
* Renames `MoveHand{Ahead/Right/Back/Down/Left/Up}` to `MoveHeldObject{Ahead/Right/Back/Down/Left/Up}`, since that is what's actually happening as iTHOR does not have a "Hand" in the ManipulaTHOR sense of the word. 
* Renames `RotateHand(float x, float y, float z)` to `RotateHeldObject(Vector3 rotation)` to rotate the held object to an absolute rotation.
* Renames `RotateHandRelative(x, y, z)` to an overloaded version of `RotateHeldObject(pitch, yaw, roll)`. Bug fixes the roll direction such that positive values turn right.

Everything is backwards and sideways compatible.


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -2900,64 +2936,69 @@ public struct WhatDidITouch {
...
+            RotateHeldObject(rotation: new Vector3(x, y, z));
+        }
 
+        // rotate the held object if there is an object in it
+        public void RotateHeldObject(Vector3 rotation) {
             if (ItemInHand == null) {
-                errorMessage = "Can't rotate hand unless holding object";
-                actionFinished(false);
-                return;
+                throw new InvalidOperationException("Can't rotate hand unless holding object.");
```
winthos, line 2947: with the semantic change from "hand" to "held object" might as well change this exception message as well. Something like "Can't rotate object. No object held."

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -2900,64 +2936,69 @@ public struct WhatDidITouch {
...
         }
 
         // rotate the hand if there is an object in it
-        public void RotateHandRelative(ServerAction action) {
-
+        public void RotateHeldObject(float pitch = 0, float yaw = 0, float roll = 0) {
             if (ItemInHand == null) {
-                errorMessage = "Can't rotate hand unless holding object";
-                return;
+                throw new InvalidOperationException("Can't rotate hand unless holding object.");
```
winthos, line 2972: same as above, might as well update this message as well.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -2900,64 +2936,69 @@ public struct WhatDidITouch {
...
             );
             transform.rotation = agentRot;
 
             if (isHandObjectColliding(true)) {
-                errorMessage = "Hand object is coliding after rotation.";
                 AgentHand.transform.rotation = agentHandStartRot;
-                actionFinished(false);
-            } else {
-                actionFinished(true);
+                throw new InvalidOperationException("Hand object is coliding after rotation.");
```
winthos, line 2988: Also this. Perhaps update to "Held object is colliding after rotation."

winthos (changes requested):
Just a couple tweaks to error messages. Besides that looks good to go.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: Updated in https://github.com/allenai/ai2thor/pull/770/commits/dcc9ef823c6793f49dba3c0b7a2f1027f5096a1b.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: Updated in https://github.com/allenai/ai2thor/pull/770/commits/dcc9ef823c6793f49dba3c0b7a2f1027f5096a1b.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
mattdeitke: Updated in https://github.com/allenai/ai2thor/pull/770/commits/dcc9ef823c6793f49dba3c0b7a2f1027f5096a1b.

winthos (approved):
lgtm

ai2thor/tests/data/metadata-schema.json
```diff
@@ -1,662 +1,595 @@
...
-                    "toggleable",
-                    "visible"
-                ]
-            }
-        },
-        "isSceneAtRest": {
-            "type": "boolean"
-        },
-        "agent": {
+  "$schema": "http://json-schema.org/draft-07/schema",
```
mattdeitke, line 2: Applied autoformatting to this file. The only thing that actually changes is "hand" is now "heldObjectPose".

mattdeitke:
So a weird note on the held object rotation metadata. Due to Unity's use of quaternions to manage rotations, it's pretty unpredictable what the Euler angles will be.

As a simple example, consider the following sequences below, which shows that the rotations `(x=180, y=0, z=0)` and `(x=0, y=180, z=180)` are equivalent.

https://user-images.githubusercontent.com/28768645/119767920-a2ae2800-be6c-11eb-8959-86f2af9fea3d.mp4

Here, if you land on the rotation `(x=0, y=180, z=180)`, I've seen Unity convert it to both `(x=180, y=0, z=0)` and `(x=-180, y=0, z=0)`, and determining which one it gets converted to is pretty unpredictable.

So my concern is that the reported Euler rotations may jump unexpectedly, making them perhaps weird sensory inputs to a network.

mattdeitke:
A bunch of tests now added and passing.

Lucaweihs (commented):
While I understand the desire to use standard rotation terminology (pitch/yaw/roll) I worry that this just makes a lot of things inconsistent as we do not use these terms anywhere else in the code base. This also means that we have to remember that `z=-roll` which I could see causing confusion. Can we keep these as x/y/z?

Lucaweihs:
I meant more that we'll have to remember that `z==-roll` internally to THOR (i.e. in the C# code). Also, is it actually that clear? What's the roll access for a tomato?

mattdeitke:
If you're holding an object out in front of you, roll would always mean rotating it right, pitch would always mean rotating it up/down. I think this is more clear than `x` and `y`.

mattdeitke:
Any updates @Lucaweihs, I do still find pitch/yaw/roll (from the agent's point of view) to be much more intuitive names for relative rotations than `x/y/z`.

Lucaweihs:
I'm fine with this 👍

PR #771, opened 2021-05-24
Add fillLiquid to metadata

mattdeitke:
For [Fill Object with Liquid](https://ai2thor.allenai.org/ithor/documentation/object-state-changes/#fill-object-with-liquid):

* Adds `fillLiquid` to metadata .
* Bug fixes the action to make sure each failed path terminates immediately (i.e., with `throw new Exception()`)

### DIFF
--- a/ai2thor/tests/data/arm-metadata-schema.json
+++ b/ai2thor/tests/data/arm-metadata-schema.json
@@ -204,7 +204,7 @@
             "type": "boolean"
           },
           "fillLiquid": {
-            "type": "string"
+            "type": "object"
           },
           "dirtyable": {
             "type": "boolean"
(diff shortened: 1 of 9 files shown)


ekolve:
Can you add a test that checks the fill liquid of a mug, asserts its empty, put the mug in the coffee machine and fills with coffee, then assert that the metadata reports "coffee" as the fill liquid?

mattdeitke:
https://github.com/allenai/ai2thor/pull/771/commits/07a8e9a2a83102a465ce69aa4dda2fd74e9517e7 Tests added (and passing locally).

Should be ready to go.

ekolve (approved):

PR #777, opened 2021-06-03
ai2thor-xorg - adding option to exclude a device

ekolve:
This PR adds the following option to the ai2thor-xorg command:

```python
sudo ai2thor-xorg --exclude-device 0 start
```

This will ignore the first GPU identified by device id `0`.


Lucaweihs (approved):
LGTM

PR #780, opened 2021-06-04
Support `RandomizeMaterials` and `RandomizeColors` with Image Synthesis on.

mattdeitke:
For an explanation, read this interactive notebook which visually shows what is going on and what happens if one naively removes the disabling out of the light rays (i.e., they may appear in the segmentation frames):

https://colab.research.google.com/drive/16kzdy7DNc5dCPHBt5z0Qgm3FWvy6-4wk?usp=sharing

image

The only way this may fail is if a light ray object (which only occurs in a few rooms, and also caused #754 ) does not actually have the name "LightRay". However, in the few scenes I've found that do have light rays, all of them are named `LightRay`, and there doesn't appear to be a better way to find them.

image

### DIFF
--- a/unity/Assets/Scripts/ImageSynthesis/ImageSynthesis.cs
+++ b/unity/Assets/Scripts/ImageSynthesis/ImageSynthesis.cs
@@ -318,7 +318,6 @@ public class ImageSynthesis : MonoBehaviour {
                 objTag = sop.ObjectID;
             }
 
-
             Color classColor = ColorEncoding.EncodeTagAsColor(classTag);
             Color objColor = ColorEncoding.EncodeTagAsColor(objTag);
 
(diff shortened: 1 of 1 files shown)


winthos (approved):
looks good to me

mattdeitke:
I don't have a really satisfying answer. My intuition for why the light rays were disabled in the first place came from https://github.com/allenai/ai2thor/pull/754, where light rays showed up as opaque objects, instead of being transparent. It then seemed reasonable that this would affect the segmentation frames (which it did, as shown in Colab).

My hypothesis is that `r.material.name.ToLower().Contains("lightray")` alone is `true` for a lot of materials, whereas `r.gameObject.name.ToLower().Contains("lightray")` is only true if the gameObject is actually a light ray, thus many objects were being disabled.

ekolve (approved):

Lucaweihs (approved):
Such a weird bug but I'm happy that it's so easy to fix. LGTM.

PR #786, opened 2021-06-09
Fix to RemoveFromScene objectId assignment

winthos:
instead of re-assigning objectId every time an object is removed from the scene, only the key for that object is removed from the objectIdToSimObjPhysics dict.

This prevents issues with objectId being re-assigned incorrectly due to floating point errors.

Additionally, moved DisableAllObjectsOfType to base controller to be in line with the EnableObject() and DisableObject() actions. Added notes to these actions for when a more comprehensive update of the objectId assignment logic is revamped.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -1226,6 +1226,8 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             return true;
         }
 
+        //TODO: May need to track enabled/disabled objecst separately since some
+        //actions loop through the ObjectIdTOSimObjPhysics dict and this may have adverse effects
         public void DisableObject(string objectId) {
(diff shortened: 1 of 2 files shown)


mattdeitke (approved):
LGTM

ekolve:
It looks like DisableObjectsOfType is now in Base, but your PR doesn't remove it from the PhysicsRemoteFPSAgent..

mattdeitke:
Oh, that's true.

I didn't know it was previously in PhysicsRemoteFPSAgent.

@winthos, can you remove it from there?

https://github.com/allenai/ai2thor/blob/8a855f8e007767710b650981e1c33b2b0108916a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs#L7280

winthos:
Removed from the physics controller. should be good to go

mattdeitke:
👍 I'd wait for it to build successfully, then looks good to merge

PR #791, opened 2021-06-15
adding floor mesh colliders

winthos:
-adding floor mesh colliders to all generated sub floors, set to lits own layer
-adds floorCheck raycast against floor layer to ensure spawned objects only appear over parts of floor object with a mesh renderer


AlvaroHG:
Great!

PR #801, opened 2021-06-21
Better arm editor debugging and allowing relative arm movement.

Lucaweihs:
Now if you hold shift while in the debug editor window you can move the arm with the arrow keys and s/w. I also added an arm action for allowing relative movements (it's quite cumbersome to always have to specify absolute coordinates).

One "fix" I added that I'd like comments on: I noticed that when moving the arm, the position of the arm can become desynched with the position of the `IK_pos_rot_manipulator`. This is quite counter intuitive to me (as there is some invisible state that changes how functions behave) so I now ensure that the position of the `IK_pos_rot_manipulator` is always reset to the position of the arm after taking a `MoveArm` or a `MoveArmRelative` action.

### DIFF
--- a/unity/Assets/Scripts/ArmAgentController.cs
+++ b/unity/Assets/Scripts/ArmAgentController.cs
@@ -20,6 +20,25 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             return arm;
         }
 
+
+        /*
+        This function is identical to `MoveArm` except that rather than
+        giving a target position you instead give an "offset" w.r.t.
+        the arm's (i.e. wrist's) current location.
+
+        Thus if you want to increase the
+        arms x position (in world coordinates) by 0.1m you should
+        pass in `offset=Vector3(0.1f, 0f, 0f)` and `coordinateSpace="world"`.
+        If you wanted to move the arm 0.1m to the "right" from the agent's
+        perspective then you would pass in the same offset but set
+        `coordinateSpace="armBase"`. Note that this last movement is **not**
+        the same as passing `position=Vector3(0.1f, 0f, 0f)` to the `MoveArm`
+        action with `coordinateSpace="wrist"` as, if the wrist has been rotated,
+        right need not mean the same thing to the arm base as it does to the wrist.
+
+        Finally note that when `coordinateSpace="wrist"` then both `MoveArm` and
+        `MoveArmRelative` are identical.
+        */
         public void MoveArmRelative(
             Vector3 offset,
             float speed = 1,
--- a/unity/Assets/Scripts/ArmAgentController.cs
+++ b/unity/Assets/Scripts/ArmAgentController.cs
@@ -20,6 +20,28 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             return arm;
         }
 
+        public void MoveArmRelative(
+            Vector3 offset,
+            float speed = 1,
+            float? fixedDeltaTime = null,
+            bool returnToStart = true,
+            string coordinateSpace = "armBase",
+            bool restrictMovement = false,
+            bool disableRendering = true
+        ) {
+            IK_Robot_Arm_Controller arm = getArm();
+            arm.moveArmRelative(
+                controller: this,
+                offset: offset,
+                unitsPerSecond: speed,
+                fixedDeltaTime: fixedDeltaTime.GetValueOrDefault(Time.fixedDeltaTime),
+                returnToStart: returnToStart,
+                coordinateSpace: coordinateSpace,
+                restrictTargetPosition: restrictMovement,
+                disableRendering: disableRendering
+            );
+        }
+
         public void MoveArm(
             Vector3 position,
             float speed = 1,
--- a/unity/Assets/Scripts/DebugDiscreteAgentController.cs
+++ b/unity/Assets/Scripts/DebugDiscreteAgentController.cs
@@ -2,11 +2,16 @@
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.UI;
+using System;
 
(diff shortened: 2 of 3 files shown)


mattdeitke:
> I also added an arm action for allowing relative movements (it's quite cumbersome to always have to specify absolute coordinates).

Relative movements have already been supported. You just have to change the coordinate space. See: https://ai2thor.allenai.org/manipulathor/documentation/#MoveArm-coordinatespace.

elimvb:
The de-sync between the wrist position and its out-of-range pos_rot_mainpulator was a working-as-intended decision, but if re-syncing them after the fact makes more sense, I'm all for it.

Lucaweihs:
@mattdeitke From my understanding, changing the coordinate frame gets us part of the way there but has a couple of deficiencies: 

1. The only way to really move things in a relative (or offset) fashion is to pick the coordinate frame of the wrist. This works fine if the wrist isn't rotated but if I do rotate the wrist then I can't do relative movements w.r.t. to the base (which is the most natural type of relative moment from the agent's perspective in my mind).
2. What if I want to do relative movements but in the global coordinate frame? I don't think this is supported at all.

winthos:
> The only way to really move things in a relative (or offset) fashion is to pick the coordinate frame of the wrist. This works fine if the wrist isn't rotated but if I do rotate the wrist then I can't do relative movements w.r.t. to the base (which is the most natural type of relative moment from the agent's perspective in my mind).

@Lucaweihs I think you can just use the `armBase` coordinate space to make sure things are always relative to wherever the arm's base is. Perhaps this is enough and we don't need a full new action in order to accomplish this? Perhaps I'm misunderstanding what you mean by the "base" though, base of the wrist, base of the arm, the agent itself?

>What if I want to do relative movements but in the global coordinate frame? I don't think this is supported at all.
Using the `world` coordinate space is relative to a global coordinate frame. The world origin is always constant, so it will always be relative to that origin. Again, perhaps I'm misunderstanding what you are trying to do here.

Also @elimvb do you remember why we decided it was working as intended? I recall an early bug where the pos rot manipulator was desynced causing the final position of the wrist to not end up where it was expected, like it was offset by a couple of delta time increments. Do you recall if this is related to that?

Lucaweihs:
> @Lucaweihs I think you can just use the armBase coordinate space to make sure things are always relative to wherever the arm's base is. Perhaps this is enough and we don't need a full new action in order to accomplish this? Perhaps I'm misunderstanding what you mean by the "base" though, base of the wrist, base of the arm, the agent itself?

I don't believe that works, here's an example:
* Suppose that the arm is at position (1, 1, 1) in the armBase coordinate space.
* I now want to move the arm to the right by 0.1m in the armBase coordinate space.

How can I do this? If I use the `MoveArm` function with the armBase coordinate space then I have to explicitly tell it to move the arm to position (1.1, 1, 1) rather than telling it to move by (0.1, 0, 0). This is super unpleasant because I might not know the exact position of the arm in the armBase coordinate space (it's not being returned in the metadata to my understanding) and, even if the position of the arm was being returned in every conceivable coordinate space, it mean that every manipulathor user would have to do these computations themselves.

winthos:
Ah I see what you mean now. This sort of "move some amount relative to my current coordinate" does currently only work when in the wrist relative space, since it is the wrist that is moving so its coordinate space effectively moves with it. This sort of "move right of where I am currently by 0.1m" is currently required to be calculated on the user end when using other coordinate spaces.

I'm not sure if this should be a whole new action as you have added or if it should just be another parameter included in the default arm movement actions. I could see something like specifying `coordinateSpace=armBase` and then `moveRelativeToCurrentPos=True` or something like that. I think we can probably come up with a better name for this, but I'm currently leaning towards integrating this into the current `MoveArm` action since its already sort of supported when using `wrist` coordinate space.

elimvb:
> Also @elimvb do you remember why we decided it was working as intended? I recall an early bug where the pos rot manipulator was desynced causing the final position of the wrist to not end up where it was expected, like it was offset by a couple of delta time increments. Do you recall if this is related to that?

@winthos I don't recall an error like the one you're describing. The working-as-intended decision was because I wanted the pos_rot_manipulator's transform to work independently of the arm's range. So if the manipulator was put out of the agent's reach, the arm would hyperextend as much as it could to reach it. Thus, the wrist transform wouldn't be synced with manipulator in this case, but instead get as close as it could without breaking the arm. However, Luca made the valid point that this forces us to keep track of two pieces of end-joint metadata, which gets really confusing when you're doing follow-up transforms. So he added an extra step where, once the wrist gets as close as it can to the pos_rot_manipulator, the manipulator's work is complete for that step, and snaps over to the wrist's location so that they are both synced for the next transform. That seems like an intuitive compromise of logic to me, given our variables.

winthos:
@elimvb Gotchya. I agree, this change is a good compromise given the behavior that we are expecting.

Overall I think this looks good to go, but I do want @ehsanik to check to make sure none of the changes break anything on her end.

unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -20,6 +20,28 @@ public class ArmAgentController : PhysicsRemoteFPSAgentController {
             return arm;
         }
 
+        public void MoveArmRelative(
```
winthos, line 23: as mentioned in my review of #803, I'd like a quick comment summary for this action to make documentation upkeep easier.

winthos (approved):

unity/Assets/Scripts/ArmAgentController.cs
Lucaweihs: I have added this documentation.

ehsanik (approved):
This looks good to me and in fact, I have been implementing this action on the python side. It would definitely make life easier.

PR #803, opened 2021-06-23
Fixing arm collision bug and cleaning up move/rotation actions.

Lucaweihs:
This should only be merged after #801.

This PR:

* Fixes a bug where an objects held by the arm would not interact with other (non-kinematic) objects.
* Adds Move(Ahead/Back/Left/Right) and Rotate(Left/Right) actions to the arm controller.
* Removes the Move(Ahead/Back/Left/Right) and Rotate(Left/Right) actions form the base controller (these are always overwritten and so should not exist).
* Move the Move(Ahead/Back/Left/Right) and Rotate(Left/Right) to the action dispatch API.

### DIFF
--- a/unity/Assets/Scripts/ArmAgentController.cs
+++ b/unity/Assets/Scripts/ArmAgentController.cs
@@ -20,6 +20,28 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             return arm;
         }
 
+        public void MoveArmRelative(
+            Vector3 offset,
+            float speed = 1,
+            float? fixedDeltaTime = null,
+            bool returnToStart = true,
+            string coordinateSpace = "armBase",
+            bool restrictMovement = false,
+            bool disableRendering = true
+        ) {
+            IK_Robot_Arm_Controller arm = getArm();
+            arm.moveArmRelative(
+                controller: this,
+                offset: offset,
+                unitsPerSecond: speed,
+                fixedDeltaTime: fixedDeltaTime.GetValueOrDefault(Time.fixedDeltaTime),
+                returnToStart: returnToStart,
+                coordinateSpace: coordinateSpace,
+                restrictTargetPosition: restrictMovement,
+                disableRendering: disableRendering
+            );
+        }
+
         public void MoveArm(
             Vector3 position,
             float speed = 1,
--- a/unity/Assets/Scripts/ArmAgentController.cs
+++ b/unity/Assets/Scripts/ArmAgentController.cs
@@ -149,6 +149,19 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             }
         }
 
+        public override void MoveAhead(
+            float? moveMagnitude = null,
+            string objectId = "",
+            float maxAgentsDistance = -1f,
+            bool forceAction = false,
+            bool manualInteract = false,
+            bool allowAgentsToIntersect = false
+        ) {
+            // TODO: This is an ugly necessity until we refactor all
+            // of the agent controllers
+            MoveAhead(moveMagnitude: moveMagnitude, fixedDeltaTime: Time.fixedDeltaTime);
+        }
+
         public void MoveAhead(
             float? moveMagnitude = null,
             float speed = 1,
@@ -163,6 +176,19 @@ namespace UnityStandardAssets.Characters.FirstPerson {
                 returnToStart: returnToStart,
                 disableRendering: disableRendering
             );
+        }
+
+        public override void MoveBack(
+            float? moveMagnitude = null,
+            string objectId = "",
+            float maxAgentsDistance = -1f,
+            bool forceAction = false,
+            bool manualInteract = false,
+            bool allowAgentsToIntersect = false
+        ) {
+            // TODO: This is an ugly necessity until we refactor all
+            // of the agent controllers
+            MoveBack(moveMagnitude: moveMagnitude, fixedDeltaTime: Time.fixedDeltaTime);
         }
 
         public void MoveBack(
@@ -181,6 +207,19 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             );
         }
 
+        public override void MoveRight(
+            float? moveMagnitude = null,
+            string objectId = "",
+            float maxAgentsDistance = -1f,
+            bool forceAction = false,
+            bool manualInteract = false,
+            bool allowAgentsToIntersect = false
+        ) {
+            // TODO: This is an ugly necessity until we refactor all
+            // of the agent controllers
+            MoveRight(moveMagnitude: moveMagnitude, fixedDeltaTime: Time.fixedDeltaTime);
+        }
+
         public void MoveRight(
             float? moveMagnitude = null,
             float speed = 1,
@@ -195,6 +234,19 @@ namespace UnityStandardAssets.Characters.FirstPerson {
                 returnToStart: returnToStart,
                 disableRendering: disableRendering
             );
+        }
+
+        public override void MoveLeft(
+            float? moveMagnitude = null,
+            string objectId = "",
+            float maxAgentsDistance = -1f,
+            bool forceAction = false,
+            bool manualInteract = false,
+            bool allowAgentsToIntersect = false
+        ) {
+            // TODO: This is an ugly necessity until we refactor all
+            // of the agent controllers
+            MoveLeft(moveMagnitude: moveMagnitude, fixedDeltaTime: Time.fixedDeltaTime);
         }
 
         public void MoveLeft(
@@ -213,6 +265,16 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             );
         }
 
+        public override void RotateRight(
+            float? degrees = null,
+            bool manualInteract = false,
+            bool forceAction = false
+        ) {
+            // TODO: This is an ugly necessity until we refactor all
+            // of the agent controllers
+            RotateRight(degrees: degrees, fixedDeltaTime: Time.fixedDeltaTime);
+        }
+
         public void RotateRight(
             float? degrees = null,
             float speed = 1.0f,
@@ -229,6 +291,16 @@ namespace UnityStandardAssets.Characters.FirstPerson {
                 disableRendering: disableRendering,
                 fixedDeltaTime: fixedDeltaTime
             );
+        }
+
+        public override void RotateLeft(
+            float? degrees = null,
+            bool manualInteract = false,
+            bool forceAction = false
+        ) {
+            // TODO: This is an ugly necessity until we refactor all
+            // of the agent controllers
+            RotateLeft(degrees: degrees, fixedDeltaTime: Time.fixedDeltaTime);
         }
 
         public void RotateLeft(
--- a/unity/Assets/Scripts/ArmAgentController.cs
+++ b/unity/Assets/Scripts/ArmAgentController.cs
@@ -151,19 +151,11 @@ namespace UnityStandardAssets.Characters.FirstPerson {
 
         public override void MoveAhead(
             float? moveMagnitude = null,
-            string objectId = "",
-            float maxAgentsDistance = -1f,
-            bool forceAction = false,
-            bool manualInteract = false,
-            bool allowAgentsToIntersect = false
-        ) {
-            // TODO: This is an ugly necessity until we refactor all
-            // of the agent controllers
-            MoveAhead(moveMagnitude: moveMagnitude, fixedDeltaTime: Time.fixedDeltaTime);
-        }
-
-        public void MoveAhead(
-            float? moveMagnitude = null,
+            string objectId = "",                // TODO: Unused, remove when refactoring the controllers
+            float maxAgentsDistance = -1f,       // TODO: Unused, remove when refactoring the controllers
+            bool forceAction = false,            // TODO: Unused, remove when refactoring the controllers
+            bool manualInteract = false,         // TODO: Unused, remove when refactoring the controllers
+            bool allowAgentsToIntersect = false, // TODO: Unused, remove when refactoring the controllers
             float speed = 1,
             float? fixedDeltaTime = null,
             bool returnToStart = true,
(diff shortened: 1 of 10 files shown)


unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -127,37 +149,104 @@ public class ArmAgentController : PhysicsRemoteFPSAgentController {
...
 
-        public override void MoveBack(ServerAction action) {
-            throw new InvalidOperationException("When using the arm, please call controller.step(action=\"MoveAgent\", ahead=-X, right=0).");
+        public void MoveBack(
+            float? moveMagnitude = null,
+            float speed = 1,
+            float? fixedDeltaTime = null,
+            bool returnToStart = true,
+            bool disableRendering = true
+        ) {
```
mattdeitke, line 174: same as `MoveLeft`.

unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -127,37 +149,104 @@ public class ArmAgentController : PhysicsRemoteFPSAgentController {
...
-
-        public override void MoveLeft(ServerAction action) {
-            throw new InvalidOperationException("When using the arm, please call controller.step(action=\"MoveAgent\", ahead=0, right=-X).");
+        public void MoveAhead(
+            float? moveMagnitude = null,
+            float speed = 1,
+            float? fixedDeltaTime = null,
+            bool returnToStart = true,
+            bool disableRendering = true
+        ) {
```
mattdeitke, line 158: same as `MoveLeft`

mattdeitke (changes requested):
Still looking over it, but as an initial review, there are now a few places where calling `Move/Rotate` will now break, if the user passes in certain kwargs.

**Note:** ArmAgent inherits from PhysicsRemoteAgentController
https://github.com/allenai/ai2thor/blob/main/unity/Assets/Scripts/ArmAgentController.cs#L11

mattdeitke:
Actually, on second thought.. I think a better solution would be to have a new agent controller for `agentMode="default"` that also inherits from `PhysicsRemoteFPSAgentController`, but does not include have any overlap with the `agentMode="arm"` controller (e.g., `DefaultAgentController.cs`)

This will make things much easier to maintain (e.g., if somebody adds a parameter to `MoveAhead` in PhysicsRemote, they won't also have to "catch" it in the `AgentController`.

unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
```diff
@@ -441,6 +489,14 @@ public class IK_Robot_Arm_Controller : MonoBehaviour {
                 );
                 clone.isTrigger = true;
 
+                // The other set is used to interact with moveable objects
+                cols.Add(Instantiate(
+                    c,
+                    c.transform.position,
+                    c.transform.rotation,
+                    FourthJoint
+                ));
```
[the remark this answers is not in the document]
Lucaweihs: Done

unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -127,37 +149,104 @@ public class ArmAgentController : PhysicsRemoteFPSAgentController {
...
-        public override void Rotate(Vector3 rotation) {
-            throw new InvalidOperationException("When using the arm, please call controller.step(action=\"RotateAgent\", degrees=-X).");
+        public void RotateLeft(
+            float? degrees = null,
+            float speed = 1.0f,
+            bool waitForFixedUpdate = false,
+            bool returnToStart = true,
+            bool disableRendering = true,
+            float fixedDeltaTime = 0.02f
+        ) {
```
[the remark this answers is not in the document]
Lucaweihs: I now manually override it but instead call the new version of RotateLeft while ignoring the unnecessary parameters (e.g. `manualInteract`). This isn't ideal but I'm not sure how else we can avoid action ambiguity (the action dispatcher might pick the old version of `RotateLeft` if, for instance, only `degrees` was specified).

Lucaweihs:
> Actually, on second thought.. I think a better solution would be to have a new agent controller for agentMode="default" that also inherits from PhysicsRemoteFPSAgentController, but does not include have any overlap with the agentMode="arm" controller (e.g., DefaultAgentController.cs) 

@mattdeitke I started down this path but it quickly became a **huge** refactor that we should discuss with Eric (and everyone else) first. I completely agree that something like this should be done, especially as it's become really painful to figure our where certain actions are defined and there is a large amount of "action leakage" where certain actions are available to agents when they shouldn't be.

unity/Assets/Scripts/ArmAgentController.cs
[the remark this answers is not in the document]
winthos: I think this should be fine until we have a larger conversation about a bigger rework once Eric is back.

unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -20,6 +20,28 @@ public class ArmAgentController : PhysicsRemoteFPSAgentController {
             return arm;
         }
 
+        public void MoveArmRelative(
```
winthos, line 23: I'd like a brief header comment here as well as for each of the new user accessible actions in this PR. Details like what each parameter does, how it is different from the default `MoveArm` action would streamline updating the documentation.

unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -127,37 +149,176 @@ public class ArmAgentController : PhysicsRemoteFPSAgentController {
             }
         }
 
-        public override void MoveAhead(ServerAction action) {
-            throw new InvalidOperationException("When using the arm, please call controller.step(action=\"MoveAgent\", ahead=X, right=0).");
+        public override void MoveAhead(
```
winthos, line 152: Another note for all these move action updates, we need to make sure to update the ManipulaTHOR api to also include these on top of the default `MoveAgent`

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2525,6 +2488,45 @@ abstract public class BaseFPSAgentController : MonoBehaviour {
             }
         }
 
+        public void ParentObjects(string parentId, string childId) {
```
winthos, line 2491: Again, add a quick comment for the sake of documenting this action. I'm not sure if this particular one is something we want to keep internal for now or not though, I'll leave that up to you.

Also should this be `ParentObject` instead of `ParentObjects`?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2525,6 +2488,45 @@ abstract public class BaseFPSAgentController : MonoBehaviour {
...
+            SimObjPhysics parent = physicsSceneManager.ObjectIdToSimObjPhysics[parentId];
+            SimObjPhysics child = physicsSceneManager.ObjectIdToSimObjPhysics[childId];
+
+            child.gameObject.transform.parent = parent.gameObject.transform;
+            child.GetComponent<Rigidbody>().isKinematic = parent.GetComponent<Rigidbody>().isKinematic;
+
+            actionFinished(true);
+        }
+
+        public void UnparentObject(string objectId) {
```
winthos, line 2517: same as `ParentObjects`

winthos (changes requested):

unity/Assets/Scripts/ArmAgentController.cs
Lucaweihs: Added in #801.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Done

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Done

winthos (approved):
looks good

unity/Assets/Scripts/StochasticRemoteFPSAgentController.cs
```diff
@@ -238,44 +248,68 @@ public class StochasticRemoteFPSAgentController : BaseFPSAgentController {
             actionFinished(success: true);
         }
 
-        public override void MoveAhead(ServerAction action) {
-            action.x = 0.0f;
-            action.y = 0;
-            action.z = 1.0f;
-            MoveRelative(action);
+        public void MoveAhead(
```
ehsanik, line 251: I was under the impression that we made these changes somewhere at some point. @mattdeitke Did you ever incorporate similar changes somewhere? Do I recall correctly? If yes we should make sure we are consistent with those changes.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2525,6 +2488,69 @@ abstract public class BaseFPSAgentController : MonoBehaviour {
...
+
+        /*
+        Naively parents one sim object under another game object. The child object
+        will become kinematic, this is necessary as parenting non-kinematic rigidbodies
+        can cause all sorts of issues.
+
+        Note that this function is "dangerous" in that it can have unintended interactions
+        with other actions. Please only use this action if you understand what you're doing.
+        */
+        public void ParentObject(string parentId, string childId) {
```
ehsanik, line 2500: Can this function result in a child having multiple parents that are disconnected? Basically, currently we will only have multiple parents for a child if there is a parent-child relationship between the two parents. Is this a correct observation?

unity/Assets/Scripts/StochasticRemoteFPSAgentController.cs
mattdeitke: I didn't add `Move{Ahead/Left/Right/Back}` to the arm agent. I basically left a big TODO, which Luca is doing in this PR :)

ehsanik (approved):
I can confirm that these changes have fixed the bug Luca has mentioned. I tested it in this jupyter notebook: https://github.com/allenai/manipulathor/blob/bring_object/scripts/testing_lucas_changes.ipynb
Knocking kinematic objects in the room with the object in hand was not possible in the previous builds (it was only possible with the arm itself, which is why we missed this issue in my opinion). I verified that the bug is resolved on the notebook mentioned above.

unity/Assets/Scripts/ArmAgentController.cs
```diff
@@ -127,37 +149,176 @@ public class ArmAgentController : PhysicsRemoteFPSAgentController {
...
+            bool disableRendering = true
+        ) {
+            MoveAgent(
+                right: moveMagnitude.GetValueOrDefault(gridSize),
+                speed: speed,
+                fixedDeltaTime: fixedDeltaTime,
+                returnToStart: returnToStart,
+                disableRendering: disableRendering
+            );
         }
```
mattdeitke, line 237: (cc @erick) Won't this break the action dispatcher? Read:
https://github.com/allenai/ai2thor/blob/f49351084337547964732c00f8cc96af98182380/unity/Assets/Scripts/ActionDispatcher.cs#L46

My original idea was to just have
```c#
        public override void MoveRight(
            float? moveMagnitude = null,
            string objectId = "",
            float maxAgentsDistance = -1f,
            bool forceAction = false,
            bool manualInteract = false,
            bool allowAgentsToIntersect = false
        ) {
            ...
        }
```
and basically ignore `allowAgentsToIntersect`.

But it seems dangerous to have an ambiguous action with:
```c#
        public override void MoveRight(
            float? moveMagnitude = null,
            string objectId = "",
            float maxAgentsDistance = -1f,
            bool forceAction = false,
            bool manualInteract = false
        ) {
            ...
        }

        public override void MoveRight(
            float? moveMagnitude = null,
            string objectId = "",
            float maxAgentsDistance = -1f,
            bool forceAction = false,
            bool manualInteract = false,
            bool allowAgentsToIntersect = false
        ) {
            ...
        }
```

unity/Assets/Scripts/DroneFPSAgentController.cs
```diff
@@ -61,22 +61,24 @@ public class DroneFPSAgentController : BaseFPSAgentController {
 #endif
         }
 
-        public override void RotateRight(ServerAction action) {
+        public void RotateRight(ServerAction action) {
```
mattdeitke, line 64: Is the drone supposed to be able to `Move{Ahead/Right/Left/Back}`? I thought @winthos mentioned yes to this a long time ago.

Removing such actions from base would mean they need to be added again here.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -1793,61 +1810,102 @@ public class PhysicsRemoteFPSAgentController : BaseFPSAgentController {
...
+            float maxAgentsDistance = -1f,
+            bool forceAction = false,
+            bool manualInteract = false,
+            bool allowAgentsToIntersect = false
+        ) {
+            if (!moveMagnitude.HasValue) {
+                moveMagnitude = gridSize;
+            } else if (moveMagnitude < 0f) {
+                throw new InvalidOperationException("moveMagnitude but be > 0");
+            }
```
mattdeitke, line 1849: Should an invalid operation exception occur when `moveMagnitude == 0` (i.e., not moving the agent)?

unity/Assets/Scripts/StochasticRemoteFPSAgentController.cs
```diff
@@ -65,37 +65,47 @@ public class StochasticRemoteFPSAgentController : BaseFPSAgentController {
...
             }
-            var moveLocal = new Vector3(action.x, 0, action.z);
-            var moveMagnitude = moveLocal.magnitude;
-            if (moveMagnitude > 0.00001) {
-                // random.NextGaussian(RotateGaussianMu, RotateGaussianSigma);
-                var random = new System.Random();
+
+            var moveLocal = new Vector3(x, 0, z);
+            float xzMag = moveLocal.magnitude;
+            if (xzMag > 0.00001) {
```
mattdeitke, line 88: Perhaps `0.00001` should have an `f` at the end for float?

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: The object hierarchy in Unity is a tree so every object can only ever have a single parent. All this function really does is ensure that two objects will move together.

unity/Assets/Scripts/DroneFPSAgentController.cs
Lucaweihs: Good point: I've added them in.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
Lucaweihs: Good point, done.

unity/Assets/Scripts/StochasticRemoteFPSAgentController.cs
Lucaweihs: Technically not my code but fair enough, I changed it to 1e-5.

unity/Assets/Scripts/ArmAgentController.cs
Lucaweihs: Ok I've made a change suggested by @ekolve: creating single functions that contain all parameters from the default and arm agent controllers. This is a bit ugly but it is unambiguous and we can can remove parameters when we eventually do a refactor.

unity/Assets/Scripts/StochasticRemoteFPSAgentController.cs
mattdeitke: That too, but `1e-5` still makes it a double instead of a float. I added a small change in https://github.com/allenai/ai2thor/pull/803/commits/f78edae25668549627726f3e32cc6d5f116e070a, which simply makes `1e-5` into `1e-5f`.

mattdeitke (approved):
LGTM

PR #805, opened 2021-06-25
Allowing for rotating the wrist & elbow in a relative fashion.

Lucaweihs:
This should only be merged after #803. Note that I've targeted this merge at the `arm-collision-test` branch so that it's easier to see the differences, it should actually be merged to main after approval.

This PR:

* Replaces the `RotateMidLevelHand` method with a `RotateWristRelative` action. As it sounds, this allows for rotating the arm's wrist given an input pitch, yaw, and roll. I've hooked up this action into the debug editor, try it out by using `shift+alt+(arrow keys and w/s)` (I think it's fairly intuitive).
* Adds `RotateElbow` and `RotateElbowRelative` commands to allow for rotating the arm's elbow. Try this with `shift+alt+(q/e)`
* Enables a special "teleportation" case in arm movement/rotation: when taking an arm action with `speed` set to `float.PositiveInfinity` and `fixedDeltaTime` is set to `0`, the action completes instantly ignoring physics. This is similar to the `Teleport` function we have for agent moment.

Note that I haven't tried to fix any existing self-collision issues, I don't think these are too problematic in general.

### DIFF
--- a/unity/Assets/Scripts/ArmAgentController.cs
+++ b/unity/Assets/Scripts/ArmAgentController.cs
@@ -385,6 +385,60 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             );
         }
 
+        /*
+        Rotates the elbow (in a relative fashion) by some given
+        number of degrees. Easiest to see how this works by
+        using the editor debugging and shift+alt+(q/e).
+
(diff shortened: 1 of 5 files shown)


winthos (approved):
looks good to go once the other pre-req PRs are in

PR #808, opened 2021-06-30
fix: disp_screen is an object

guhur:
This was generating an error:

```TypeError: 'Display' object is not subscriptable```

It was wrongly used. The line 123 (for example) shows the correct usage.


ekolve:
Great catch! Thank you for the PR.

ekolve (approved):

Issue #812, opened 2021-07-07
ToggleMapView with fixed map orientation


xubo92:
Hi there,

I was trying to use ToggleMapView to get a top-down view. it works but with a minor problem. When I rotate the agent, from the top-down view, the agent is actually not rotated but the map rotates. this is kinda counter-intuitive. Is there any way to have a agent rotate and move but the map keep fixed?

Thanks!

mattdeitke:
> When I rotate the agent, from the top-down view, the agent is actually not rotated but the map rotates. this is kinda counter-intuitive.

Yes, this is why its not documented. I'm going to open up a PR which solves this problem using a third party camera. Keeping open for now.

Issue #445, opened 2020-08-14
Support 'MapFrame' Officially


mattdeitke:
**Expected functionality.**

Currently, this can be achieved from Python with
```python
c = Controller()
agent_view = c.last_event.frame
event = c.step('ToggleMapView')
map_view = event.frame
c.step('ToggleMapView')  # goes back to agent view with actions available.
```

But, It should work with:
```python
c = Controller(renderMapFrame=True)
event = c.step('MoveAhead')
map_view = event.map_frame
agent_view = event.frame
```

It does not have to be super optimized, since this is purely for visualization purposes. Just a wrapper from the Unity side that makes it easier.

---

I can run some tests to confirm that ToggleMapView works with all scenes, but I believe this issue was fixed with the video.

Lucaweihs:
I have some python code that automatically generates functions which produce a 1-1 correspondence between pixel coordinates (in the map view) to (x,z) in THOR, is this something that we would want to include here? For instance, something like

```python
x, z = event.map_coord_to_thor_xz(row, col)
assert (row, col) == event.thor_xz_to_map_coord(x, z)
```

mattdeitke:
Yeah, that seems quite useful from the Python side! I wouldn't return all the x/z positions in the metadata. The best place for it would probably be within Python `Controller`.

thomason-jesse:
> I have some python code that automatically generates functions which produce a 1-1 correspondence between pixel coordinates (in the map view) to (x,z) in THOR, is this something that we would want to include here? For instance, something like
> 
> ```python
> x, z = event.map_coord_to_thor_xz(row, col)
> assert (row, col) == event.thor_xz_to_map_coord(x, z)
> ```

I would love to see this code. I'm trying to do this mapping using `SceneBounds` but with particularly stretched/elongated maps there's a lot of drift between the topdown view and the corresponding (x, z) coords.

Lucaweihs:
Hi @thomason-jesse,

Check out this [get_agent_map_data](https://github.com/allenai/cordial-sync/blob/master/utils/visualization_utils.py#L109) function as well as this [visualize_agent_path](https://github.com/allenai/cordial-sync/blob/master/utils/visualization_utils.py#L206) function. These were made with an older version of THOR in mind but I believe they should work now still. Let me know if you have any issues.

keeper1314:
我在主函数中加入代码：为什么不能改变agent的位置？
if __name__ == "__main__":
    import matplotlib.pyplot as plt
    
[...]

PR #814, opened 2021-07-08
Top-Down Image Frame Updates

mattdeitke:
Now allows third party cameras to be used in place of ToggleMapView with

```python
event = controller.step(action="GetMapViewCameraProperties")
controller.step(
    action="AddThirdPartyCamera",
    **event.metadata["actionReturn"]
)
```
I don't think these should be merged into a single action, because, eventually, Third Party Cameras should be deprecated in favor of having the camera actually be an agent.

Resolves https://github.com/allenai/ai2thor/issues/812.
Fixes https://github.com/allenai/ai2thor/issues/445.
Deprecates `ToggleMapView`.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -3218,7 +3218,10 @@ namespace UnityStandardAssets.Characters.FirstPerson {
 
             if (structureObjs != null) {
                 foreach (StructureObject structure in structureObjs) {
-                    if (structure.WhatIsMyStructureObjectTag == StructureObjectTag.Ceiling) {
+                    if (
+                        structure.WhatIsMyStructureObjectTag == StructureObjectTag.Ceiling
+                        && structure.gameObject.name == "Ceiling"
+                    ) {
                         ceiling = structure;
                         break;
                     }
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -3154,6 +3154,7 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             actionFinished(true);
         }
 
+        [ObsoleteAttribute(message: "This action is deprecated. Use GetMapViewCameraProperties with a third party camera instead.", error: false)]
         public void ToggleMapView() {
             SyncTransform[] syncInChildren;
 
@@ -3164,8 +3165,6 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             foreach (StructureObject structure in structureObjs) {
                 switch (structure.WhatIsMyStructureObjectTag) {
                     case StructureObjectTag.Ceiling:
-                        ceiling = structure;
-                        goto case StructureObjectTag.LightFixture;
                     case StructureObjectTag.LightFixture:
                     case StructureObjectTag.CeilingLight:
                         structureObjsList.Add(structure);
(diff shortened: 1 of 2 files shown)


mattdeitke:
Here is a notebook sharing results: https://colab.research.google.com/drive/1GSIF78B62hNskyr-SYtUMgMdsZ7tJpJQ?usp=sharing

winthos (commented):
The new action looks good to go.

I think the only thing is, if we are going to completely deprecate `ToggleMapView`, we need to be explicit in the documentation that if you do want this new 3rd party, top down camera to follow the agent as the old `ToggleMapView` camera did, they will need to update the spawned 3rd party camera since it won't automatically follow the agent position and orientation.

I can think of use cases where you would want that old behavior as an option, so perhaps deprecating it entirely is a bit too hasty?

mattdeitke:
> I can think of use cases where you would want that old behavior as an option

Can you provide some? I can't think of any reason one would want to move the top-down camera and the agent at the same time, especially when it means the images from the top-down camera will no longer capture the entire scene.

winthos (approved):

mattdeitke:
Now ready to merge.
See https://colab.research.google.com/drive/1GSIF78B62hNskyr-SYtUMgMdsZ7tJpJQ?usp=sharing for all results.

ekolve (approved):

Lucaweihs (approved):
I'm happy with this change, LGTM.

mattdeitke:
Pushed a fix for FloorPlan309, where the camera was actually being placed below the scene. This was because the Ceiling in this room (and in none of the others) actually extends below the floor.
image

xubo92:
Hi @mattdeitke 

Thanks for this new feature!

Since I was using ai2thor==3.1.0 (on Windows), how can I merge this new feature and make it work in ai2thor-3.1.0?

I tried to `git clone ai2thor==3.1.0` and manually change the code files needed and `pip install -e .`, but that does not work and report error "ValueError: Invalid action: GetMapViewCameraProperties". 

Thanks!

xubo92:
Just a follow-up, since I'm on Windows, does that mean I need to rebuild everything from ai2thor/unity folder in Unity to make this new feature work? The new feature is related to the change of the controller, so I guess the answer is yes?

Thanks!

mattdeitke:
Hey @xubo92,

Yes, if you rebuild locally from the latest commit on the `main` branch, it will all work, even with version 3.1.0 of the Python API.

xubo92:
Hi @mattdeitke 

I tried this new feature and found out it only works for single agent. When I set `agentCount` more than 1, then the `third_party_camera_frames` is empty. Is this normal? Can it by any chance support multi-agent case? 

Thanks!

xubo92:
And a follow-up question for this feature: how can I enable this top-down view all the time for real-time visualization while running controller for many steps? 

I tried this code https://colab.research.google.com/drive/1GSIF78B62hNskyr-SYtUMgMdsZ7tJpJQ?usp=sharing#scrollTo=CvZ2PkYp95U2 , it works good when I use plt to show one image. But when I only run my controller and want to visualize in a top-down view constantly, it does not work and still using first-person view.

Thanks!

mattdeitke:
> I tried this new feature and found out it only works for single agent. When I set agentCount more than 1, then the third_party_camera_frames is empty. Is this normal? Can it by any chance support multi-agent case?

Here is a demo of it working in the multi-agent case: https://colab.research.google.com/drive/1C_-9zqAqhSUcZvrxVlwlTMZQzWnEzY9k?usp=sharing.

One needs to also pass in `agentId` and then look at the event that corresponds to that id.

> how can I enable this top-down view all the time for real-time visualization while running controller for many steps?

How are you visualizing the frames?

For the single agent case, check out `plot_frames` in [ai2thor-colab](https://github.com/allenai/ai2thor-colab/blob/main/ai2thor_colab/__init__.py#L131). It doesn't yet work with the MultiAgentEvent, but I can probably extend it to if that'd be useful.

image

xubo92:
Hi @mattdeitke 

Thanks a lot for these examples!

Probably I'm not saying clearly. Here is a code snippet I use

` // initialize a controller
self.controller = Controller(agentMode=self.agent_types[0], agentCount=2, scene=self.scene,  width=1600, height=900, local_executable_path=local_executable_path)  
// setup top-down view
event = self.controller.step(action="GetMapViewCameraProperties")
event = self.controller.step(action="AddThirdPartyCamera", agentId=0, **event.metadata["actionReturn"])
// after setting top-down view, for each move, I have a unity windows pop out showing what happened in the scene (multiple agents moving), in a top-down view
for i in range(100):
    act = np.random.choice(ACTS)
    self.controller.step(action=act, agentId=0)
    self.controller.step(action=act, agentId=1)`

Basically, I'm running a RL algorithm using ai2thor with multiple agents moving around. I expect to visualize the simulated scenes continuously, in a top-down view, so that I can have a sense of the policy training process by keeping an eye on the movement of agents. I expect the top-down view can give me a view of two (or) more agents in a room (not like the first-person view that I have to choose the agentId and get its local view.)

Not sure how to achieve this.

Thanks!

xubo92:
I remember when use the old command "ToggleMapView", it works like a switch, and one only needs to call it somewhere once (initialization) and after that the default view is switched to top-down view. I thought the new feature works the same way, especially when one has multiple agents so that top-down view can give a whole scene containing every agent. 

Sorry I have a lot of questions here...

xubo92:
Hi @mattdeitke 

Thank you so much for clarifying the difference! Now I understand what is happening.

Saving frames and create a video sounds good! I'll try it.

But still curious, is there any efficient way that you know to have a second window showing the corresponding top-down view  as the Unity window updates, using something like matplotlib, or drawing a canvas and update it continuously? I've tried opencv (cv2.imshow), and it seems slow and cannot catch up with the Unity update rate.

Thanks!

yaofeng11:
Hi@mattdeitke
I wonder if it possible to achieve that when I control the robot walking in the room. And the same time,I have the second window to show the total scene of the room (for observe the motion of the robot)

mattdeitke:
Hi @yaofeng11,

There isn't currently a way to view the images from 2 separate cameras in separate Unity windows. But, what might be a workaround is you can do is save images or videos from the agent's perspective and from the top-down perspective, and then view them afterwards.

I just put this notebook demo together on that, which you may find useful:
https://colab.research.google.com/drive/1p6A_LDZA853-6boLdo4ha_1Bg6Cy3Jj9?usp=sharing

Let me know if that helps,
Matt :)

yaofeng11:
Dear Mr Matt:
&nbsp;Thanks for your help. It bothered me for a whole week. Now I know how to solve it, and I will try it. Much appreciate.
Yaofeng :)







------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "allenai/ai2thor"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年1月13日(星期四) 下午3:07
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [allenai/ai2thor] Top-Down Image Frame Updates (#814)





 
Hi @yaofeng11,
 
There isn't currently a way to view the images from 2 separate cameras in separate Unity windows. But, what might be a workaround is you can do is save images or videos from the agent's perspective and from the top-down perspective, and then view them afterwards.
 
I just put this notebook demo together on that, which you may find useful:
 https://colab.research.google.com/drive/1p6A_LDZA853-6boLdo4ha_1Bg6Cy3Jj9?usp=sharing
 
Let me know if that helps,
 Matt :)
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #816, opened 2021-07-14
ThirdPartyCamera New Action and Bug Fixes

winthos:
~~- adds new action `CoordinateFromRaycastThirdPartyCamera` that allows getting coordinates of objects hit from a ThirdPartyCamera raycast (same as GetCoordinateFromRaycast for the main agent camera)~~

- updates some DebugInput field syntax to not use outdated `ActionDispatcher` syntax, instead use new `ProcessControlCommand` overload

- removes extra `AddComponent` call in `AddThirdPartyCamera` since camera prefabs now already have a camera component on them

~~- unit tests for ThirdPartyCamera instantiation as well as new CoordinateFromRaycastThirdPartyCamera~~

Edit: All functions related to `CoordinateFromRaycastThirdPartyCamera` are now isolated to the [`CoordinateFramRaycastTPC`](https://github.com/allenai/ai2thor/tree/CoordinateFromRaycastTPC) branch to make future updates to ThirdPartyCamera into a "ghost agent" cleaner.

### DIFF
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -508,7 +508,6 @@ public class AgentManager : MonoBehaviour {
 
         GameObject gameObject = GameObject.Instantiate(Resources.Load("ThirdPartyCameraTemplate")) as GameObject;
         gameObject.name = "ThirdPartyCamera" + thirdPartyCameras.Count;
-        //gameObject.AddComponent(typeof(Camera));
         Camera camera = gameObject.GetComponentInChildren<Camera>();
 
         // set up returned image
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -567,39 +567,6 @@ public class AgentManager : MonoBehaviour {
         );
     }
 
-    //same as `GetCoordinateFromRaycast` for the main agent camera, but used with any ThirdPartyCamera in scene
-    public void CoordinateFromRaycastThirdPartyCamera(float x, float y, int thirdPartyCameraId = 0) {
-        // count is out of bounds
-        if (thirdPartyCameraId >= thirdPartyCameras.Count || thirdPartyCameraId < 0) {
-            throw new ArgumentOutOfRangeException(
-                $"thirdPartyCameraId: {thirdPartyCameraId} (int: default=0) must in 0 <= thirdPartyCameraId < len(thirdPartyCameras)={thirdPartyCameras.Count}."
-            );
-        }
-
-        if (x < 0 || y < 0 || x > 1 || y > 1) {
-            throw new ArgumentOutOfRangeException($"x and y must be in [0:1] not (x={x}, y={y}).");
-        }
-
-        Camera thirdPartyCamera = thirdPartyCameras[thirdPartyCameraId];
-
-        Ray ray = thirdPartyCamera.ViewportPointToRay(new Vector3(x, 1 - y, 0));
-        RaycastHit hit;
-
-        if (Physics.Raycast(
-            ray: ray,
-            hitInfo: out hit,
-            maxDistance: Mathf.Infinity,
-            layerMask: LayerMask.GetMask("Default", "Agent", "SimObjVisible", "PlaceableSurface"),
-            queryTriggerInteraction: QueryTriggerInteraction.Ignore)) {
-            this.activeAgent().actionFinished(success: true, actionReturn: hit.point);
-
-        } else
-            this.activeAgent().actionFinished(success: true, actionReturn: null);
-
-
-
-    }
-
     // Here, we don't want some dimensions set. For instance, set x, but not y.
     public class OptionalVector3 {
         public float? x = null;
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -41,6 +41,7 @@ public class AgentManager : MonoBehaviour {
     private bool renderNormalsImage;
     private bool renderFlowImage;
     private Socket sock = null;
+    [SerializeField]
     private List<Camera> thirdPartyCameras = new List<Camera>();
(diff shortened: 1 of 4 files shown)


mattdeitke:
Haven't reviewed, but an aside: This is another reason third party cameras should be deprecated in favor of a new ghost agent, so actions like this don't have to exist (as the ghost agent would already have access to the base's `GetCoordinatesFromRaycast`.

winthos:
> Haven't reviewed, but an aside: This is another reason third party cameras should be deprecated in favor of a new ghost agent, so actions like this don't have to exist (as the ghost agent would already have access to the base's `GetCoordinatesFromRaycast`).

agreed, this is a quick fix for now since some team members require this functionality asap.

mattdeitke:
> agreed, this is a quick fix for now since some team members require this functionality asap.

For now, should `CoordinateFromRaycastThirdPartyCamera` just exist on a standalone branch that is never merged into `main`? The `main` branch would still be able to merge into the standalone branch it in the future. But, it would avoid putting this oddly specific function, which will be deprecated soon. @Lucaweihs has done similar things in previous projects, where instead of merging something into main, it just exists on a standalone branch, and it seems fine.

From Python, one can still easily use it as a separate branch with
```python
controller = Controller(
    branch=branch_name
)
```

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -508,7 +508,7 @@ private enum serverTypes { WSGI, FIFO };
 
         GameObject gameObject = GameObject.Instantiate(Resources.Load("ThirdPartyCameraTemplate")) as GameObject;
         gameObject.name = "ThirdPartyCamera" + thirdPartyCameras.Count;
-        gameObject.AddComponent(typeof(Camera));
+        //gameObject.AddComponent(typeof(Camera));
```
Lucaweihs, line 511: Good point :)! Maybe just delete this line?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -568,6 +568,34 @@ Vector3 defaultsOnNull
...
+        RaycastHit hit;
+        Physics.Raycast(
+            ray: ray,
+            hitInfo: out hit,
+            maxDistance: Mathf.Infinity,
+            layerMask: LayerMask.GetMask("Default", "Agent", "SimObjVisible", "PlaceableSurface"),
+            queryTriggerInteraction: QueryTriggerInteraction.Ignore
+        );
+
+        this.activeAgent().actionFinished(success: true, actionReturn: hit.point);
```
Lucaweihs, line 596: Remind me: what happens if the raycast doesn't hit anything? Is `hit.point` a bunch of infinities?

Lucaweihs (approved):
LGTM. I generally agree with Matt that small 1-function updates can be kept to their own branches when they're not widely useful but this PR has some other updates which I'd liked to see merged (e.g. tests, removing unnecessary lines, etc).

unity/Assets/Scripts/AgentManager.cs
winthos: if it doesn't hit anything, the actual `Physics.Raycast` will return false, and the `RaycastHit hit` won't be updated, so `hit.point` will be at its default (0, 0, 0).

unity/Assets/Scripts/AgentManager.cs
winthos: will do!

unity/Assets/Scripts/AgentManager.cs
Lucaweihs: In that case can we maybe return `null` rather than `hit.point`?

winthos: yeah I think this would be safer, I'll add this in

unity/Assets/UnitTests/TestThirdPartyCamera.cs
```diff
@@ -0,0 +1,106 @@
...
+            bool result = false;
+            Vector3 coord = (Vector3)actionReturn;
+
+            result = Mathf.Approximately(coord.x, 0.0f);
+            Assert.AreEqual(result, true);
+
+            result = Mathf.Approximately(coord.y, 0.0f);
+            Assert.AreEqual(result, true);
+
+            result = Mathf.Approximately(coord.z, 2.472f);
```
AlvaroHG, line 91: What is it hitting here? how did you get then number?
This is why I think in the future we should also have tests use simple procedural scenes for tests since it's not obvious this should be the behavior from reading the test.

AlvaroHG:
LGTM

unity/Assets/UnitTests/TestThirdPartyCamera.cs
winthos: yeah this value is based on spawning a camera into FloorPlan_1, so if FloorPlan_1 changes it could invalidate this test. Some sort of procedural test scene should be the way to go in the future, but currently the tests default initializing and loading up FloorPlan_1. I got the value from a direct debug log of the return of the action. We currently use similar values obtained directly from the console in things like the `PutObject` test.

mattdeitke (approved):
LGTM.

Left a small comment on initializing dictionaries, but its equivalent to what is currently there, and is fine to merge regardless :)

unity/Assets/Scripts/DebugInputField.cs
```diff
@@ -1195,12 +1195,27 @@ public class DebugInputField : MonoBehaviour {
...
-                            orthographic: true,
-                            orthographicSize: 5
-                        );
+                        Dictionary<string, object> action = new Dictionary<string, object>();
+
+                        action["action"] = "AddThirdPartyCamera";
+                        action["position"] = Vector3.zero;
+                        action["rotation"] = Vector3.zero;
+                        action["orthographic"] = true;
+                        action["orthographicSize"] = 5;
```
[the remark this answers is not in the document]
winthos: updated!

Issue #818, opened 2021-07-17
Missing object meta after InitialRandomSpawn


BoyuanChen:
Hi team,

I found that for certain environment and certain random seed, after InitialRandomSpawn action, there will be objects missing from the returned object metadata. One example to reproduce this issue is attached. I tested it on both 3.3.1 and 3.3.4.

After running the following script, the two print statements will print 38 and 37 respectively. The `credit card` object is missing.

Could you please help take a look at this? Thanks!!

```python

import ai2thor.controller

scene = 'FloorPlan214_physics'
controller = ai2thor.controller.Controller(
    width=300,
    height=300,
    local_build=False,
    start_unity=True,
    scene=scene,
    port=8201,
    host='127.0.0.1',
    agentMode='default',
    )
event = controller.reset(scene)
object_meta = event.metadata['objects']
print(len(object_meta))


controller.reset(scene)
shuffled_event = controller.step(
    action='InitialRandomSpawn',
    randomSeed=17,
    forceVisible=False,
    numPlacementAttempts=5,
    placeStationary=True
    )

object_meta = shuffled_event.metadata['objects']
print(len(object_meta))
```

mattdeitke:
Thanks @BoyuanChen!

I ran into this issue a long time ago with the rearrangement project, but was never able to reproduce it at the time. I can reproduce this, though!

Colab link: https://colab.research.google.com/drive/16QJjtq6p9Dje9_UGabLlLCiFhmQyG5-s?usp=sharing

BoyuanChen:
Great! Thanks @mattdeitke !

Another side note that may be helpful is that I only encountered this when setting forceVisible=False. Out of 120 scenes, there were about 11 scenes that have this issue. However, I do not think this is scene-specific since changing random seed may or may not have this issue again.

mattdeitke:
This is interesting to note on `forceVisible`.

Do you by chance have the 11 scenes? I can just run them forever with a bunch of random seeds in the background to just see if they're able to reproduce this issue.

I've previously just run `InitialRandomSpawn` 50 times, with different random seeds, in every scene to see if I was able to reproduce the issue, but that never resulted finding any cases. Although, I might not have been using `forceVisible=False`.

Right now, we think the issue might have to do with an object having an incorrect bounding box annotation, rather than an issue with the `InitialRandomSpawn` action itself. So, it might be a bit of a manual updating process with a few outlying objects.

Lucaweihs:
[...]

PR #821, opened 2021-07-19
Object Bounds Fix

winthos:
- adds a line to ensure the child game object used for `BoundingBox` for all sim objects correctly has its local position zeroed out. 

- forces any objects that are about to break to drop an child objects contained by them. This fixes issues of some objects disappearing on `InitialRandomSpawn` when certain receptacles break. Previously, if something like an Apple was placed inside a breakable plate via `InitialRandomSpawn` with `placeStationary = True` it would not be "dropped" by the receptacle and was deactivated a the plate breaks.

- adds new `MakeAllObjectsUnbreakable` and `MakeAllObjectsBreakable` actions to be used with `InitialRandomSpawn`

- adds a check to not reposition receptacles that have had another object spawn inside of them during the `InitialRandomSpawn` process. The current logic only checks if the new position of a receptacle has enough free space for the receptacle only, not any objects also "contained" by the receptacle. One cause of objects breaking were sequences like: 

```
Saltshaker spawns inside Plate,

Plate (+ Saltshaker) spawns inside Microwave since Plate had enough room to fit in microwave, 

Saltshaker causes plate to be shoved down into the bottom of the microwave since the saltshaker is now clipping with the top of the microwave, 

plate breaks (which also previously deleted the Saltshaker since it wasn't decoupled from the plate)
```

- The above change means fewer objects will be shuffled around, which is a major (albeit temporary) change in the random spawn logic. This will be fixed with the full `InitialRandomSpawn` rework to come.

The new way to ensure no objects break on `InitialRandomSpawn` is to do something like the following following scene initialization:

```python
    event = controller.step(  
        action="MakeAllObjectsUnbreakable"
    )

    event = controller.step(
        action="InitialRandomSpawn",
        randomSeed=seed,
        forceVisible=False,
        numPlacementAttempts=5,
        placeStationary=True
    )

    event = controller.step(  
    action="MakeAllObjectsBreakable"
    )
```

Note that these actions before and after `InitialRandomSpawn` can't currently be integrated into `InitialRandomSpawn` itself. I tried automatically calling these but some sort of physics resolution external to `InitialRandomSpawn` is causing issues. I believe in order to integrate this automatically, it will require a rewrite of the entire `InitialRandomSpawn` sequence which is being done in a future update anyway.

Additionally [here is a colab demonstrating the fix](https://colab.research.google.com/drive/1nJCa5lym3f7mcMzcq7VWC01MsUHlKdVe?usp=sharing).
These changes fix issues in regards to #818

### DIFF
--- a/unity/Assets/Scripts/InstantiatePrefabTest.cs
+++ b/unity/Assets/Scripts/InstantiatePrefabTest.cs
@@ -271,6 +271,35 @@ public class InstantiatePrefabTest : MonoBehaviour {
         // get the bounding box of the sim object we are trying to place
         BoxCollider oabb = sop.BoundingBox.GetComponent<BoxCollider>();
 
+        Vector3 oabb_original_center = new Vector3();
+        Vector3 oabb_original_size = new Vector3();
+
+        //store current bounds dimensions: center, size
+        oabb_original_center = oabb.center;
+        oabb_original_size = oabb.size;
+
+        if (sop.IsReceptacle) {
+            //make sure to enable BoundingBox boxcollider component otherwise BOUNDS IS ZERO
+            oabb.enabled = true;
+            Bounds b = oabb.bounds;
+
+            foreach (GameObject c in sop.ContainedGameObjects()) {
+
+                //grow the oabb to encapuslate any contained object colliders so check spawn area can account for more space occupied
+                BoxCollider cBB = c.GetComponent<SimObjPhysics>().BoundingBox.GetComponent<BoxCollider>();
+                cBB.enabled = true;
+                b.Encapsulate(cBB.bounds);
+                cBB.enabled = false;
+            }
+
+            oabb.center = b.center;
+            oabb.size = b.size;
+
+            //reset this to disabled, will enable later in once rotations start to be checked
+            oabb.enabled = false;
+
+        }
+
         // zero out rotation and velocity/angular velocity, then match the target receptacle's rotation
         sop.transform.rotation = rsp.ReceptacleBox.transform.rotation;
         Rigidbody sopRB = sop.GetComponent<Rigidbody>();
@@ -363,6 +392,10 @@ public class InstantiatePrefabTest : MonoBehaviour {
                 // translate position of the target sim object to the rsp.Point and offset in local y up
                 sop.transform.position = rsp.Point + rsp.ReceptacleBox.transform.up * (quat.distance + yoffset);// rsp.Point + sop.transform.up * DistanceFromBottomOfBoxToTransform;
                 sop.transform.rotation = quat.rotation;
+
+                //also reset the object's BoundingBox values at this point
+                oabb.center = oabb_original_center;
+                oabb.size = oabb_original_size;
 
                 // now to do a check to make sure the sim object is contained within the Receptacle box, and doesn't have
                 // bits of it hanging out
@@ -480,8 +513,9 @@ public class InstantiatePrefabTest : MonoBehaviour {
             }
         }
 
-        // reset rotation if no valid spawns found
-        // oh now we couldn't spawn it, all the spawn areas were not clear
+        // reset rotation, position, and bounding box original center and size if not able to be placed
+        oabb.center = oabb_original_center;
+        oabb.size = oabb_original_size;
         sop.transform.rotation = originalRot;
         sop.transform.position = originalPos;
         return false;
--- a/unity/Assets/Scripts/InstantiatePrefabTest.cs
+++ b/unity/Assets/Scripts/InstantiatePrefabTest.cs
@@ -177,10 +177,6 @@ public class InstantiatePrefabTest : MonoBehaviour {
         // try a number of spawnpoints in this specific receptacle up to the maxPlacementAttempts
         int tries = 0;
         foreach (ReceptacleSpawnPoint p in goodRsps) {
-            if(sop.Type == SimObjType.Plate)
-            {
-                print($"trying to place {sop.name} in {p.ParentSimObjPhys}");
-            }
             if (PlaceObject(sop, p, PlaceStationary, degreeIncrement, AlwaysPlaceUpright)) {
                 return true;
             }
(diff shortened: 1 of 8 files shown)


mattdeitke (approved):
LGTM

unity/Assets/Scripts/InstantiatePrefabTest.cs
```diff
@@ -359,7 +359,7 @@ bool AlwaysPlaceUpright
         foreach (RotationAndDistanceValues quat in ToCheck) {
             // if spawn area is clear, spawn it and return true that we spawned it
             if (CheckSpawnArea(sop, rsp.Point + rsp.ParentSimObjPhys.transform.up * (quat.distance + yoffset), quat.rotation, false)) {
-
+                
                 // translate position of the target sim object to the rsp.Point and offset in local y up
                 sop.transform.position = rsp.Point + rsp.ReceptacleBox.transform.up * (quat.distance + yoffset);// rsp.Point + sop.transform.up * DistanceFromBottomOfBoxToTransform;
                 sop.transform.rotation = quat.rotation;
```
Lucaweihs, line 365: Is it possible we need a
```c#
            if (!Physics.autoSyncTransforms) {
                Physics.SyncTransforms();
            }
```
given that we've updated the transform?

unity/Assets/Scripts/InstantiatePrefabTest.cs
```diff
@@ -304,7 +304,7 @@ bool AlwaysPlaceUpright
 
                 Vector3 Offset = oabb.ClosestPoint(oabb.transform.TransformPoint(oabb.center) + -rsp.ReceptacleBox.transform.up * 10); // was using rsp.point
                 BoxBottom = new Plane(rsp.ReceptacleBox.transform.up, Offset);
-                DistanceFromBoxBottomTosop = BoxBottom.GetDistanceToPoint(sop.transform.position);
+                DistanceFromBoxBottomTosop = Math.Abs(BoxBottom.GetDistanceToPoint(sop.transform.position));
```
Lucaweihs, line 307: I haven't read this super closely but are we sure we want the absolute value here? Just glancing at how this distance is used below it looks like it might need to be negative in some cases.

unity/Assets/Scripts/InstantiatePrefabTest.cs
```diff
@@ -433,6 +433,11 @@ bool AlwaysPlaceUpright
                 // set true if we want objects to be stationary when placed. (if placed on uneven surface, object remains stationary)
                 // if false, once placed the object will resolve with physics (if placed on uneven surface object might slide or roll)
                 if (PlaceStationary == true) {
+                    // if the target receptacle is a pickupable receptacle, set it to kinematic true as will sence we are placing stationary
```
Lucaweihs, line 436: "as will sence we are placing stationary" ?

unity/Assets/Scripts/InstantiatePrefabTest.cs
```diff
@@ -433,6 +433,11 @@ bool AlwaysPlaceUpright
                 // set true if we want objects to be stationary when placed. (if placed on uneven surface, object remains stationary)
                 // if false, once placed the object will resolve with physics (if placed on uneven surface object might slide or roll)
                 if (PlaceStationary == true) {
```
Lucaweihs, line 435: Can't this just be `if (PlaceStationary)`?

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -3960,10 +3960,39 @@ public struct WhatDidITouch {
...
+                    if (sop.DoesThisObjectHaveThisSecondaryProperty(SimObjSecondaryProperty.CanBreak)) {
+                        sop.GetComponentInChildren<Break>().Unbreakable = true;
+                    }
+                }
+            }
+
+            actionFinished(true);
+        }
+
+        public void MakeAllObjectsBreakable() {
```
Lucaweihs, line 3990: `MakeAllObjectsUnbreakable` and `MakeAllObjectsUnbreakable` have basically the same logic. Can we instead have a `UpdateBreakabilityOfAllObjects` function that takes a `breakable` boolean as input?

unity/Assets/Scripts/PhysicsSceneManager.cs
```diff
@@ -173,6 +173,16 @@ public class PhysicsSceneManager : MonoBehaviour {
...
+    // public void MakeBreakableAndSleep() {
+    //     foreach (SimObjPhysics sop in GameObject.FindObjectsOfType<SimObjPhysics>()) {
+    //         if(sop.IsBreakable) {
+    //             sop.GetComponent<Rigidbody>().Sleep();
+    //             sop.GetComponentInChildren<Break>().Unbreakable = false;
+    //         }
+
+    //     }
+    // }
+
```
Lucaweihs, line 185: Is this necessary?

unity/Assets/Scripts/PhysicsSceneManager.cs
```diff
@@ -567,6 +585,11 @@ List<SimObjType> excludedReceptacleTypes
                 foreach (SimObjPhysics receptacleSop in IterShuffleSimObjPhysicsDictList(objTypeToReceptacles, rng)) {
                     List<ReceptacleSpawnPoint> targetReceptacleSpawnPoints;
 
+                    if(receptacleSop.ContainedGameObjects().Count > 0) {
+                        //this object already has something in it, skip over it since we currently can't account for detecting bounds of a receptacle + any contained objects
+                        continue;
+                    }
```
Lucaweihs, line 591: Does this mean we won't place more than 1 object on, e.g., a table?

Lucaweihs (commented):
A few comments but looks otherwise good.

unity/Assets/Scripts/InstantiatePrefabTest.cs
winthos: yes, I'll add this in to be safe.

unity/Assets/Scripts/InstantiatePrefabTest.cs
winthos: I think so. This whole section is quite messy, but basically we are trying to get the distance from the bottommost point of the object bounding box to the actual object's transform.position, and use this as an offset in the `y` direction so that the object is placed "on top" of a receptacle. The "on top" is relative to whatever the receptacle's local `y` direction is, so the absolute value should still be fine even if a receptacle was upside down, for example.

unity/Assets/Scripts/InstantiatePrefabTest.cs
winthos: yup, will tidy this up

unity/Assets/Scripts/PhysicsSceneManager.cs
winthos: perhaps... I will double check this. The intent is to only filter out objects that are pickupable that also already contain some object. Those shouldn't be moved. So things like a countertop or a table should not be skipped over, but things like a plate or pot should.

unity/Assets/Scripts/PhysicsSceneManager.cs
winthos: no, this was leftover from a previous attempt at a fix. I'll remove it.

unity/Assets/Scripts/InstantiatePrefabTest.cs
Lucaweihs: Gotcha, sounds good!

unity/Assets/Scripts/PhysicsSceneManager.cs
Lucaweihs: Once you've checked this, feel free to merge from my perspective.

ekolve (approved):
One request to remove a debug statement, otherwise lgtm.

unity/Assets/Scripts/PhysicsSceneManager.cs
winthos: ok, i've updated this to only skip pickupable objects, and things are working as expected.

unity/Assets/UnitTests/TestPutObject.cs
```diff
@@ -49,11 +49,13 @@ public class TestPutObject : TestBase {
 
             GameObject creditCard = GameObject.Find("CreditCard_acee2f3e");
 
+            Debug.Log(creditCard.transform.position.x);
```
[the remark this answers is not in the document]
winthos: removed

PR #823, opened 2021-07-20
exposed unity random seed to python api surface

sagadre:
Main change is on L3057 in BaseFPSAgentController.cs


mattdeitke:
To add to some context, I much prefer this solution, of setting the random seed globally, rather than every random action including a random seed, which then sets the seed inside of the action.

So, instead of:
```python
controller.step(action="RandomizeMaterials", randomSeed=42)
controller.step(action="RandomizeMaterials", randomSeed=42)
```

you have

```python
controller.step(action="SetRandomSeed", seed=42)
controller.step(action="RandomizeMaterials")
controller.step(action="RandomizeMaterials")
```

This is much more consistent with how one expects to use a random seed in other libraries. For instance, in NumPy, one doesn't use:
```python
np.random.randn((125, 125), random_seed=42)
np.random.randn((125, 125), random_seed=45)
np.random.randn((125, 125), random_seed=46)
```
You instead set the seed once globally and keep sampling from it.

mattdeitke (approved):
LGTM.

Added a test for it, and visually tested in Colab: https://colab.research.google.com/drive/1QjMJrlXKS-b_o1N2MMbcf_0oTGdaaYCi?usp=sharing

sagadre:
Thanks Matt, I like your suggestion!

In the interim, one thing to note is the following sequences of actions will set the random seed twice, once for each call instead of just once.
```python
controller.step(action="SetRandomSeed", seed=42)
controller.step(action="InitialRandomSpawn") # default randomSeed=0, so the previous line didn't really do anything
```
However, this case should not really cause issues for the end user (behavior should still be reproducible even though the seed switches).

sagadre:
Another thing to consider is adding a random seed param to the `Controller` constructor and to `reset` so someone could set the seed from there and then not worry about having to call `SetRandomSeed`. Would still be useful to have this function for some use cases.

mattdeitke:
> Another thing to consider is adding a random seed param to the Controller constructor and to reset so someone could set the seed from there and then not worry about having to call SetRandomSeed. Would still be useful to have this function for some use cases.

I tend to prefer against having a seed in reset/initialize. Some reasons:
1. Supporting 2 ways to do something (i.e., setting a random seed as an action or upon initialization) is a bit dangerous. If one reads the documentation, and sees that there's a `seed` parameter upon initialize, it doesn't seem likely that one would also expect it to be a standalone action, and vice-versa. The idea of having 2 ways to do something is also unpythonic.
2. One common use case that can only be supported with the action is the ability to call reset right before some random action, so as to reset the seed when that action is reached. For instance:

```python
import random

controller.step(action="SetRandomSeed", seed=42)
controller.step(action="RandomizeMaterials")

# --- some code that may utilize unity's random function, but the scene has not reset ---
random_n_steps = random.choice(range(5, 20))
for _ in range(random_n_steps):
     # assume this is a stochastic movement that randomly samples moveMagnitude in Unity
    controller.step(action="MoveAhead")

controller.step(action="SetRandomSeed", seed=42)
controller.step(action="RandomizeLighting")
```
If one only had `SetRandomSeed` in `reset`, then one would not be able to guarantee that the random numbers generated for the `RandomizeLighting` action were reproducible, since the seed set upon `reset` would be sampled from an nondeterministic amount of times.

3. Initialization is super large right now. Between all agent types, it is probably the hardest thing to maintain because it is so messy with so many existing parameters.
4. Using `actionFinishedEmit` is extremely fast since it can largely cache the previously returned `Event` (without having to pass back new object metadata or image frames). Thus, keeping `SetRandomSeed` as a standalone action incurs a largely negligible performance cost.

sagadre:
oops this case could be an issue with the current codebase
```python
controller.step(action="SetRandomSeed", seed=42)
e1 = controller.step(action="InitialRandomSpawn") # default randomSeed=0, so the previous line didn't really do anything
controller.step(action="SetRandomSeed", seed=41)
e2 = controller.step(action="InitialRandomSpawn")
```
A user might expect `e1 != e2` but this is not the case. Would not be an issue for `RandomizeMaterials` etc. tho.

mattdeitke:
I think this is more of a fault of `InitialRandomSpawn`, which is getting deprecated soon, having a `randomSeed` parameter, rather than it being a fault of the current setup.

ekolve (approved):
I am fine with the change with one caveat that we need to ensure that we users are aware of that the order that actions are executed in will change the result if the actions invoke the RNG.  For example,

```
SetRandomSeed
RandomizeMaterials
RandomizeMaterials
```

vs

```
SetRandomSeed
RandomizeMaterials
SomeOtherRandomAction
RandomizeMaterials
```

The latter will result in a different set of random materials.  This behavior is not different from say using Python's random and explicitly setting the seed.  There is also the issue of an action on the Unity side indirectly using the RNG through some other API that we call, which will affect downstream calls to the RNG.  This only matters when someone is say generating a dataset and then adds an additional call to an action within a sequence of other actions that use the RNG.

ai2thor/tests/test_unity.py
```diff
@@ -1722,6 +1712,24 @@ def test_unsupported_manipulathor(controller):
...
+    f2_1 = controller.reset().frame
+    f2_2 = controller.step(action="SetRandomSeed", seed=42).frame
+    f2_3 = controller.step(action="RandomizeMaterials").frame
+
+    assert_images_near(f1_1, f2_1)
+    assert_images_near(f1_1, f1_2)
+    assert_images_near(f2_1, f2_2)
+    assert_images_near(f1_3, f2_3)
+    assert_images_far(f2_1, f2_3)
+    assert_images_far(f1_1, f1_3)
```
Lucaweihs, line 1730: This test will pass if `RandomizeMaterials` always does the same randomization ignoring the seed. Could you add a test to make sure 
```python
controller.step(action="SetRandomSeed", seed=41).frame
```
is different from `controller.step(action="SetRandomSeed", seed=42).frame`? I would suggest doing something like
```python
    f2_1 = controller.reset().frame
    controller.step(action="SetRandomSeed", seed=41).frame
    f2_2 = controller.step(action="RandomizeMaterials").frame
    f2_3 = controller.step(action="SetRandomSeed", seed=42).frame
    f2_4 = controller.step(action="RandomizeMaterials").frame
```
and then aserting taht `f2_2` is different is appropriately. I'd also suggest giving more informative names than `fi_j` as this test is hard to read.
as this also now tests that changing the seed

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -3054,6 +3054,12 @@ bool includeInvisible
             AgentHand.transform.rotation = this.transform.rotation;
         }
 
+        // set random seed used by unity
+        public void SetRandomSeed(int seed) {
+            UnityEngine.Random.InitState(seed);
+            actionFinishedEmit(true);
+        }
```
Lucaweihs, line 3061: What about the `System.Random` seed?

Lucaweihs (changes requested):
A few comments but I'm otherwise happy with this update.

ai2thor/tests/test_unity.py
mattdeitke: added https://github.com/allenai/ai2thor/pull/823/commits/540e161aa9f23ed5de827c40f06b8fd63727f65b

unity/Assets/Scripts/BaseFPSAgentController.cs
mattdeitke: I've added a new `systemRandom` variable to BaseFPSAgentController, which is now used in place of always constructing a new `System.Random()` object.

Some notes:
1. There was no way to set the seed of `System.Random` globally.
2. `System.Random` is necessary for generating doubles, instead of just floats (which can be done with `UnityEngine.Random`).
3. It is intentionally not used with a few actions that have `randomSeed` as a parameter. The only public one is InitialRandomSpawn, which is being reworked, such that, among other things, it will no longer have a random seed parameter.
4. After calling `controller.reset()`, there are no guarantees that the seeds remain in tact.

Lucaweihs (approved):
LGTM.

ekolve (approved):

PR #824, opened 2021-07-20
Temperature Decay API Updates

mattdeitke:
Wanting to update the temperature API to make it more consistent with the rest of AI2-THOR before releasing the Colab questions.

Breaks backwards and sideways compatibility on the following actions, which have never been used as far I can tell, from some searches on GitHub:
- `SetRoomTempDecayTimeForType` is replaced with `SetTemperatureDecayTime(float objectId, float decayTime)`. Instead of using an objectType, it uses an `objectId` for consistency with everything else.
- `SetGlobalRoomTempDecayTime` is replaced with `SetTemperatureDecayTime(float decayTime)`
- `SetDecayTemperatureBool(bool allowDecayTemperature)` is replaced and broken down into `DisableTemperatureDecay` and `EnableTemperatureDecay`.

For this rarely used object metadata key:
- Changes `canChangeTempToHot` to be `isHeatSource`
- Changes `canChangeTempToCold` to be `isColdSource`

These names were positively misleading, as they don't mean an object can't be changed to hot, but it means an object is a heat/cold source, where, when another object comes in contact with it and it is on, it changes that other object to be hot/cold.

Consider in FloorPlan10, that these are the only objects that report `canChangeTempToHot` in the metadata:

```
StoveBurner|+00.84|+00.92|-01.10
StoveBurner|+01.08|+00.92|-01.50
StoveBurner|+00.84|+00.92|-01.50
StoveBurner|+01.08|+00.92|-01.10
Toaster|+00.98|+00.90|+00.33
CoffeeMachine|+00.89|+00.90|-02.13
Microwave|+01.04|+01.68|-01.30
```

which didn't include things that could actually go inside of these receptacles.

### DIFF
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -62,44 +62,6 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             return ItemInHand;
         }
 
-        // get all sim objects of action.type, then sets their temperature decay timers to value
-        [ObsoleteAttribute(message: "This action is deprecated. Use SetTemperatureDecayTime instead.", error: false)]
-        public void SetRoomTempDecayTimeForType(string objectType, float TimeUntilRoomTemp = 0.0f) {
-            // get all objects of type passed by action
-            SimObjPhysics[] simObjects = GameObject.FindObjectsOfType<SimObjPhysics>();
-            List<SimObjPhysics> simObjectsOfType = new List<SimObjPhysics>();
-            foreach (SimObjPhysics sop in simObjects) {
-                if (sop.Type.ToString() == objectType) {
-                    simObjectsOfType.Add(sop);
-                }
-            }
-            foreach (SimObjPhysics sop in simObjectsOfType) {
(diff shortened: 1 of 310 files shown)


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -62,36 +62,39 @@ public class PhysicsRemoteFPSAgentController : BaseFPSAgentController {
...
             SimObjPhysics[] simObjects = GameObject.FindObjectsOfType<SimObjPhysics>();
-
-            // use SetHowManySecondsUntilRoomTemp to set them all
             foreach (SimObjPhysics sop in simObjects) {
-                sop.SetHowManySecondsUntilRoomTemp(TimeUntilRoomTemp);
+                sop.SetHowManySecondsUntilRoomTemp(decayTime);
             }
-
             actionFinished(true);
         }
```
mattdeitke, line 99: This is basically everything, in terms of new actions being added.

The rest of it is a find and replace with `canChangeTempToHot` to `isHeatSource` and `canChangeTempToCold` to `isColdSource`.

ekolve (approved):

winthos (approved):

PR #826, opened 2021-07-20
Held object yaw rotation fix

mattdeitke:
Starting from:
image

Fixes positive yaw to rotate clockwise:
image

Instead of counterclockwise (what is currently there):
image

Missed this in #770, presumably because I might only using testing it with symmetric objects about the z axis.


mattdeitke:
Actually, it was just an error with my input from the debug editor. This is already working.

PR #829, opened 2021-07-23
Parameterizing height/width in `ai2thor-xorg start`

Lucaweihs:
Allowing for changing the height/width of the x display when running `ai2thor-xorg start` by using the `--height` and `--width` flags.

### DIFF
--- a/scripts/ai2thor-xorg
+++ b/scripts/ai2thor-xorg
@@ -11,10 +11,14 @@ import signal
 
 # Turning off automatic black formatting for this script as it breaks quotes.
 # fmt: off
-
+from typing import List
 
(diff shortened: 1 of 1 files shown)


mattdeitke:
A related question here, are things slower if height and width are increased dramatically - e.g.,
```python
DEFAULT_HEIGHT = 2500
DEFAULT_WIDTH = 2500
```
And is there any good reason for the defaults being fairly low? It seems many people have run into this issue, where the frames don't render fully because the height/width is too small.

ekolve (approved):

Lucaweihs:
I haven't had any chance to experiment with what the FPS impact of changing the display is, I would guess it's minimal but until we get some concrete numbers I'll keep it low for now.

PR #832, opened 2021-07-28
Rename `ObjectTemperature` to `temperature`

mattdeitke:
### Overview

This is a backwards breaking change (in line with https://github.com/allenai/ai2thor/pull/824) in the object metadata that renames the `ObjectTemperature` key to `temperature`.

### Motivation

The capitalization of the `ObjectTemperature` metadata key is inconsistent with all the other metadata keys (i.e., should be `camelCase`).

### Old references

- ObjectTemperature was referenced a few times in the documentation (see: https://github.com/allenai/ai2thor.allenai.org/search?q=ObjectTemperature). However, these were mostly in the description of temperature decay descriptions.
- However, on the [2.0 release note](https://github.com/allenai/ai2thor/blob/main/doc/static/ReleaseNotes/ReleaseNotes_2.0.md#state-changes-added-to-object-metadata), it was actually incorrectly documented, saying that the key was `objectTemperature` instead of `ObjectTemperature`.
- As far as I can tell, it has never been used.

### DIFF
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -1387,7 +1387,7 @@ public class ObjectMetadata {
                          //
                          // temperature placeholder values, might get more specific later with degrees but for now just track these three states
     public enum Temperature { RoomTemp, Hot, Cold };
-    public string ObjectTemperature;// return current abstracted temperature of object as a string (RoomTemp, Hot, Cold)
+    public string temperature;// return current abstracted temperature of object as a string (RoomTemp, Hot, Cold)
                                     //
     public bool isHeatSource;// can change other object temp to hot
     public bool isColdSource;// can change other object temp to cool
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -1355,69 +1355,121 @@ public class ObjectMetadata {
     public string name;
     public Vector3 position;
     public Vector3 rotation;
+
     // public float cameraHorizon; moved to AgentMetadata, objects don't have a camerahorizon
     public bool visible;
-    public bool obstructed; // if true, object is obstructed by something and actions cannot be performed on it. This means an object behind glass will be obstructed=True and visible=True
+
+    // If true, object is obstructed by something and actions cannot be performed on it.
+    // This means an object behind glass will be obstructed=True and visible=True
+    public bool obstructed;
+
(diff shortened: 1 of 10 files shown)


unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1387,7 +1387,7 @@ public class ObjectMetadata {
                          //
                          // temperature placeholder values, might get more specific later with degrees but for now just track these three states
     public enum Temperature { RoomTemp, Hot, Cold };
-    public string ObjectTemperature;// return current abstracted temperature of object as a string (RoomTemp, Hot, Cold)
+    public string temperature;// return current abstracted temperature of object as a string (RoomTemp, Hot, Cold)
```
ekolve, line 1390: Could we just have a single temperature Enum? This is ServerAction so it doesn't matter a ton.  Alternatively, you could use this as an opportunity to change all the actions that depend on `Temperature` and `temperature` to the ActionDispatcher style.

ekolve (changes requested):

unity/Assets/Scripts/AgentManager.cs
mattdeitke: This was in `ObjectMetadata`, not `ServerAction`, but the same problem still applied, where one shouldn't have both `temperature` and `Temperature`. Although, unexpectedly, `Temperature` was never being returned on the python side (presumably because it was an enum).

In https://github.com/allenai/ai2thor/pull/832/commits/ee94dffa5f2e0762d9ee7f4959e920642d07534c, I moved the two existing enums out of `ObjectMetadata` for consistency, so they don't unexpectedly appear in the metadata if enums are serialized at some point.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -1355,69 +1355,121 @@ public class ObjectMetadata {
...
-    ///
     public string[] receptacleObjectIds;
-    public float distance;// dintance fromm object's transform to agent transform
-    public String objectType;
+
+    // distance from object's transform to agent transform
+    public float distance;
+
+    // what type of object is this?
+    public string objectType;
```
mattdeitke, line 1463: Also note: I've changed `String` to `string`. There didn't seem to be any reason the `String` class was being used here.

Most of the other changes just move the comments above the fields so the lines are super long.

winthos (approved):
LGTM after the metadata enum change

ekolve (approved):

PR #852, opened 2021-08-19
Refactor Agents into plain C# classes from Monobehaviors

ekolve:
This is a moderate size refactoring, so I apologize in advance for the large number of changes.  This PR removes Monobehavior from the FPSAgentcontroller inheritance tree.  The motivation behind this change is to simplify creating new agents and splitting existing agents into new agents.  

In this PR, the components on the FPSControllerPrefab: PhysicsRemoteFPSAgentController, StochasticRemoteFPSAgentController, ArmAgentController and DroneFPSAgentController are removed.  In their place, a new component named BaseAgentComponent has been added.

Each of the aforementioned agents now wrap the BaseAgentComponent in order to access properties such as `TallVisCap` or `BotVisCap`.  Essentially, what is being done is the BaseAgentComponent acts as the interface to all of the Unity specific components required by the agent (Hand, Arm, VisibilityCapsule, etc.).  All the actual behavior of the agents gets defined within non-monobehavior classes that have as an instance variable a reference to this component.

One benefit to this approach is that adding a new property within the editor is greatly simplified.  Instead of having to add to each of the distinct components (Physics, Arm, Stochastic, Drone), you simply need to add it to the BaseAgentComponent.

Note the following two screenshots:

Screen Shot 2021-08-13 at 11 59 21 AM

Screen Shot 2021-08-19 at 1 21 46 PM

The fields that are configured within the editor has been reduced to only the fields that point to other GameObjects.

The other benefit to this approach is that creating a new agent simply involves subclasses the BaseFPSAgentController then adding any actions you want in the new agent.  You will still need to handle when the new agent is constructed in the AgentManager (I will have a simpler way to manager that soon).

One final benefit is that in many of the agents we have performed initialization in the Start() callback.  This generally works fine, but the main issue with it is that its not predictable when this will get called and typically a frame must pass in order for it to be triggered.  Much of the initialization that lived in Start() methods has been moved to the agent's constructor where it will get executed immediately.


unity/Assets/Scripts/SimObjPhysics.cs
```diff
@@ -881,6 +876,7 @@ public class PhysicsMaterialValues {
             Debug.LogError(this.name + " is not at uniform scale! Set scale to (1, 1, 1)!!!");
         }
 #endif
+        agentManager = GameObject.Find("PhysicsSceneManager").GetComponentInChildren<AgentManager>();
```
Lucaweihs, line 879: It seems weird for me for sim obj physics to know about agent managers. Also possibly slow for every object to have to do a find.

Lucaweihs (commented):
Added some comments here and there but definitely like the changes/direction, feels much cleaner.

AlvaroHG:
LGTM, Good initial refactor.

Issue #498, opened 2020-10-12
How to generate a video 


laopao2002:
Hi,

I find that there is no document or tutorial which shows how to generate a video. I can get the frames after each event and then merge to a video, but it's choppy. Or I can use a screen recording software to record a video from the renderer, but the actions (such as move ahead, rotate and pick up something etc) perform suddenly. 

Is there another way to generate a video? I find a file called "video controller.py", but when I run the following code, some errors occur:

from ai2thor.controller import VideoController
with VideoController() as vc:
    vc.play(vc.MoveAhead())
    vc.wait(5)
    vc.play(vc.MoveAhead())
    vc.exportVideo('thor.mp4')

And if I want to get a smooth video, what should I do? 

Thank you!

mattdeitke:
For the video controller, can you try the following:
```python
with VideoController() as vc:
    vc.play(vc.MoveAhead())
    vc.wait(5)
    vc.play(vc.MoveAhead())
    vc.exportVideo('thor.mp4')
```
Note that it only works best with navigation actions.

mattdeitke:
Hey, what error did you get from running the code?

laopao2002:
Hey, the error was:

<ipython-input-6-58218b3734d1> in <module>
----> 1 with VideoController() as vc:
      2     vc.play(vc.MoveAhead())
      3     vc.wait(5)
      4     vc.play(vc.MoveAhead())
      5     vc.exportVideo('thor.mp4')

<ipython-input-5-7101101bff0c> in __init__(self, cam_rot, cam_pos, cam_fov, **controller_kwargs)
     13                  cam_fov=60,
     14                  **controller_kwargs):
---> 15         super().__init__(continuous=True, **controller_kwargs)
     16         self.step(
     17             action='AddThirdPartyCamera',

~/anaconda3/lib/python3.8/site-packages/ai2thor/controller.py in __init__(self, quality, fullscreen, headless, port, start_unity, local_executable_path, width, height, x_display, host, scene, image_dir, save_image_per_frame, docker_enabled, depth_format, add_depth_noise, download_only, include_private_scenes, server_class, **unity_initialization_parameters)
    445                 )
    446 
--> 447             event = self.reset(scene)
    448             if event.metadata['lastActionSuccess']:
    449                 init_return = event.metadata['actionReturn']

<ipython-input-5-7101101bff0c> in reset(self, scene)
     31         self.step(
     32             action='AddThirdPartyCamera',
[...]

PR #856, opened 2021-08-23
Bugfixes to make VideoController work that were probably from changes in the main ai2thor Controller definition. Works as of 2021-08-23 under ai2thor 3.3.5

d-val:
This is a bug fix to issue https://github.com/allenai/ai2thor/issues/498#issuecomment-893693197

### DIFF
--- a/ai2thor/video_controller.py
+++ b/ai2thor/video_controller.py
@@ -18,7 +18,6 @@ from PIL import Image
 import math
 from math import erf, sqrt
 
-
 class VideoController(Controller):
     def __init__(
         self,
@@ -27,28 +26,18 @@ class VideoController(Controller):
         cam_fov=60,
         **controller_kwargs,
     ):
-        super().__init__(continuous=True, **controller_kwargs)
-        self.step(
-            action="AddThirdPartyCamera",
-            rotation=initial_camera_rotation,
-            position=initial_camera_position,
-            fieldOfView=initial_camera_fov,
-        )
-
         self.saved_frames = []
         self.ceiling_off = False
         self.initial_cam_rot = cam_rot.copy()
         self.initial_cam_pos = cam_pos.copy()
         self.initial_cam_fov = cam_fov
 
-    def reset(self, scene=None, **init_params):
-        """Changes the scene and adds a new third party camera to the initial position."""
-        super().reset(scene, **init_params)
+        super().__init__(continuous=True, **controller_kwargs)
         self.step(
             action="AddThirdPartyCamera",
-            rotation=self.initial_cam_rot,
-            position=self.initial_cam_pos,
-            fieldOfView=self.initial_cam_fov,
+            rotation=cam_rot.copy(),
+            position=cam_pos.copy(),
+            fieldOfView=cam_fov,
         )
 
     def play(self, *action_generators):
@@ -386,3 +375,30 @@ class VideoController(Controller):
                 self.saved_frames.append(rgb)
             success, image = vidcap.read()
             i += 1
+
+    def exportVideo(self, path):
+        """Merges all the saved frames into a .mp4 video and saves it to `path`"""
+        if self.saved_frames:
+            path = path if path[:-4] == '.mp4' else path + '.mp4'
+            if os.path.exists(path):
+                os.remove(path)
+            print((self.saved_frames[0].shape[1], self.saved_frames[0].shape[0]))
+            video = cv2.VideoWriter(
+                path,
+                cv2.VideoWriter_fourcc(*'DIVX'),
+                30,
+                (self.saved_frames[0].shape[1], self.saved_frames[0].shape[0])
+            )
+            for i, frame in enumerate(self.saved_frames):
+                print('|', end='')
+                # assumes that the frames are RGB images. CV2 uses BGR.
+                video.write(cv2.cvtColor(frame, cv2.COLOR_BGR2RGB))
+            cv2.destroyAllWindows()
+            video.release()
+            print('done')
+
+with VideoController() as vc:
+    vc.play(vc.MoveAhead())
+    vc.wait(5)
+    vc.play(vc.MoveAhead())
+    vc.exportVideo('thor.mp4')
--- a/ai2thor/video_controller.py
+++ b/ai2thor/video_controller.py
@@ -18,6 +18,7 @@ from PIL import Image
 import math
 from math import erf, sqrt
 
+
 class VideoController(Controller):
     def __init__(
         self,


ai2thor/video_controller.py
```diff
@@ -27,28 +27,18 @@ def __init__(
...
-        super().reset(scene, **init_params)
+        super().__init__(continuous=True, **controller_kwargs)
         self.step(
             action="AddThirdPartyCamera",
-            rotation=self.initial_cam_rot,
-            position=self.initial_cam_pos,
-            fieldOfView=self.initial_cam_fov,
+            rotation=cam_rot.copy(),
+            position=cam_pos.copy(),
+            fieldOfView=cam_fov,
```
mattdeitke, line 41: This is a nice catch with the bug fix for making a copy of the dicts!

ai2thor/video_controller.py
```diff
@@ -386,3 +376,30 @@ def merge_video(self, other_video_path):
...
+                30,
+                (self.saved_frames[0].shape[1], self.saved_frames[0].shape[0])
+            )
+            for i, frame in enumerate(self.saved_frames):
+                print('|', end='')
+                # assumes that the frames are RGB images. CV2 uses BGR.
+                video.write(cv2.cvtColor(frame, cv2.COLOR_BGR2RGB))
+            cv2.destroyAllWindows()
+            video.release()
+            print('done')
```
mattdeitke, line 399: It looks like `export_video` already exists on the VideoController (See: https://github.com/allenai/ai2thor/blob/main/ai2thor/video_controller.py#L349-L365), and this method appears to be a duplicate of that with some extra print statements? 

Can this be removed?

mattdeitke (changes requested):
Thanks for the PR @d-val, and nice catch removing `initial_camera_rotation`, `initial_camera_position`, and `initial_camera_fov`!

Mostly looks good, though I left a few small comments to address before merging.

ai2thor/video_controller.py
```diff
@@ -27,28 +27,18 @@ def __init__(
...
-            rotation=initial_camera_rotation,
-            position=initial_camera_position,
-            fieldOfView=initial_camera_fov,
-        )
-
         self.saved_frames = []
         self.ceiling_off = False
         self.initial_cam_rot = cam_rot.copy()
         self.initial_cam_pos = cam_pos.copy()
         self.initial_cam_fov = cam_fov
```
mattdeitke, line 34: I've reproduced the bug. To fix it, I think all that need to be changed is:

```python
        self.step(
            action="AddThirdPartyCamera",
            rotation=initial_camera_rotation,
            position=initial_camera_position,
            fieldOfView=initial_camera_fov,
        )

        self.saved_frames = []
        self.ceiling_off = False
        self.initial_cam_rot = cam_rot.copy()
        self.initial_cam_pos = cam_pos.copy()
        self.initial_cam_fov = cam_fov
```
becomes
```python
        self.saved_frames = []
        self.ceiling_off = False
        self.initial_cam_rot = cam_rot.copy()
        self.initial_cam_pos = cam_pos.copy()
        self.initial_cam_fov = cam_fov

        self.step(
            action="AddThirdPartyCamera",
            rotation=self.initial_cam_rot,
            position=self.initial_cam_pos,
            fieldOfView=self.initial_cam_fov,
        )
```

ai2thor/video_controller.py
```diff
@@ -386,3 +376,30 @@ def merge_video(self, other_video_path):
...
+                video.write(cv2.cvtColor(frame, cv2.COLOR_BGR2RGB))
+            cv2.destroyAllWindows()
+            video.release()
+            print('done')
+
+with VideoController() as vc:
+    vc.play(vc.MoveAhead())
+    vc.wait(5)
+    vc.play(vc.MoveAhead())
+    vc.exportVideo('thor.mp4')
```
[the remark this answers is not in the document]
d-val: removed! :)

ai2thor/video_controller.py
d-val: Yes, good point. I was trying to support previous calls to `exportVideo' instead of `export_video'.

ai2thor/video_controller.py
```diff
@@ -27,28 +27,18 @@ def __init__(
...
 
-    def reset(self, scene=None, **init_params):
-        """Changes the scene and adds a new third party camera to the initial position."""
-        super().reset(scene, **init_params)
+        super().__init__(continuous=True, **controller_kwargs)
         self.step(
             action="AddThirdPartyCamera",
-            rotation=self.initial_cam_rot,
-            position=self.initial_cam_pos,
-            fieldOfView=self.initial_cam_fov,
```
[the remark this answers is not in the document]
d-val: The reset issue is weird. The VideoController object calls the __init__ of Controller (as expected) which then calls VideoController's reset method which is expected to output an event at `event = self.reset(scene)' in Controller's __init__. This breaks..

ai2thor/video_controller.py
d-val: This still doens't work for me:

```
Traceback (most recent call last):
  File "video_controller_new.py", line 391, in <module>
    with VideoController() as vc:
  File "video_controller_new.py", line 30, in __init__
    super().__init__(continuous=True, **controller_kwargs)
  File "ai2thor/controller.py", line 535, in __init__
    event = self.reset(scene)
  File "video_controller_new.py", line 50, in reset
    rotation=self.initial_cam_rot,
AttributeError: 'VideoController' object has no attribute 'initial_cam_rot'
```

mattdeitke: To clarify, did you move:
```python
        self.step(
            action="AddThirdPartyCamera",
            rotation=initial_camera_rotation,
            position=initial_camera_position,
            fieldOfView=initial_camera_fov,
        )
```
below:
```python
        self.saved_frames = []
        self.ceiling_off = False
        self.initial_cam_rot = cam_rot.copy()
        self.initial_cam_pos = cam_pos.copy()
        self.initial_cam_fov = cam_fov
```
If so, it should be setting `self.initial_cam_rot` directly prior to using it.

ai2thor/video_controller.py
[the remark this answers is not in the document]
mattdeitke: I see what you mean. Nothing is being returned here with the `reset` method. I would change it to return the event from adding the third party camera:

```python
    def reset(self, scene=None, **init_params):
        """Changes the scene and adds a new third party camera to the initial position."""
        super().reset(scene, **init_params)
        return self.step(
            action="AddThirdPartyCamera",
            rotation=self.initial_cam_rot,
            position=self.initial_cam_pos,
            fieldOfView=self.initial_cam_fov,
        )
```

ai2thor/video_controller.py
d-val: Yes I did. :) I think the problem is that the VideoController and Controller objects's reset function are going back and forth and self.<object property> is defined in one class's scope and not in the other one. There's a number of ways to resolve it but you'd know better given the many uses you want to support 

```
class VideoController(Controller):
    def __init__(
        self,
        cam_rot=dict(x=85, y=225, z=0),
        cam_pos=dict(x=-1.25, y=7.0, z=-1.0),
        cam_fov=60,
        **controller_kwargs,
    ):
        super().__init__(continuous=True, **controller_kwargs)

        self.saved_frames = []
        self.ceiling_off = False
        self.initial_cam_rot = cam_rot.copy()
        self.initial_cam_pos = cam_pos.copy()
        self.initial_cam_fov = cam_fov

        self.step(
            action="AddThirdPartyCamera",
            rotation=self.initial_cam_rot,
            position=self.initial_cam_pos,
            fieldOfView=self.initial_cam_fov,
        )

    def reset(self, scene=None, **init_params):
        """Changes the scene and adds a new third part
    ...
```

ai2thor/video_controller.py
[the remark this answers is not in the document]
d-val: Done!

PR #862, opened 2021-08-27
Adds isInteractable, removes obstructed from the object metadata

winthos:
Previously, the `obstructed` value of object metadata was undocumented and only reflected the opposite of the visibility of an object, ie:
```
objMeta.obstructed = !isVisible;
```

This is not useful, and does not reflect the original use of this value. At some point we streamlined the object meta values and in that cleanup, it seems the use of `obstructed` as well as `isInteractable` got muddled.

Prior to this PR, when an object is specified as a target for actions like `PickupObject` or `PushObject` it would only check to ensure that the object was visible. If the object was occluded by a transparent object, like glass doors, then these actions would still succeed even though the target object behind the glass would not technically be reachable. This allowed action sequences like being able to push a sponge that was behind glass, simply because it was visible.

Now sim objects once again report if they are interactable, meaning that they are both visible and not behind some transparent object. Logic for `GetTargetObject` has been combined with similar helper functions into a single `getInteractableSimObjectFromId` function. This PR adds back these above mentioned additional checks to the `getInteractableSimObjectFromId` helper function that actually performs the raycasts to see if an object is visible and or obstructed, and only if an object is both visible and interactable will the action succeed.

```
objMeta.isInteractable = simObj.isInteractable
```

Now the `obstructed` value has been replaced with the more descriptive `isInteractable` as this will only ever be true if a transparent object is in between the object in question and the agent camera.

If an object is fully occluded (ie, sponge inside of a drawer) then it will return its visibility as false, as well as its `isInteractable` as false.

Some additional changes:
- changed all state change actions to use `getInteractableSimObjectFromId` helper function, just updating them
- unit test to check for interactability behind glass in a bathroom


mattdeitke (approved):
LGTM.

I'm ignoring the multi-agent issue though.

Also, would like to somehow relate this PR to deprecating the use of the word `visible`, but can do that in a future PR.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1671,7 +1671,9 @@ abstract public class BaseFPSAgentController : MonoBehaviour {
             // in the multiagent setting, explicitly giving this information for now.
             objMeta.visible = isVisible; // simObj.isVisible;
 
-            objMeta.obstructedByTransparentObject = !simObj.isInteractable;// object is not interactable. It may still be visible, but occluded by some transparent object (glass)
+            //determines if the objects is unobstructed and interactable. Objects visible behind see-through geometry like glass will be isInteractable=False even if visible
+            //note using forceAction=True will ignore the isInteractable requirement
+            objMeta.isInteractable = simObj.isInteractable;
```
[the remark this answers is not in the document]
Lucaweihs: The issue here is that a `SimObjPhysics` should not "know" if it is interactable. We should do the same thing we do for visibility, i.e. when generating the object metadata we run
```c#
HashSet<SimObjPhysics> visibleSimObjsHash = new HashSet<SimObjPhysics>(GetAllVisibleSimObjPhysics(...));
```
with something like 
```
HashSet<SimObjPhysics> visibleSimObjsHash = new HashSet<SimObjPhysics>(GetAllVisibleSimObjPhysics(...));
HashSet<SimObjPhysics> interactableSimObjsHash = new HashSet<SimObjPhysics>(GetAllInteractableSimObjPhysics(...));
```
Of course it would be more efficient to have the two above function calls combined into one.

Lucaweihs (changes requested):
As mentioned in an above comment, a `SimObjPhysics` should not contain information about it's interactability.

unity/Assets/Scripts/BaseFPSAgentController.cs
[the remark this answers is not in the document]
ekolve: The two options I see are to consider moving to the distance based visibility scheme introduced last year, but not necessarily compatible with the current default scheme.  Then this method can be used in `getTargetObject`:
https://github.com/allenai/ai2thor/blob/fb126c24d7682bafcbd5192fd347408ed893f2bb/unity/Assets/Scripts/BaseFPSAgentController.cs#L2744

Alternatively, we could create a new method based on this method:
https://github.com/allenai/ai2thor/blob/fb126c24d7682bafcbd5192fd347408ed893f2bb/unity/Assets/Scripts/BaseFPSAgentController.cs#L2856
that is able to just tell you whether a single SimObjPhysics is visible.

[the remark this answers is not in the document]
ekolve: Sorry for all the back and forth, I read through the code once again - and this and the getTargetObject below don't actually have any problems with multi-agent usage since a visibility check is performed immediately before consulting the `isInteractable` flag.  That said,  I agree with @Lucaweihs's point about `isInteractable` should not reside on the SimObjPhysics since its a boolean that reflects that state of both the agent and the object (not just the object).  I have an idea for how to refactor this - I can update this PR with the change.

ekolve (approved):
My changes that remove `simObj.isInteractable` and tracks interactability independent of the visibility were merged in - so lgtm.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2031,16 +2036,25 @@ abstract public class BaseFPSAgentController {
...
-            // if object is in the scene and visible, assign it to 'target'
-            SimObjPhysics target = getInteractableSimObjectFromId(objectId: objectId, forceVisible: forceAction);
+            
+            SimObjPhysics sop = getSimObjectFromId(objectId);
+            if (sop == null) {
+                throw new NullReferenceException("Object with id'" + objectId + "' is null");
+            }
+            
+            SimObjPhysics[] interactable;
+            bool visible = GetAllVisibleSimObjPhysics(camera: this.m_Camera, maxDistance: this.maxVisibleDistance, out interactable, filterSimObjs: new List<SimObjPhysics> { sop }).Length == 1;
```
Lucaweihs, line 2046: It seems like we should have a nicer way of doing this, is the reason we use `GetAllVisibleSimObjPhysics` rather than `isSimObjVisible` because we need to account for the difference between `GetAllVisibleSimObjPhysicsCollider` and `GetAllVisibleSimObjPhysicsDistance`?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2955,14 +2986,16 @@ abstract public class BaseFPSAgentController {
...
-                                }
+                            sop.debugIsVisible = visCheck.visible;
+                            sop.debugIsInteractable = visCheck.interactable;
+#endif                
+                            if (visCheck.visible && !currentlyVisibleItems.Contains(sop)) {
+                                currentlyVisibleItems.Add(sop);
+                            }
+
+                            if (visCheck.interactable && !interactableItems.Contains(sop)) {
+                                interactableItems.Add(sop);
```
Lucaweihs, line 2998: It seems like doing these `Contains` checks might be slow as `currentlyVisibleItems` and `interactableItems` are lists. Why not make these `HashSets`?

unity/Assets/Scripts/DroneFPSAgentController.cs
```diff
@@ -116,7 +116,7 @@ public class DroneFPSAgentController : BaseFPSAgentController {
         }
 
         // generates object metadata based on sim object's properties
-        public override ObjectMetadata ObjectMetadataFromSimObjPhysics(SimObjPhysics simObj, bool isVisible) {
+        public override ObjectMetadata ObjectMetadataFromSimObjPhysics(SimObjPhysics simObj, bool isVisible, bool isInteractable) {
```
Lucaweihs, line 119: `isInteractable` doesn't seem to be used below?

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -5910,8 +5876,8 @@ bool markActionFinished
                     // Debug.Log(Vector3.Distance(tmp, transform.position));
                     if (Vector3.Distance(tmp, transform.position) < maxDistance) {
                         // if this particular point is in view...
-                        if (CheckIfVisibilityPointInViewport(sop, point, m_Camera, false) ||
-                            CheckIfVisibilityPointInViewport(sop, point, m_Camera, true)) {
+                        if ((CheckIfVisibilityPointInViewport(sop, point, m_Camera, false) |
+                            CheckIfVisibilityPointInViewport(sop, point, m_Camera, true)).visible) {
```
Lucaweihs, line 5880: This would be better as 
```
                        if ((CheckIfVisibilityPointInViewport(sop, point, m_Camera, false).visible ||
                            CheckIfVisibilityPointInViewport(sop, point, m_Camera, true).visible)) {
```
as using `||` allows C# to skip running the second visibility check if the first one is true.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -3018,14 +3008,9 @@ public struct WhatDidITouch {
...
         ) {
-            if (!physicsSceneManager.ObjectIdToSimObjPhysics.ContainsKey(objectId)) {
-                errorMessage = "Object ID appears to be invalid.";
-                actionFinished(false);
-                return;
-            }
 
             // if object is in the scene and visible, assign it to 'target'
-            SimObjPhysics target = getInteractableSimObjectFromId(objectId, forceAction);
+            SimObjPhysics target = getTargetObject(objectId: objectId, forceAction: forceAction);
```
Lucaweihs, line 3013: Can we change the name of `getTargetObject`? Currently it's unclear what it does (`getInteractableSimObjectFromId` is, for example, much more obvious).

Lucaweihs: Looking at the code a little more, it seems that `getTargetObject` is just a thin wrapper around `getInteractableSimObjectFromId` that actually replicates some of the behavior of `getInteractableSimObjectFromId`. I think these two functions can safely be merged into one.

Lucaweihs (commented):
Updates to interactability look good to me. Left a few more comments on efficiency and naming.

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: Yes, currently the default visibility schema uses the capsule collider based approach. Eric and I were talking about potentially switching over the default to use purely a distance based check like in `isSimObjVisible` as the new default, but that will be addressed in a different PR down the line.

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: it seems `getInteractableSimObjectFromId` is not used anymore after these changes, so I will just remove it. And yes `getTargetObject` is meant to grab a target object by objectId. I'll rename `getTargetObject` to a more descriptive `getTargetObjectFromId` for clarity.

Lucaweihs: I think even `getTargetObjectFromId` is not descriptive enough (what is a "target" object?). The code seems to suggest that we're only returning the object when it's interactable, thus it makes more sense to me to have `getTargetObject` be named something like `getInteractableSimObjectFromId`. Also I believe that `getInteractableSimObjectFromId` is used inside `getTargetObject` unless you've already refactored this.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -4501,4 +4526,17 @@ float allowedError
...
     }
+    
+    public class VisibilityCheck {
+        public bool visible;
+        public bool interactable;
+
+        public static VisibilityCheck operator |(VisibilityCheck a, VisibilityCheck b) {
+            VisibilityCheck c = new VisibilityCheck();
+            c.interactable = a.interactable | b.interactable;
+            c.visible = a.visible | b.visible;
```
Lucaweihs, line 4537: Also we should probably replace `|` with `||` in the below?
```c#
            c.interactable = a.interactable | b.interactable;
            c.visible = a.visible | b.visible;
```

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: `getInteractableSimObjectFromId` has already been moved out of `getTargetObject/getTargetObjectFromId`. I found no other functions that made reference to `getInteractableSimObjectFromId`.

I think changing the name isn't necessarily needed because it doesn't only return objects when they are interactable, since passing forceAction can cause an object to be returned as long as its Id is in the scene. This is sort of just being picky on the semantics though, so I think changing `getTargetObject` to `getinteractableSimObjectFromId` is fine if that is more clear.

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: yup!

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2031,16 +2036,25 @@ abstract public class BaseFPSAgentController {
             if (!physicsSceneManager.ObjectIdToSimObjPhysics.ContainsKey(objectId)) {
                 throw new ArgumentException($"objectId: {objectId} is not the objectId on any object in the scene!");
             }
-
-            // if object is in the scene and visible, assign it to 'target'
-            SimObjPhysics target = getInteractableSimObjectFromId(objectId: objectId, forceVisible: forceAction);
+            
+            SimObjPhysics sop = getSimObjectFromId(objectId);
+            if (sop == null) {
+                throw new NullReferenceException("Object with id'" + objectId + "' is null");
```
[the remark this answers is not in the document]
winthos: this has been updated to use string interpolation

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: updated!

unity/Assets/Scripts/DroneFPSAgentController.cs
winthos: this has been added to drone metadata

unity/Assets/Scripts/BaseFPSAgentController.cs
winthos: these have been updated to use HashSets instead of lists

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: all instances of this function have now been renamed to `getInteractableSimObjectFromId`

Lucaweihs (approved):
LGTM!

PR #869, opened 2021-08-31
refactor agent modes into agent classes

ekolve:
This PR moves all the logic that lived in `BaseFPSAgentController.SetAgentMode()` into each of the corresponding agent classes.  A new `LocobotFPSAgentController` class was added that simply has a different set of renderers from the default.  This maintains the current behavior if someone sets the `agentControllerType` to `stochastic`, but doesn't set the `agentMode="locobot"`.

The other change is to the FirstPersonCharacterCull which now dynamically retrieves which renderers to hide instead of using the manually configured renderers on the component.


winthos (approved):
changes are much more organized than the current version. looks good.

mattdeitke (changes requested):
>  This maintains the current behavior if someone sets the agentControllerType to stochastic, but doesn't set the agentMode="locobot".

Hmm.. do we really want this? The stochastic controller is inherently tied to the LoCoBot, down to the point where the agent is unable to MoveLeft, MoveRight, Crouch, Stand, or do any type of interaction. So, trying to use the stochastic agent controller type with the default agent or the drone will lead to many confusing breaking points (and is the reason `agentControllerType` remains undocumented, as it is implied from `agentMode`).

If somebody wants stochastic actions with some other agent type, I think they should control the amount of stochasticity from Python. This will also allow them to sample whatever random variable distribution they want, instead of just using Gaussians. For instance, one may have:
```python
import numpy as np

controller = Controller(snapToGrid=False)

move_size = 0.25
for _ in range(20):
    noise = np.random.exponential()  # replace with your favorite continuous random variable
    step_size = move_size + noise
    controller.step(action="MoveAhead", moveMagnitude=step_size)
```

Here, I think I would prefer to put all of StochasticRemoteFPSAgentController into LocobotFPSAgentController, and have LocobotFPSAgentController inherit from Base.

ekolve:
I agree with @mattdeitke regarding the change to limit the Stochastic controller to just locobot mode.  The intent of this PR was to refactor how the agentModes/visibilityCapsules were configured and not necessarily introduce any API changes. 

The line in question is this one:
https://github.com/allenai/ai2thor/blob/main/unity/Assets/Scripts/AgentManager.cs#L165

I would be perfectly fine changing that line to throw an exception instead to prevent this combination of agentMode/agentControllerType if we have consensus on it.

mattdeitke:
> I would be perfectly fine changing that line to throw an exception instead to prevent this combination of agentMode/agentControllerType if we have consensus on it.

Yeah, I'm in support of this.

winthos:
> Yeah, I'm in support of this.

I am in support of this support.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -551,21 +551,35 @@ abstract public class BaseFPSAgentController {
...
-               ) {
-                // set agent mode to Default, Bot or Drone accordingly
-                SetAgentMode(action.agentMode);
-            } else {
-                errorMessage = "agentMode must be set to 'default' or 'bot' or 'drone' or 'hand'";
-                Debug.Log(errorMessage);
-                actionFinished(false);
-                return;
-            }
+            this.SetAgentMode();
```
Lucaweihs, line 581: I'm not sure the name `SetAgentMode` really makes sense anymore. Can we change this to `InitializeAgent` or something similar?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -551,21 +551,35 @@ abstract public class BaseFPSAgentController {
...
+
+            // camera position
+            m_Camera.transform.localPosition = new Vector3(0, 0.675f, 0);
+
+            // camera FOV
+            m_Camera.fieldOfView = 90f;
+
+            // set camera stand/crouch local positions for Tall mode
+            standingLocalCameraPosition = m_Camera.transform.localPosition;
+            crouchingLocalCameraPosition = m_Camera.transform.localPosition + new Vector3(0, -0.675f, 0); // bigger y offset if tall
```
Lucaweihs, line 577: Should this be under the `PhysicsFPSAgentController`? I'm not sure it makes sense for the `BaseFPSAgentController` to know about the `TallVisCap` or that standing/crouching is possible.

Lucaweihs (commented):
Added some minor comments but overall really like these changes. I am happy with Winson/Matt's suggestions to merge the locobot/stochastic controllers.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: Done.

unity/Assets/Scripts/BaseFPSAgentController.cs
ekolve: I agree.  How does everyone feel about `initializeBody`? Pretty much all the code in `SetAgentMode()` has to do with setting up the capsule and possibly the arm.

ekolve:
> > Yeah, I'm in support of this.
> 
> I am in support of this support.

PR is now updated to fail when this combination is passed in and a test has been added as well.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: `InitializeBody` sounds good to me!

mattdeitke (approved):
LGTM

PR #873, opened 2021-09-01
moving isDirtyCheck into outer if condition in Dirty

ekolve:
Tiny refactor to improve efficiency of OnTriggerStay of the Dirty component.


winthos (approved):
lgtm

PR #879, opened 2021-09-08
Remove TextmeshPro

ekolve:
Removing TextMeshPro package. Package is unused.

For completeness, the motivation behind the PR is to try and get all the package requirements between the CloudRendering build (Unity 2021.2 Alpha) and our current build to be the same so we don't have to maintain a separate branch for the CloudRendering code.


winthos (approved):
looks good

PR #880, opened 2021-09-08
ai2thor-xorg - adding Interactive false to Screen section

ekolve:
Setting Interactive to "False" to prevent processes from getting killed.

https://download.nvidia.com/XFree86/Linux-x86_64/340.108/README/xconfigoptions.html


Lucaweihs (approved):
LGTM

PR #884, opened 2021-09-09
OnGui in DebugInputField when using a non-physicscontroller

ekolve:
A cast error was thrown when using a non-physics controller in the DebugInputField.


winthos (approved):
looks good

PR #888, opened 2021-09-20
missing component fix / remove unused prefab

ekolve:
This removes several prefabs that references components that have been converted to plain csharp classes.


winthos:
looks good to go, all these assets were used in the development of the drone functionality which is now part of the main FPSController asset, so all this is no longer needed.

PR #892, opened 2021-09-23
adding Python specific changes for CloudRendering backend

ekolve:


ai2thor/controller.py
```diff
@@ -410,6 +411,22 @@ def __init__(
...
         self.include_private_scenes = include_private_scenes
         self.x_display = None
+        self.gpu_device = gpu_device
+        cuda_visible_devices = list(
+            map(int, filter(lambda y: y.isdigit(),
+            map(lambda x: x.strip(),
+                os.environ.get('CUDA_VISIBLE_DEVICES', '').split(',')))))
+
+        if self.gpu_device:
+            if (type(self.gpu_device) is not int or self.gpu_device < 0):
```
Lucaweihs, line 421: `not np.issubdtype(type(self.gpu_device), np.integer)` might be safer if someone passes in a numpy int.

ai2thor/controller.py
```diff
@@ -410,6 +411,22 @@ def __init__(
         self.add_depth_noise = add_depth_noise
         self.include_private_scenes = include_private_scenes
         self.x_display = None
+        self.gpu_device = gpu_device
+        cuda_visible_devices = list(
+            map(int, filter(lambda y: y.isdigit(),
+            map(lambda x: x.strip(),
+                os.environ.get('CUDA_VISIBLE_DEVICES', '').split(',')))))
+
+        if self.gpu_device:
```
Lucaweihs, line 420: Won't this if statement be skipped if `gpu_device == 0`?

ai2thor/controller.py
```diff
@@ -410,6 +411,22 @@ def __init__(
...
+            if (type(self.gpu_device) is not int or self.gpu_device < 0):
+                raise ValueError("Invalid gpu_device: '%s'. gpu_device must be >= 0" % self.gpu_device)
+            elif cuda_visible_devices:
+                if self.gpu_device >= len(cuda_visible_devices):
+                    raise ValueError("Invalid gpu_device: '%s'. gpu_device must less than number of CUDA_VISIBLE_DEVICES: %s" % (self.gpu_device, cuda_visible_devices))
```
Lucaweihs, line 425: Is it possible for `cuda_visible_devices` to be a subset of the available devices on a machine? I.e. `cuda_visible_devices=[6,7]`? If so then maybe we want to change `self.gpu_device >= len(cuda_visible_devices)` to `self.gpu_device not in len(cuda_visible_devices)`?

ai2thor/controller.py
```diff
@@ -410,6 +411,22 @@ def __init__(
...
+                else:
+                    self.gpu_device = cuda_visible_devices[self.gpu_device]
```
Lucaweihs, line 427: Oh I see, this seems a bit weird to me, if I specify `self.gpu_device == 3` then I'd expect gpu 3 to be used and not GPU 7 if I had specified `cuda_visible_devices=[3,4,5,7]`. That said, if this is a common convention (indexing into `cuda_visible_devices` rather than using absolute indices) then I'm happy with that.

Lucaweihs (commented):
Minor comments but looks good to me! Can we add a test that confirms the `x_display` / `CloudRendering` backends return (approximately) the same images for a few test locations / fovs?

ai2thor/controller.py
ekolve: Yes, this was what the preference was from testing with users at CMU.

ai2thor/controller.py
ekolve: Fixed.  Changed type check to use the `numbers` module.

ai2thor/controller.py
ekolve: Fixed.

ekolve:
The build and tests run on Mac, so we can't automate a test comparing the pixels, but a test can be added that can be run on Linux only.

Lucaweihs:
Got it, I didn't think too deeply about this but I suppose this means that the CloudRendering backend won't work on mac? Is this because it requires CUDA? It would be great to have a Linux-only pixel test if this isn't a huge amount of work but I imagine this can also be saved for a future PR, will approve for now.

Lucaweihs (approved):

ekolve:
Correct, the CloudRenderingEngine only runs on Linux and use the Vulkan library for rendering (instead of OpenGL) - so no Mac.

PR #897, opened 2021-09-27
Add platform as Controller() constructor param

ekolve:
This PR adds a new parameter `platform` to the `ai2thor.controller.Controller` constructor.  Prior to this, in order to run say a CloudRendering and Linux64 (Xorg dependent Unity), you would have to do something like this:

```python
import ai2thor.controller
import ai2thor.platform
ai2thor.platform.CloudRendering.enabled = True
ai2thor.platform.Linux64.enabled = False

cloudrendering_controller = ai2thor.controller.Controller()

ai2thor.platform.CloudRendering.enabled = False
ai2thor.platform.Linux64.enabled = True

xorg_controller = ai2thor.controller.Controller()
```

With this change, the following can now be done:
```python
import ai2thor.controller
from ai2thor.platform import CloudRendering, Linux64
xorg_controller = ai2thor.controller.Controller(platform=Linux64)
cloundrendering_controller = ai2thor.controller.Controller(platform=CloudRendering)
```

I feel this change makes the process of expressing a particular backend `platform` explicit and more concise. If no platform is provided, we will iterate through the list of candidate platforms and find one that has its dependencies satisfied.

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -1138,9 +1138,10 @@ class Controller(object):
             platform_system(), self.width, self.height, self.x_display, self.headless
         )
 
-        candidate_platforms = ai2thor.platform.select_platforms(request)
 
-        if platform is not None:
+        if platform is None:
+            candidate_platforms = ai2thor.platform.select_platforms(request)
+        else:
             candidate_platforms = [platform]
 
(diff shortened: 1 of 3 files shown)


Lucaweihs (approved):
Looks nice! I added small comment but am fine with things being merged without this being changed.

Issue #910, opened 2021-10-07
Objects visible to Thirdpartycamera


Karthika-Arumugam:
Hi @mattdeitke,

Given the third party camera id, is it possible to get the list of objects visible in the camera view?. For third party camera currently we get only frames, we need information on what all objects visible in that camera frame. Please advise. Thanks

PR #918, opened 2021-10-14
adds `ObjectsFromThirdPartyCamera` action

winthos:
Adds new action that takes a third party camera index and returns all objects both visible and interactable from that Third Party Camera.

Also added a unit test for this new action.

Addresses issue #910


Lucaweihs:
Functionality now exists.

PR #926, opened 2021-10-28
ResourceAssetManager / ColorChanger refactor

ekolve:
This PR adds a wrapper (ResourceAssetManager) around the Resources class that Unity provides to load assets at runtime instead of making references to them at compile time.  The main feature of the ResourceAssetManager is the ability to find resources by a label which is simple a string applied to an asset.  Labels can be modified through the [AssetDatabase](https://docs.unity3d.com/ScriptReference/AssetDatabase.GetLabels.html) or through the Editor.  In the screenshot below I have filtered all the assets by the label `RawTestMaterials` in the Project window.
Screen Shot 2021-10-28 at 11 36 43 AM

In the screenshot below a single asset can be seen with its labels as well as the menu used to modify the labels.
Screen Shot 2021-10-28 at 11 37 53 AM

The ResourceAssetManager generates a catalog from the AssetDatabase from all the materials that live under `Assets/Resources` and indexes them by their labels.  At build time, this catalog is serialized using JSON and written to a file under `Assets/Resources` so that it can be accessed by the Player at runtime.

The ColorChanger component has been modified to use this new framework to lazily load assets instead of references to the assets themselves.  All the assets that ColorChanger previously referenced have been move under the `Assets/Resources` directory (which explains the large number of files changed in this PR).  As well, the identifiers used for each of the sets of materials (e.g. `rawTrainMaterials`, `rawKitchenMaterials`, `bread`) have been used as the label names for the assets in the form of `RawTrainMaterials`, `RawKitchenMaterials` and `BreadMaterials`.

The primary motivation behind this change was to reduce the amount of GPU memory consumed by the Vulkan driver (required by CloudRendering), which uses ~2x the amount of GPU memory when compare to OpenGL.  Currently, the Unity Player consumes 600MB with Vulkan using CloudRendering.  With this change, the GPU usage drops to 238MB.

I would ultimately like to migrate the new components in the `nanna` branch to use this framework in combination with labels to identify the prefabs that get spawned in.


AlvaroHG:
LGTM, hard to see in the PR but looked in the commit. Good to have on the `nanna` branch.

winthos (approved):
Looks good to me! Could we potentially have some sort of readme or documentation to help track what each new label is meant for? Most of them are pretty self explanatory, but I think it wouldn't hurt to have a more detailed description like "used for training scenes" so if and when we add in new labels, or apply existing labels to new materials (like some of the new materials in the nanna branch) we can keep track of it all.

mattdeitke (approved):
LGTM

elimvb:
LGTM

PR #936, opened 2021-11-15
Clone GameObject when generating OOBB

ekolve:
I discovered that the way that we generate OOBB will wake some rigid bodies up causing them to never settle.  This can be observed by calling `Physics.Simulate()` repeatedly and examining the position and rotation values for the `Lettuce` object in FloorPlan28.  The issue is that when any change is done to the transform of a GameObject Unity will wake the RigidBody.  What this PR does is clone the GameObject and place it at the origin and then perform the OOBB calculation.

One thing that I also noticed was that during the `Object.Instantiate` call , `iTween.Init` needed be be called on the newly cloned objects otherwise iTween would throw an error about a null GameObject. This is the reason for moving the `iTween.Init` calls to `Awake` from `Start` for `CanOpen` and `CanToggle`.


winthos (approved):
looks good!

PR #940, opened 2021-11-29
Calling changequality for CloudRendering after startup; 

ekolve:
The CloudRendering platform does not respect command-line parameters, so the Quality is updated manually after startup/reset.

* moving ChangeQuality to AgentManager
* Calling ChangeQuality for CloudRendering after startup


winthos (approved):
good to go

PR #942, opened 2021-12-01
Remove default lighting from nanna

mattdeitke:
Removes all directional lights, point lights, and reflection probes from the default nanna scene. These must now be specified in the json.


winthos (approved):
yup! get rid of em!

PR #951, opened 2021-12-07
Stretch robot restructured arm updated secondary camera image synthesis

elimvb:
Simple way of updating secondary camera image synthesis based on main camera

### DIFF
--- a/unity/Assets/Scripts/StretchAgentController.cs
+++ b/unity/Assets/Scripts/StretchAgentController.cs
@@ -15,7 +15,13 @@ namespace UnityStandardAssets.Characters.FirstPerson {
 
         public override void updateImageSynthesis(bool status) {
             base.updateImageSynthesis(status);
-            agentManager.updateThirdPartyCameraImageSynthesis(true);
+
+            // updateImageSynthesis is run in BaseFPSController's Initialize method after the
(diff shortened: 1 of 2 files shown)


unity/Assets/Scripts/StretchAgentController.cs
```diff
@@ -13,6 +13,11 @@ public partial class StretchAgentController : PhysicsRemoteFPSAgentController {
         public StretchAgentController(BaseAgentComponent baseAgentComponent, AgentManager agentManager) : base(baseAgentComponent, agentManager) {
         }
 
+        public override void updateImageSynthesis(bool status) {
+            base.updateImageSynthesis(status);
+            agentManager.updateThirdPartyCameraImageSynthesis(true);
```
ekolve, line 18: This should get changed to pass `status` to `agentManager.updateThirdPartyCameraImageSynthesis`

ekolve: It would also make sense to add a comment above the `updateThirdPartyCameraImageSynthesis` line explaining why this is necessary since it wouldn't be obvious that the secondary camera is assumed to have been added.

ekolve (changes requested):
Looks good, just a couple changes requested.

ekolve (approved):
lgtm!

PR #953, opened 2021-12-12
Initial Nanna JSON Cleanup

mattdeitke:
## Changes

Converts every field in the .json to be in camel case. Currently there is a mix between camel case and snake case. Camel case was chosen because it is most consistent with the current version of AI2-THOR and Unity/C#.

Renames:
- Light parameter `objectId` to `linkedObjectId`, where when the `linkedObjectId` is toggled, it is supposed to toggle the light on/off.
- Room parameter `type` to `roomType`.

Removes:
- `roomId` from light parameters. It is not necessary and is never used.

I have run it on a few test files and it appears to work successfully.

## Backwards Compatibility

Breaks backwards compatibility on all snake case json keys and for the 2 keys that have been renamed.


AlvaroHG:
When this format was conceived, we decided to go snake case because most of the interaction and changes to the json happen in python and this style is matches the way we write python code.

The current classes that are used in he ProceduralHouse json serialization hierarchy are consistently in snake case, I could not find one variable that was in camel case. There are some that are unity auxiliary classes that have camel cased variables, but are not a part of the json hierarchy.

Also under an object `room` it seems redundant to have `room_id` be a variable instead of just `id`.

I would vote to keep it as it is in snake case and not break compatibility with everyone that has been using and keep a decision that was made a few months ago unless there is something more substantial that requires this change.

mattdeitke:
> When this format was conceived, we decided to go snake case because most of the interaction and changes to the json happen in python and this style is matches the way we write python code.

Hmm, I might have missed this decision a few months ago, but sticking with snake case is not a light decision and completely breaks consistency with AI2-THOR. If we use snake case here, some variables passed into actions will now have snake case, whereas others use camel case.

I especially don't think this makes sense for passing in `object_id` to the `CreateHouse` function, and then later referencing it as `objectId` in all other actions (e.g., `OpenObject`, `TeleportObject`, etc.)

It should also be noted that what is passed into `CreateHouse` is just a normal Python dict and not a really a JSON file. Therefore, I'm don't think that is enough to justify this being different from the rest of what is passed into AI2-THOR.

> The current classes that are used in he ProceduralHouse json serialization hierarchy are consistently in snake case, I could not find one variable that was in camel case. There are some that are unity auxiliary classes that have camel cased variables, but are not a part of the json hierarchy.

There are quite a few. Some examples: `materialId`, `ceilingMaterialId`, `rectangleFloor`, `cornersClockwise`, etc.

Check this file: https://github.com/allenai/ai2thor/blob/nanna/unity/Assets/Scripts/ProceduralData.cs. 

> Also under an object `room` it seems redundant to have `room_id` be a variable instead of just `id`.

I disagree with this because `roomId` is referenced all the time in other parts of the `CreateHouse` dict (e.g., in the construction of walls). Moreover, it will presumably be referenced in future actions (similar to how `objectId` is referenced all over the place). Using `id` everywhere doesn't make it clear that this `id` will be referred to everywhere else as `roomId`.

For the same reason, I think all `id` variables should be more explicit, but that is currently beyond the scope of this initial cleanup.

> I would vote to keep it as it is in snake case and not break compatibility with everyone that has been using and keep a decision that was made a few months ago unless there is something more substantial that requires this change.

I am really quite against sticking with snake case, but open to hearing other thoughts and a responses. I just don't see why one action would use snake case when all the other actions would use camel case.

Moreover, this would make it really confusing defining new actions moving forward. For instance, suppose we have a new action `PlaceObjectInRoom`. Should it be specified as:
```python
controller.step(action="PlaceObjectInRoom", room_id=X)
```
Or
```python
controller.step(action="PlaceObjectInRoom", roomId=X)
```
Either way, if we now use snake case for the `CreateHouse` action, it would make either decision weird.

As a side note: currently, the only `nanna` action is `SpawnAsset` and this uses camel case.

Finally, I don't think breaking compatibility is a big issue right now. I've already converted my codebases to be working on `nanna-camel`, which has all of these updates, and it is all working. I will also help anybody else convert their codebases, since I think using snake casing is definitely not a light decision. Also, as mentioned above, there are still variables that use camel case, so even converting them to snake case will break compatibility.

ekolve (approved):

AlvaroHG:
Compatibility was the main issue I had, but if it's not considered an issue right now I am ok with the change.  I see now what you mean, as some Unity actions have as arguments fields from json representing a house. While the house itself is consistently in snake case, all other unity actions have camel cased arguments. LGTM

winthos (approved):

PR #958, opened 2021-12-15
Fix Python+WSGI -> Unity Editor

ekolve:
Calls to `ChangeResolution` fail when connecting from Python to the Unity Editor.  This PR resolves that.


winthos (approved):

mattdeitke (approved):
LGTM

PR #959, opened 2021-12-18
First 5 Target houses setup

winthos:
New scenes using objects from the Test group are setup and available at:
`Assets/Scenes/Procedural/Target_House_0[1-5]`


winthos:
incorrectly assigned to merge into main, dont do that!

PR #971, opened 2022-01-19
Remove colorBounds metadata property

ekolve:
When using the `renderInstanceSegmentation` flag, we generate a `colorBounds` property which is used to populate the `instance_detections2D` field in the event (Python side).  The current way this is done is extremely inefficient since we effectively grab the data twice for the segmented image.  We then iterate pixel by pixel to determine the bounds.  This is has been moved to the Python side and is done using numpy.  I observed a performance improvement of ~60%.  If for some reason `colorBounds` is being used on the Python side by an end user or for some reason we decide we need it, we can populate the metadata directly from the python side with the bounding boxes and corresponding colors.


AlvaroHG:
Great find, LGTM

mattdeitke (approved):
LGTM :)

PR #976, opened 2022-02-01
zero-copy depth frame

ekolve:
This PR modifies the data that is sent from the Unity Depth shader to be a float32 that can be directly read from numpy using `numpy.frombuffer` which avoids performing any copying.  Additionally, the default format that we expose to users is in meters.  Currently, the Depth shader outputs normalized values (0.0 - 1.0), which we then convert to meters by multiply by the difference between the far and near camera planes.  This conversion is now done directly in the shader which allows us to make this a zero-copy operation.  

If we choose to say change the default output to be normalized values in the future or we simply want to make all conversion faster, we can push the value the user passes into `ai2thor.controller.Controller` into the Depth shader to directly perform the conversion there.  The improvement to FPS was around 20% for a 600x600 frame and 5-10% for the default 300x300.

I validated that the new numbers are close to the current numbers by doing several tests that compared the values pixel wise as well as a test that moved the agent 0.0125 meters at a time away from a wall and compared the relative values for both the new and old.  My plan is to merge this into main after the ECCV deadline.

256-depth
float32-depth


dhruvbatra:
Clever catch :-).

unity/Assets/Scripts/ImageSynthesis/ImageSynthesis.cs
```diff
@@ -427,7 +431,7 @@ enum ReplacelementModes {
         var renderRT = (!needsRescale) ? finalRT :
             RenderTexture.GetTemporary(mainCamera.pixelWidth, mainCamera.pixelHeight, depth, format, readWrite, antiAliasing);
         if (tex == null) {
-            tex = new Texture2D(width, height, TextureFormat.RGB24, false);
+            tex = new Texture2D(width, height, TextureFormat.RGBA32, false);
```
AlvaroHG, line 434: Is this `Encode` method called elsewhere? It could be good to parametrize the textureformat/bit-depth so that only the code calling for depth uses 32-bits/RGBA32

unity/Assets/Scripts/ImageSynthesis/Shaders/Depth.shader
```diff
@@ -20,6 +22,9 @@ Shader "Hidden/Depth" {
              uniform sampler2D _CameraDepthTexture;
              uniform fixed _DepthLevel;
              uniform half4 _MainTex_TexelSize;
+
+             float _NearClipPlane;
+             float _FarClipPlane;
```
AlvaroHG, line 27: This uniforms are not needed as unity provides the builtin variable `_ProjectionParams`, where `_ProjectionParams.y` is the near plane and `_ProjectionParams.z` is the far plane. https://docs.unity3d.com/Manual/SL-UnityShaderVariables.html

unity/Assets/Scripts/ImageSynthesis/ImageSynthesis.cs
```diff
@@ -235,6 +235,10 @@ enum ReplacelementModes {
         if (!depthMaterial || depthMaterial.shader != depthShader) {
             depthMaterial = new Material(depthShader);
         }
+
+        depthMaterial.SetFloat("_NearClipPlane", capturePasses[1].camera.nearClipPlane);
+        depthMaterial.SetFloat("_FarClipPlane", capturePasses[1].camera.farClipPlane);
```
AlvaroHG, line 240: We can save this calls if we use the built-in shader variables. Comments below.

unity/Assets/Scripts/ImageSynthesis/Shaders/Depth.shader
ekolve: Nice! I'll switch to using those.

unity/Assets/Scripts/ImageSynthesis/Shaders/Depth.shader
```diff
@@ -51,24 +56,21 @@ Shader "Hidden/Depth" {
...
-				 float highBits = 256 * 256 * (depth01 - lowBits - medBits / 256);
-			  	 highBits = floor(256 * highBits) / 256;
+                 float f1 = float(depthUint & 255) / 255.0;
+                 float f2 = float((depthUint >> 8) & 255) / 255.0;
+                 float f3 = float((depthUint >> 16) & 255)  / 255.0;
+                 float f4 = float((depthUint >> 24) & 255) / 255.0;
 
-				 return fixed4(lowBits, medBits, highBits, 0.0);
-                 // return depth01;
+                 return fixed4(f1, f2, f3, f4);
```
AlvaroHG, line 70: We may be able to change the shader signature and shader's operating texture format to be a 32bit float to not have to transform to ints. I remember having problems in Metal with something similar so just a note

unity/Assets/Scripts/ImageSynthesis/Shaders/Depth.shader
```diff
@@ -51,24 +56,21 @@ Shader "Hidden/Depth" {
              fixed4 frag(output o) : COLOR
              {
                  // depth01 = pow(LinearEyeDepth(depth01), _DepthLevel);
-                 float depth01 = Linear01Depth(UNITY_SAMPLE_DEPTH(tex2D(_CameraDepthTexture, o.uv)));
-
-                 // float lowBits = depth01 / 2;
-                 // float medBits = depth01 / 4;
-                 // float highBits = depth01 / 8;
+                 float multiplier = _FarClipPlane - _NearClipPlane;
+                 float depth01 = Linear01Depth(UNITY_SAMPLE_DEPTH(tex2D(_CameraDepthTexture, o.uv))) * multiplier;
```
AlvaroHG, line 60: If we want to stick with our 3D "units" ("meters"), as opposed to normalized, we may as well just use the`LinearEyeDepth` function instead, which returns the depth like that. Where 1 is a surface that is 1 unit from the camera’s pivot along the camera’s z axis.

AlvaroHG:
Final note, it fits more within the depth standard to return it as 0-1 values from the data, and transform it. But it's true we use "meters" a lot more, and it's the default so  not opposed to this change. Further optimizations may include having different shaders for each format and changing the actual bits that are being sent because I could imagine someone using the raw data and realizing it doesn't change regardless of the setting they passed.

A couple of comments, LGTM after that.

unity/Assets/Scripts/ImageSynthesis/ImageSynthesis.cs
ekolve: The 2D texture is used by all the other passes.  So the segmentation cameras also re-use the same texture. I have updated the Python code to handle receiving a 4 byte frame versus a 3 byte frame with the PR.

unity/Assets/Scripts/ImageSynthesis/Shaders/Depth.shader
ekolve: I just tried `LinearEyeDepth`: 
```
float depth01 = LinearEyeDepth(UNITY_SAMPLE_DEPTH(tex2D(_CameraDepthTexture, o.uv)));
```
and I was getting a result that didn't include the near plane offset - If say the agent was 3.0m from a wall, the depth value was getting reported back as 3.0m instead of 2.985 (what we currently get back).

As well, I prefer the way we have it currently because it gives us the flexibility to push the conversion into the depth shader as a parameter in the future in the form of the `multiplier`.

AlvaroHG: Makes sense

PR #979, opened 2022-02-08
Lazy Masks / Detections2D

ekolve:
Currently, when using instance segmentation (`ai2thor.controller.Controller(renderInstanceSegmentation=True)`) instance masks, class masks and 2D detections are generated for every object in the image.  This process is quite costly and end-users don't often use every single bounding-box or mask that is generated. This PR introduces lazily generated masks and 2D detections.  In the absolute best case, users will get a 4x speed improvement (this means that no masks or 2D detections are generated, but the instance segmentation frame is sent from Unity to Python). In the worst case (we generate all the bounding boxes and masks) users should see a 2x improvement over the current implementation.   This speed improvement is largely due to the elimination of the use of `np.unique` and `np.tile` in favor of a zero-copy based approach.  The relevant properties on `Event` are: `instance_masks`, `class_masks`, `instance_detections_2D`, and `class_detections_2D`.  All these properties now point to a class that implements the `Mapping` interface, but the behavior should be identical to the current scheme (i.e. if a key doesn't exist in the lazy-map a KeyError will get thrown). My plan is to merge this in post-ECCV.


ai2thor/server.py
```diff
@@ -22,6 +23,263 @@ def default(self, obj):
...
+        return self._detections2d[key]
+
+    def __len__(self):
+        return len(self.instance_masks)
+
+    def __iter__(self):
+        for k in self.instance_masks.keys():
+            self.__getitem__(k)
+
+        return iter(self._detections2d)
```
Lucaweihs, line 70: Why not just `self.instance_masks.keys()`?

ai2thor/server.py
```diff
@@ -22,6 +23,263 @@ def default(self, obj):
...
+
+
+class LazyClassDetections2D(LazyInstanceDetections2D):
+
```
Lucaweihs, line 74: LGTM.com notes that this class doesn't override `__eq__` which I guess is a potential problem since it adds a new `_loaded` parameter. That said, I don't think this is a problem as you check that everything is subclassed properly.

ai2thor/server.py
```diff
@@ -22,6 +23,263 @@ def default(self, obj):
...
+        self._load_all()
+        return len(self._detections2d)
+
+    def _load_all(self):
+        if not self._loaded:
+            all_integer_keys = self.instance_masks.unique_integer_keys
+            for cls, colors in self.instance_masks.class_colors.items():
+                for color in colors:
+                    if self.instance_masks._integer_color_key(color) in all_integer_keys:
+                        self.__getitem__(cls)
```
Lucaweihs, line 85: Out of curiosity: is there a reason to prefer using `self.__getitem__(cls)` to `self[cls]`?

ai2thor/server.py
```diff
@@ -22,6 +23,263 @@ def default(self, obj):
...
+
+        detections = []
+
+        for color in self.instance_masks.class_colors.get(cls, []):
+            mask = self.instance_masks.instance_segmentation_frame_uint32 == self.instance_masks._integer_color_key(color)
+            bb = self.mask_bounding_box(mask)
+            if bb:
+                detections.append(bb)
+        if detections:
+            self._detections2d[cls] = detections
```
Lucaweihs, line 107: As `detections` is cached can we make it a tuple so people don't accidentally modify it and have the change be persistent?

ai2thor/server.py
```diff
@@ -22,6 +23,263 @@ def default(self, obj):
...
+            raise KeyError(key)
+
+        return m
+
+    def __iter__(self):
+        self._load_all()
+        return iter(self._masks)
+
+    def __len__(self):
+        self._load_all()
```
Lucaweihs, line 132: Should we make `_load_all` an `@abstractmethod` in `LazyMask`?

ai2thor/server.py
```diff
@@ -22,6 +23,263 @@ def default(self, obj):
...
+            cls = c["name"]
+
+            if "|" in c["name"]:
+                self.instance_colors[c["name"]] = c["color"]
+                cls = c["name"].split("|")[0]
+
+            if cls not in self.class_colors:
+                self.class_colors[cls] = []
+
+            self.class_colors[cls].append(c["color"])
```
Lucaweihs, line 185: This seems to make the assumption that a object's name will always be of the form `OBJECT_TYPE|SOMETHING`, I don't think that's generally true (especially in nanna). I think we'll need to cross reference this against `metadata["objects"]` to get the class.

ai2thor/tests/test_event.py
```diff
@@ -5,6 +5,7 @@
 import json
 import numpy as np
 import pytest
+from ai2thor.tests.constants import TESTS_DATA_DIR, TEST_SCENE
```
Lucaweihs, line 8: Seems like `TESTS_DATA_DIR` is imported on line 10 again?

ai2thor/server.py
Lucaweihs: Ah, looking below I suppose this is for backwards compatibility reasons? I think this is something we should think about breaking as it's a bit weird to have things like `StandardUpperCabinetHeightWidth` as a class.

Lucaweihs (commented):
Looks nice! A couple of minor comments here and there, it would be great to have a bit of documentation about the inputs/outputs for some of the more important, user facing, functions.

ai2thor/tests/test_event.py
ekolve: good catch, removed.

ai2thor/server.py
ekolve: Yes, exactly.  This PR should have the same class/objectId entries as the previous version.  For reference, this is where the class names come from: https://github.com/allenai/ai2thor/blob/main/unity/Assets/Scripts/ImageSynthesis/ImageSynthesis.cs#L300

ai2thor/server.py
ekolve: good suggestion, done.

ai2thor/server.py
ekolve: I'm fine with the change, though its currently possible for users to modify the detections on the metadata since that is a list and would behave identically to this. The detections are cached for the `cls`, but only for the current frame.  The next frame will result in a new Lazy class.

ai2thor/server.py
ekolve: The following error gets thrown if you just return `self.instance_masks.keys()`

```python
TypeError: iter() returned non-iterator of type 'KeysView
```
But works fine if you return `iter(self.instance_masks.keys())`

This is much nicer since it allows people to iterate through the keys without triggering the 2D generation.

ai2thor/server.py
ekolve: I went ahead and implemented the `__eq__` method to silence the lgtm alert.

ai2thor/server.py
```diff
@@ -22,6 +23,263 @@ def default(self, obj):
         return super(NumpyAwareEncoder, self).default(obj)
 
 
+class LazyInstanceDetections2D(Mapping):
```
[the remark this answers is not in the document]
ekolve: Added type hints for all the new classes/methods.  Ran mypy on ai2thor/server.py and it returns no errors.

Lucaweihs (approved):
LGTM, having the type hints is very nice :)!

PR #980, opened 2022-02-10
Remove SSAO component from FirstPersonCharacter

ekolve:
This PR removes the ScreenSpaceAmbientOcclusion component from the FirstPersonCharacter/FPSController prefab.  We no longer use SSAO and it can get re-enabled if `ChangeQuality` is called with a value other than Low or Very Low.  This results in the following frame:
out

Note the difference when compared to the frame received with quality set to Ultra, but never calling `ChangeQuality`
base-frame-1


Lucaweihs (approved):
LGTM

elimvb:
LGTM

PR #982, opened 2022-02-14
Enable JavascriptInterface within AgentManager

ekolve:
The WebGL demo was throwing a nullreferenceexception from DebugInputField because the `CurrentActiveController()` had not yet been initialized within the AgentManager.  The line in question was attempting to enable the `JavascriptInterface` component.  This has been moved to the AgentManager instead immediately after the `primaryAgent` initialization.


AlvaroHG:
LGTM

PR #989, opened 2022-02-22
Serialize droneMetadata

ekolve:
Currently, drone specific metadata is getting stripped due to msgpack.  Msgpack determines the formatter for a class at compile-time unless a type specific formatter is defined.  This is by design for performance, but it causes properties that have been added to subclasses to get omitted when assigned to properties that are of the base type.  This PR adds two custom formatters for each of the properties that reference drone metadata.


winthos (approved):
looks good! What an interesting one.

PR #997, opened 2022-03-01
Expert action accepting target position

AlvaroHG:
Adds  option of `position` argument to `ObjectNavExpertAction` to track a position instead of a SimObject when `objectType` and `objectId` are not passed.


AlvaroHG:
Merging, trivial change

PR #998, opened 2022-03-10
snapToGrid error when stepDegrees not 90, 180, 270

AlvaroHG:
* Throws error when setting `snapToGrid` to `true` and `rotateStepDegrees` to something different than 0, 90, 180, 270 or modulo 0 of 360, which mean can make it go outside of the grid.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -554,6 +554,11 @@ namespace UnityStandardAssets.Characters.FirstPerson {
 
         public abstract void InitializeBody();
 
+         private bool ValidRotateStepDegreesWithSnapToGrid(float rotateDegrees) {
+            // float eps = 0.00001f;
+            return rotateDegrees == 90.0f || rotateDegrees == 180.0f || rotateDegrees == 270.0f || (rotateDegrees % 360.0f) == 0.0f;
+        }
+
         public void Initialize(ServerAction action) {
             this.InitializeBody();
             m_Camera.GetComponent<FirstPersonCharacterCull>().SwitchRenderersToHide(this.VisibilityCapsule);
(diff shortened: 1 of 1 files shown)


AlvaroHG:
Wondering if I should epsilon test the angle values, but we do not want any error propagating for the grid so the initial values should be set exactly, but open to thoughts.

winthos:
yeah comparing epsilons is a slippery slope I think. Better to have stricter limits since that would be the intended use case for snap to grid to begin with.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -595,6 +600,13 @@ abstract public class BaseFPSAgentController {
                 this.rotateStepDegrees = action.rotateStepDegrees;
             }
 
+             if (action.snapToGrid && !ValidRotateStepDegreesWithSnapToGrid(action.rotateStepDegrees)) {
+                errorMessage = $"Invalid values 'rotateStepDegrees': ${action.rotateStepDegrees} and 'snapToGrid':${action.snapToGrid}. 'snapToGrid': 'True'a is not supported when 'rotateStepDegrees' is different from grid rotation steps of 0, 90, 180, 270 or 360.";
```
winthos, line 604: looks good to go except slight typo in the error message `'True'a is not`, remove the "a"

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -554,6 +554,11 @@ abstract public class BaseFPSAgentController {
 
         public abstract void InitializeBody();
 
+         private bool ValidRotateStepDegreesWithSnapToGrid(float rotateDegrees) {
+            // float eps = 0.00001f;
```
winthos, line 558: instead of some epsilon, might be worth using `Mathf.approximately` to deal with floats being weird https://docs.unity3d.com/ScriptReference/Mathf.Approximately.html

unity/Assets/Scripts/BaseFPSAgentController.cs
AlvaroHG: Done

unity/Assets/Scripts/BaseFPSAgentController.cs
AlvaroHG: approximately also uses an epsilon to test so I'll keep it as it is.

winthos: ah yes I see. As is works for me then.

winthos (approved):
all good!

PR #1013, opened 2022-04-19
mapping gpu_device to Vulkan device index

ekolve:
This PR ensures that the `gpu_device` Controller param matches the indexes returned from nvidia-smi.


mattdeitke (approved):
LGTM :)

Lucaweihs:
@AlvaroHG - It would be great if you could try to reproduce this as well.

PR #1021, opened 2022-05-03
Changing the maxUpwardLookAngle of the locobot so that it can look upwards by up to 30 degrees.

Lucaweihs:
This reverts a change made several months ago and allows the locobot to look upwards above the horizon line.

This PR also allows the locobot to look downwards by up to 60 degrees as is mentioned in the documentation.

### DIFF
--- a/unity/Assets/Scripts/LocobotFPSAgentController.cs
+++ b/unity/Assets/Scripts/LocobotFPSAgentController.cs
@@ -52,9 +52,9 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             standingLocalCameraPosition = m_Camera.transform.localPosition;
             crouchingLocalCameraPosition = m_Camera.transform.localPosition + new Vector3(0, -0.2206f, 0);// smaller y offset if Bot
 
-            // limit camera from looking too far down
-            this.maxDownwardLookAngle = 30f;
-            this.maxUpwardLookAngle = 30f + 1e-6f; // Adding 1e-6f in case of floating point issues.
+            // limit camera from looking too far down/up
+            this.maxDownwardLookAngle = 60f + 1e-6f; // Adding 1e-6f in case of floating point issues.
+            this.maxUpwardLookAngle = 30f + 1e-6f;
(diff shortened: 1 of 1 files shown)


unity/Assets/Scripts/LocobotFPSAgentController.cs
```diff
@@ -52,9 +52,9 @@ public class LocobotFPSAgentController : BaseFPSAgentController {
             standingLocalCameraPosition = m_Camera.transform.localPosition;
             crouchingLocalCameraPosition = m_Camera.transform.localPosition + new Vector3(0, -0.2206f, 0);// smaller y offset if Bot
 
-            // limit camera from looking too far down
-            this.maxDownwardLookAngle = 30f;
-            this.maxUpwardLookAngle = 29f;
+            // limit camera from looking too far down/up
+            this.maxDownwardLookAngle = 60f + 1e-6f; // Adding 1e-6f in case of floating point issues.
```
winthos, line 56: the epsilon should not be needed. The helper function `CheckForUpDownAngleLimit` that parses these max angles accounts for the epsilon already

winthos (changes requested):

unity/Assets/Scripts/LocobotFPSAgentController.cs
```diff
@@ -52,9 +52,9 @@ public class LocobotFPSAgentController : BaseFPSAgentController {
...
+            this.maxUpwardLookAngle = 30f + 1e-6f;
```
winthos, line 57: same for this one, should have no need for the epsilon, just the float is fine

unity/Assets/Scripts/LocobotFPSAgentController.cs
Lucaweihs: Great, done for both.

winthos (approved):
looks good! We will merge this in after NeurIPS

mattdeitke (approved):
LGTM! Feel free to merge after NeurIPS :)

PR #1023, opened 2022-05-09
Remove `Destroy` inside of GenerateMetadata

mattdeitke:
We've found rare cases where the `Controller` hangs forever if `Destroy` is called. Hanging occurs very inconsistently and after several millions of steps during training, making it extremely hard to debug.

This attempts to work around the use of the Destroy function here by reusing the same surrogateChild object if it has been created, otherwise, it creates a new one.

### DIFF
--- a/unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
+++ b/unity/Assets/Scripts/IK_Robot_Arm_Controller.cs
@@ -40,6 +40,8 @@ public partial class IK_Robot_Arm_Controller : MonoBehaviour {
 
     private const float extendedArmLength = 0.6325f;
 
+    private GameObject surrogateChild = null;
+
     public CollisionListener collisionListener;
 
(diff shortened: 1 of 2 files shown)


elimvb:
I hadn't thought a surrogate child would cause problems when I first made this metadata component, so I used Unity's automated transform-conversion-while-reorganizing-hierarchies as a quick-and-easy solution to find a "relative Quaternion". Now that I know this is prone to issues, I did some deeper research and found the equation I was really looking for: `Quaternion rootToSubjectRelativeOrientation = Quaternion.Inverse(rootObject.rotation) * subjectObject.rotation` Exact same output, no more surrogate children.

elimvb (commented):
Removed all traces of surrogate children. No longer necessary.

mattdeitke:
Awesome! That sounds good to me :) Great solution!

PR #1028, opened 2022-05-19
Rendering issue when gripper camera

AlvaroHG:
Fixes editor rendering when gripper and secondary cameras overwriting depth buffer.


winthos (approved):
good to go!

PR #1029, opened 2022-05-20
Fixes webgl build

AlvaroHG:
Updates cloudrendering with newtonsoft library dependency, and fixes property access of json object.


jonborchardt (approved):
Lg2m

PR #1032, opened 2022-06-09
Fix to `addAgents` for proc scenes.

Lucaweihs:
Ensures that depth of the agent camera is set in nanna houses.


unity/Assets/Scripts/AgentManager.cs
```diff
@@ -348,28 +348,30 @@ private enum serverTypes { WSGI, FIFO };
...
-            action.z = reachablePositions[i + 4].z;
-            addAgent(action);
-            yield return null; // must do this so we wait a frame so that when we CapsuleCast we see the most recently added agent
-        }
-        for (int i = 0; i < this.agents.Count; i++) {
-            this.agents[i].m_Camera.depth = 1;
-        }
-        this.agents[0].m_Camera.depth = 9999;
+    private void addAgents(ServerAction action) {
+        if (action.agentCount > 1) {
```
AlvaroHG, line 352: Why do we change it to a non-async function? what impact does this have?

Lucaweihs: I think the async functionality was there just to ensure that the transforms' positions/rotations were updated as we do some collision checks in `addAgents`. I believe this should be solved by using `Physics.SyncTransforms()` rather than waiting for a frame but let me know if not.

AlvaroHG:
This makes sense, just a question about the change from an async action. LGTM after that.

unity/Assets/Scripts/AgentManager.cs
AlvaroHG: Ok cool, yeah seems like that was the only purpose to wait till the end of frame this looks good.

PR #1033, opened 2022-06-22
MCS-1138  - Add Scene Option to Toggle All Lights Off or On at Specific Action Steps

alexcraneunity:
This introduces the option to add light-toggling into the JSON scene files, which causes the lights in the scene to be disabled at the given stepBegin, and re-enabled at the given stepEnd.  Multiple pairs of begin and end can be added to toggle the lights multiple times in a scene.


alexcraneunity:
Please disregard, this was created with the wrong base repository.

PR #1039, opened 2022-07-11
remove agentControllerType from ServerAction

winthos:
The `agentControllerType` member of the ServerAction class has not been used for a while, as all our agent modes now default to the correct physics/stochastic properties innately. This removes references to `agentControllerType` that were needlessly making the Initialize code harder to parse.

Currently the build fails with these changes as there is a test checking for a specific error message when the agentControllerType has a mismatch, however I believe this test should be altered since actually checking for this is effectively redundant now that the concept of a "controller type" has been integrated into the agent modes themselves.

### DIFF
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -165,91 +165,48 @@ public class AgentManager : MonoBehaviour {
         //"locobot" agentMode can use either default or "stochastic" agentControllerType
         //"drone" agentMode can ONLY use "drone" agentControllerType, and NOTHING ELSE (for now?)
         if (action.agentMode.ToLower() == "default") {
-            if (action.agentControllerType.ToLower() != "physics" && action.agentControllerType.ToLower() != "stochastic") {
-                Debug.Log("default mode must use either physics or stochastic controller. Defaulting to physics");
-                action.agentControllerType = "";
-                SetUpPhysicsController();
-            }
+            SetUpPhysicsController();
 
-            // if not stochastic, default to physics controller
-            if (action.agentControllerType.ToLower() == "physics") {
-                // set up physics controller
-                SetUpPhysicsController();
-            }
-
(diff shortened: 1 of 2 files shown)


mattdeitke:
Can you update the test?

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -165,91 +165,48 @@ private enum serverTypes { WSGI, FIFO };
...
+            physicsSceneManager.MakeAllObjectsMoveable();
+
+            if (action.massThreshold.HasValue) {
+                if (action.massThreshold.Value > 0.0) {
+                    SetUpMassThreshold(action.massThreshold.Value);
+                } else {
+                    var error = "massThreshold must have nonzero value - invalid value: " + action.massThreshold.Value;
+                    Debug.Log(error);
+                    primaryAgent.actionFinished(success: false, errorMessage: error);
+                    return;
```
Lucaweihs, line 208: It seems like lots of this code is heavily duplicated. Can we do something like:
```c#
string agentMode = action.agentMode.ToLower();
...
} else if (agentMode == "stretch" || agentMode == "arm") {
    if (agentMode == "stretch) {
        SetUpStretchController(action);
    } else if (agentMode == "arm") {
        SetUpArmController(true);
    } else {
        throw new InvalidArgumentsException();
    }
    ...
    PREVIOUSLY DUPLICATED CODE
    ....
}
```
Also, it seems like `action.massThreshold` shouldn't be restricted to use by the arm agents. Can we move it outside of these conditional statements so it just happens after initialization no matter which agent is chosen?

Lucaweihs (commented):
So happy this is removed. Added one suggestion to reduce code duplication.

Lucaweihs:
Any idea why this build is failing?

winthos:
closed as these changes have been synced with the nanna branch and will get merged into the main build from there

PR #1041, opened 2022-07-15
Nanna arm camera remove

AlvaroHG:
Removed Camera component from GripperCamera Prefab, disabling Component seemed to still render at least in Editor so removed after checking for any code references to this camera.
 
 Camera Parameters for reference when adding back: 
Screen Shot 2022-07-15 at 3 12 23 PM


Lucaweihs:
Oh interesting, it was still rendering even when disabled? Can we benchmark this on the python side? Perhaps run this build against the version before it was removed and a version where it's just disabled and compare FPS?

AlvaroHG:
I am not sure if it is rendering when disabled on the standalone release, reading about it should not render. In the editor it still shows a preview window for the camera even when disabled so I decided to go with the safer approach to the quick fix. I think we should merge this as the initial fix, and do the benchmarking for investigation since it's slightly unrelated and more about understanding Unity's general rules for camera components.

Lucaweihs (approved):
Sounds good to me!

PR #1043, opened 2022-07-19
Improvements to how timeouts are handled.

Lucaweihs:
We have code that times out the AI2-THOR unity process if it doesn't respond to a request in some number of seconds when using the WSGI server. Similar code did not seem to exist for the, now default, fifo server. This PR adds this timeout functionality (with tests) and adds a few small typing improvements.

### DIFF
--- a/ai2thor/controller.py
+++ b/ai2thor/controller.py
@@ -26,7 +26,7 @@ from platform import system  as platform_system
 from platform import architecture  as platform_architecture
 import uuid
 from functools import lru_cache
-
+from typing import Optional, Dict, Any, Union
 
 import numpy as np
 import ai2thor.wsgi_server
@@ -394,6 +394,7 @@ class Controller(object):
         server_class=None,
         gpu_device=None,
         platform=None,
+        server_timeout=100.0,
         **unity_initialization_parameters,
     ):
         self.receptacle_nearest_pivot_points = {}
@@ -402,6 +403,9 @@ class Controller(object):
         self.container_id = None
         self.width = width
         self.height = height
+
+        self.server_timeout = server_timeout
+        assert self.server_timeout is None or 0 < self.server_timeout
 
         self.last_event = None
         self.scene = None
@@ -561,7 +565,7 @@ class Controller(object):
             init_return = event.metadata["actionReturn"]
             if init_return:
                 self.server.set_init_params(init_return)
-                logging.info("Initialize return: {}".format(init_return))
+                logging.info(f"Initialize return: {init_return}")
 
     def _build_server(self, host, port, width, height):
 
@@ -577,20 +581,22 @@ class Controller(object):
 
         if self.server_class == ai2thor.wsgi_server.WsgiServer:
             self.server = ai2thor.wsgi_server.WsgiServer(
-                host,
+                host=host,
+                timeout=self.server_timeout,
                 port=port,
-                depth_format=self.depth_format,
-                add_depth_noise=self.add_depth_noise,
                 width=width,
                 height=height,
+                depth_format=self.depth_format,
+                add_depth_noise=self.add_depth_noise,
             )
 
         elif self.server_class == ai2thor.fifo_server.FifoServer:
             self.server = ai2thor.fifo_server.FifoServer(
-                depth_format=self.depth_format,
-                add_depth_noise=self.add_depth_noise,
                 width=width,
                 height=height,
+                timeout=self.server_timeout,
+                depth_format=self.depth_format,
+                add_depth_noise=self.add_depth_noise,
             )
 
     def __enter__(self):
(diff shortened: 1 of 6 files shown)


ai2thor/controller.py
```diff
@@ -960,18 +968,27 @@ def step(self, action=None, **action_args):
...
+                f"Restarting unity due to crash when when running action {action}"
+                f" in scene {self.last_event.metadata['sceneName']}:\n{traceback.format_exc()}"
+            )
             warnings.warn(message)
             self.start(width=self.width, height=self.height, x_display=self.x_display)
             self.reset()
             raise RestartError(message)
+        except Exception as e:
+            self.server.stop()
+            raise (TimeoutError if isinstance(e, TimeoutError) else RuntimeError)(
```
AlvaroHG, line 988: Would it be more clear if we had an `except` branch for `TimeoutError`?

ai2thor/controller.py
```diff
@@ -1335,8 +1352,13 @@ def start(
             unity_params = self.server.unity_params()
             self._start_unity_thread(env, width, height, unity_params, image_name)
 
-        # receive the first request
-        self.last_event = self.server.receive()
+        # receive the first request, give unity at least 5 minutes to start
+        self.last_event = self.server.receive(
+            timeout=max(
+                60.0 * 5,
```
AlvaroHG, line 1358: Maybe have a variable for this `max_timeout_seconds` ?

ai2thor/controller.py
```diff
@@ -1335,8 +1352,13 @@ def start(
...
+                self.server_timeout * 5 if self.server_timeout is not None else 0.0
```
AlvaroHG, line 1359: Why is this multiplied by `5` wouldn't it be better for the value to be the actual `timeout_value` so there were not unexpected behaviors?

ai2thor/fifo_server.py
```diff
@@ -93,19 +101,52 @@ def __init__(
...
 
-    def _recv_message(self):
+    def _read_with_timeout(self, server_pipe, message_size: int, timeout: Optional[float]):
+        if timeout is None:
+            return server_pipe.read(message_size)
+
+        start_t = time.time()
+        message = b""
+
+        while message_size > 0:
```
AlvaroHG, line 122: Is this way of reading with the while loop and directly from the file descriptor any slower? I don't think it should but an option could be if `server_timeout` is `None` then we could use the old function.

ai2thor/wsgi_server.py
```diff
@@ -27,11 +33,16 @@
...
 # get with timeout to allow quit
-def queue_get(que, unity_proc=None):
+def queue_get(que, unity_proc=None, timeout: Optional[float] = 100.0):
     res = None
     attempts = 0
-    max_attempts = 200
+    max_attempts = (
+        float("inf")
+        if timeout is None or timeout == float("inf") else
+        max(int(math.ceil(timeout / 0.5)), 1)
```
AlvaroHG, line 44: A line is a bit hard to read, any reason for ` / 0.5` vs ` * 2`, making it floating point?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1987,6 +1987,16 @@ abstract public class BaseFPSAgentController {
             actionFinished(true);
         }
 
+        protected IEnumerator waitForSecondsRealtime(int seconds) {
```
AlvaroHG, line 1990: Don't see this being referenced anywhere

AlvaroHG:
A couple of comments this is great! LGTM after that

ai2thor/controller.py
Lucaweihs: I agree regarding clarity but I'd prefer to have a single branch here as, otherwise, we would be duplicating the same `self.server.stop()` etc code twice.

ai2thor/controller.py
Lucaweihs: Decided to just create a `server_start_timeout` parameter. I was trying to keep the number of initialization parameters low but this was probably unnecessary.

> Also seems like the if condition will never be true because there is an assert for self.server_timeout is None in the constructor
The assert you're talking about checks that `self.server_timeout is None` or that it's positive.

ai2thor/fifo_server.py
Lucaweihs: That's already there if look at lines 116-117.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: It's being used by the new `Sleep` action which is used by the new tests that check that make sure the timeout code works properly.

ai2thor/wsgi_server.py
Lucaweihs: I've made this a bit clearer, the idea is that the 0.5 references the 0.5 used by the `que.get(...)` call a few lines layer.

PR #1046, opened 2022-07-25
Adjusted stretch arm default pose to be more compact

elimvb:
Pretty simple change. The stretch arm's pos-rot now starts at a default position of 0.1m up, and 180 y-degrees left, so that it won't collide with anything that the agent's body wouldn't as well, removing it as a liability for tight spaces
Before_After
.


mattdeitke:
Can you take a picture of the default bot before and after this change?

winthos:
looks good to go!

winthos (approved):

ehsanik (approved):

mattdeitke (approved):

PR #1048, opened 2022-07-26
Removing agentControllerType initialization parameter on nanna branch.

Lucaweihs:
Moves #1039 to the nanna branch. Hopefully also fixes the failing tests.


winthos (approved):
good to go!

PR #1056, opened 2022-08-11
Nanna explode lw

Lucaweihs:


unity/Assets/Editor/AutoSimObject.cs
```diff
@@ -34,10 +35,12 @@ public class AutoSimObject : EditorWindow {
       // instantiate the prefab
       var prefab = AssetDatabase.LoadAssetAtPath<GameObject>(basePath + modelId + "/model.obj");
       GameObject obj = GameObject.Instantiate(prefab);
+      obj.transform.rotation = Quaternion.identity;
```
mattdeitke, line 38: Why is this here? obj.transform.rotation is overwritten later.

PR #1060, opened 2022-08-17
New distance functions and improvements to bbox / magnet sphere code.

Lucaweihs:
This PR makes the following changes:

1. Fixes bug where, when using cloud rendering, we were incorrectly matching CUDA GPU indices to Vulkan GPU indices.
2. Adds an additional field (`touchedNotHeldObjects`) to the arm metadata which corresponds to a list of object ids that are currently being touched, but are not currently being held, by the arm's magnet sphere.
3. Adds a `GetObjectMetadata` action that returns the metadata for the objects corresponding to a given list of object ids.
4. Adds a `GetVisibleObjects` action that returns a list of object ids of visible objects (given a max visible distance and visibility scheme).
5. Fixes a bug where the `GetAllVisibleSimObjPhysicsDistance` function was ignoring the `maxDistance` argument.
6. Adds a `BBoxDistance` action that computes the (approximate) distance between two objects using their bounding boxes. This should be far more accurate than using the distance between the centerpoints of objects (e.g. the distance between an apply siting on the side of a couch would be large when considering centerpoints but will be near 0 when computing the distance between bounding boxes).
7. Adds `CheckUnobstructedPathBetweenObjectCenters` action that does what it says (sees if you can send a raycast between two objects' center points without hitting any other object). This is a nice way to check if two objects are actually side-by-side when they are nearby (e.g. imagine two objects on two sides of a wall, they can be close in L2 distance but are not actually "nearby" in a meaningful sense).
8. Adds a `CheckWhatObjectsOn` action that returns the object id of the object that a given object is sitting on. Useful when the receptacle bounding box is not usable (e.g. imagine a box that's been turned on its side with an apple sitting on it, the apple is "on" the box but the receptacle trigger box will, correctly, be empty but we'd still like to know that the apple is on the box in some settings).
9. Simplified the code in `WhatIsInsideMagnetSphere.cs`.


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -1542,20 +1542,21 @@ abstract public class BaseFPSAgentController {
...
-        public virtual ObjectMetadata[] generateObjectMetadata() {
-            SimObjPhysics[] simObjects = null;
-            if (this.simObjFilter != null) {
-                simObjects = this.simObjFilter;
-            } else {
-                simObjects = GameObject.FindObjectsOfType<SimObjPhysics>();
+        public virtual ObjectMetadata[] generateObjectMetadata(SimObjPhysics[] simObjects = null) {
+            if (simObjects == null) {
+                if (this.simObjFilter != null) {
+                    simObjects = this.simObjFilter;
```
AlvaroHG, line 1548: This is a little surprising, passing null and getting metadata. Could be good for this function to just work with the simObjects it's passed, but I see the filtering needs to happen somewhere.

ai2thor/controller.py
```diff
@@ -1029,12 +1029,86 @@ def unity_command(self, width, height, headless):
...
+                        raise RuntimeError(
+                            "vulkaninfo failed to run, please ask your administrator to"
+                            " install `vulkaninfo` (e.g. on Ubuntu systems this requires running"
+                            " `sudo apt install vulkan-tools`)."
+                        )
+
+                    current_gpu = None
+                    device_uuid_to_vulkan_gpu_index = {}
+                    for l in vulkan_result.stdout.splitlines():
+                        gpu_match = re.match("GPU([0-9]+):", l)
```
AlvaroHG, line 1059: Is it always 0-9 ?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2928,6 +2945,45 @@ abstract public class BaseFPSAgentController {
             }
         }
 
+        public void GetVisibleObjects(float? maxDistance = null, string visibilityScheme = null) {
+            VisibilityScheme visSchemeEnum;
+            if (visibilityScheme != null) {
+                visibilityScheme = visibilityScheme.ToLower();
+                if (visibilityScheme.ToLower() == "collider") {
```
AlvaroHG, line 2952: Change string literal comparison to -> `Enum.GetName(typeof(VisibilityScheme), VisibilityScheme.Collider, ignoreCase: true) == visibilityScheme `

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -2928,6 +2945,45 @@ abstract public class BaseFPSAgentController {
...
+                    visSchemeEnum = VisibilityScheme.Collider;
+                } else if (visibilityScheme.ToLower() == "distance") {
```
AlvaroHG, line 2954: Same as above comment

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -3516,6 +3572,265 @@ bool includeInvisible
...
+                        List<float> zVals = whichZ == 1 ? zMinMax : zCenterVals;
+
+                        if (whichX + whichY + whichZ == 0) {
+                            continue;
+                        }
+
+                        # if UNITY_EDITOR
+                        Vector3? lastPoint = null;
+                        # endif
+                        foreach (float x in xVals) {
```
AlvaroHG, line 3611: For future reference you can  do `new List<float> {xVals, yVals, zVals}.CartesianProduct().Select(p => bc.center + new Vector3(p[0], p[1], p[2])` defined in our `Extensions.cs`. But I think we should keep it as it is because it's probably the fastest.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -3516,6 +3572,265 @@ bool includeInvisible
...
+            BoxCollider c1 = sop1.BoundingBox.GetComponent<BoxCollider>();
+
+            float dist = float.PositiveInfinity;
+
+            c1.enabled = true;
+            foreach (Vector3 p in pointsOnSurfaceOfBoxCollider(c0, divisions)) {
+                Vector3 pLocal = c1.transform.InverseTransformPoint(p) - c1.center;
+                Vector3 size = c1.size;
+                if (
+                    (-0.5f * size.x < pLocal.x && pLocal.x < 0.5f * size.x) &&
```
AlvaroHG, line 3653: What does constant 0.5 mean perhaps create named constant?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -3516,6 +3572,265 @@ bool includeInvisible
...
+                }
+                Vector3 closestP = c1.ClosestPoint(p);
+# if UNITY_EDITOR
+                Debug.DrawLine(p, closestP, Color.red, 10.0f);
+# endif
+                dist = Mathf.Min(dist, Vector3.Distance(p, closestP));
+            }
+            c1.enabled = false;
+
+            c0.enabled = true;
```
AlvaroHG, line 3672: What does enabling/disabling each do?

unity/Assets/Scripts/SimObjPhysics.cs
```diff
@@ -168,7 +176,17 @@ public class PhysicsMaterialValues {
         ContainedObjectReferences.Remove(t);
     }
 
-    public void syncBoundingBoxes(bool forceCacheReset = false) {
+    public void syncBoundingBoxes(
```
AlvaroHG, line 179: This name seems a misnomer if I understand correctly? It's not syncing the boxes just setting flags to do so at some point?

AlvaroHG:
Could you add a description to the PR for future reference? like what problem you're trying to solve and how it solved it/what is covered.

Some minor comments, LGTM after that

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: `generateObjectMetadata` now has no default and will throw an error if passed a null value.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Results in an error:
```
Assets/Scripts/BaseFPSAgentController.cs(2956,107): error CS1739: The best overload for 'GetName' does not have a parameter named 'ignoreCase'
```

Will do
```
Enum.GetName(typeof(VisibilityScheme), VisibilityScheme.Collider).ToLower() == visibilityScheme
```
instead.

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: Very cool 👍

unity/Assets/Scripts/BaseFPSAgentController.cs
Lucaweihs: If the colliders aren't enabled when running functions that do physics checks (namely `ClosestPoint` in this code) then we will get incorrect results (this is a hard thing to debug as, iirc, it doesn't throw an error).

unity/Assets/Scripts/ExpRoom/PhysicsRemoteFPSAgentController_partial_ExpRoom.cs
```diff
@@ -372,16 +372,36 @@ Camera visibilityCheckCamera
 
             List<Vector3> closePoints = new List<Vector3>();
             foreach (Collider c in target.GetComponentsInChildren<Collider>()) {
-                if (c.enabled && !c.isTrigger) {
+                if (
+                    c.enabled && !c.isTrigger && (
+                        c is BoxCollider ||
+                        c is SphereCollider ||
+                        c is CapsuleCollider ||
+                        (c is MeshCollider && ((MeshCollider) c).convex)
```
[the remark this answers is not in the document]
Lucaweihs: The `ClosestPoint` function only works on these types of colliders (it fails on nonconvex mesh colliders). Note that we fallback to using an object's bounding box if it has no such colliders. I'll add a comment.

unity/Assets/Scripts/SimObjPhysics.cs
Lucaweihs: I don't think so, it will actually regenerate the bounding boxes so long as:
1. The object has moved/rotated or
2. We explicitly pass the `forceCacheReset` argument.

In this PR I added another argument to force the creation of an object-oriented bounding box as sometimes these are not created for certain objects..

Lucaweihs:
@AlvaroHG Made the requested changes and added a description.

PR #1099, opened 2022-10-24
Adding additional logging to the CI build script and using more explicit communication between processes.

Lucaweihs:
The current build script doesn't output very much information about what did, or did not, fail and in how much time. This PR adds some additional logging and should fail faster when appropriate.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -12,6 +12,9 @@ import shutil
 import subprocess
 import pprint
 import random
+from queue import Empty
+from typing import Optional
+
 from invoke import task
 import boto3
 import botocore.exceptions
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -1089,42 +1098,56 @@ def ci_build(context):
...
+                        has_any_build_failed = has_any_build_failed or not build_success
+                        if build_success:
+                            logger.info(
+                                f"Build success detected for {arch} {build['commit_id']}"
+                            )
+                        else:
+                            logger.error(f"Build failed for {arch}")
+                            p.kill()
+
+                        p.join(10.0)
```
AlvaroHG, line 1132: Is this a 10 second timeout, is this enough?

AlvaroHG:
LGTM!

tasks.py
Lucaweihs: I think it's enough but we can increase if it ends up being too short.

PR #1161, opened 2023-09-13
Dislodgement check

AlvaroHG:
- Checks for finger dislodgement by using local transform of original state
- New `Error` AgentState for Emit state machine

- Remaining bug, reset does not seem to work properly for AB agent after having done some AB actions


winthos:
looks good!

AlvaroHG:
Merging to simplify other Physics Det PR

PR #1162, opened 2023-09-13
Improved Physics Determinism

AlvaroHG:
Features:
- Adds `ActionFinished` as return type for actions
- New dispatch mechanism to manually simulate physics or run in a coroutine
- `PhysicsSimulationParams` object that can be passed to any actions to govern physics execution, autosimulation, fixedDeltaTime and max time to run an action
- Run regular actions with physics simulation padding
- Backwards compatible dispatch for void actions that call `actionFinished`
- Refactor of arguments of arm actions
- Adds Exception handling for Coroutine actions
- Adds support for the Fifo controller to work with the editor

Bugfixes:
- Makes sure it chooses children overwriten actions in child `ActionInvokables`, over parents when ranked the same.
- Fixes `TestShortest` path unity test that was timing out the builds since it was changed to call `actionFinishedEmit`

TODO: Fix hierarchy for ambiguous actions


Lucaweihs:
Functionality is now in.

PR #1187, opened 2024-02-21
Added big overhaul to get metallic-smoothness to import into THOR

elimvb:
@AlvaroHG  Please point out any additions I made for this specific push involving the addition of "MetallicSmoothness" that are NOT necessary


unity/Assets/Scripts/ProceduralTools.cs
```diff
@@ -2437,7 +2438,21 @@ CollisionDetectionMode collisionDetectionMode
...
+                }
+                runtimePrefab.metallicSmoothnessTexturePath = metallicSmoothnessTexturePath;
+                mat.EnableKeyword("_METALLICGLOSSMAP");
+                byte[] imageBytes = File.ReadAllBytes(metallicSmoothnessTexturePath);
+                Texture2D tex = new Texture2D(2, 2);
+                tex.LoadImage(imageBytes);
+                
+                mat.SetTexture("_MetallicGlossMap", tex);
+            } else {
+                mat.SetFloat("_Glossiness", 0f);
```
AlvaroHG, line 2454: Is this right? what does this do?

AlvaroHG:
Great! had one question but LGTM

unity/Assets/Scripts/ProceduralTools.cs
elimvb: If It means that if the json's metallicSmoothnessTexturePath (i.e. the map) exists, apply it to the Metallic_GlossMap channel (the under-the-hood name for MetallicSmoothness). If it doesn't, set the material's Glossiness (i.e. Smoothness) to 0. I suppose I should also add a "Set Metallic to 0 if no MetallicSmoothness map exists" line as well, though.

PR #1188, opened 2024-02-23
Rendering an object for given angles

Lucaweihs:
Adding code to let us more easily render an object from multiple angles at runtime.

### DIFF
--- a/unity/Assets/Scripts/BaseFPSAgentController.cs
+++ b/unity/Assets/Scripts/BaseFPSAgentController.cs
@@ -7403,6 +7403,101 @@ namespace UnityStandardAssets.Characters.FirstPerson {
                 });
         }
 
+        // Helper method to calculate the required distance of the camera based on the object size and camera's FOV
+        protected float CalculateCameraDistance(Camera camera, float objectSize) {
+            float cameraFOV = camera.fieldOfView;
+            float cameraAspect = camera.aspect;
+            float distance = objectSize / (2f * Mathf.Tan(0.5f * cameraFOV * Mathf.Deg2Rad));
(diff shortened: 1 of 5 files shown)


unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -7405,6 +7405,193 @@ ProceduralAsset asset
...
+            renderCamera.transform.position = originalPosition;
+            renderCamera.transform.rotation = originalRotation;
+            renderCamera.targetTexture = originalRenderTexture;
+            RenderTexture.active = null;
+            RenderTexture.Destroy(renderTexture);
+
+            actionFinished(byte_arrays != null, byte_arrays);
+        }
+
+        public void RenderObjectFromAngles(
```
AlvaroHG, line 7573: In the new system you can just make the return type `IEnumerator` and return a `yield new ActionFinished(...)` at some point and the Coroutine will be started for you! so no need to lauch a Coroutine Manually (which can lead to bugs performance issues)

AlvaroHG:
LGTM

unity/Assets/Scripts/BaseFPSAgentController.cs
AlvaroHG: In this case you can also return a call to your other function which is an `IEnumerator` `renderObjectFromAngles`

Lucaweihs: From what I could tell, this ended up being more complicated than just leaving it as it is (C# doesn't seem to have a nice way to yield from another function call). Kept as it is for now.

PR #1189, opened 2024-03-02
add SpawnBoxCollider func

KuoHaoZeng:
The newly added SpawnBoxCollider function generates a box with an active collider encapsulates the agent. Currently, it only supports the Stretch agent.

New augments:
1. `useFPINCollider`: bool. When set to True, SpawnBoxCollider is included during the initialization phase. If False, initialization proceeds as previously.
2. `colliderScaleRatio`: dict. It specifies the scale ratio of the box along each axis, for example, `{"x": 1.0, "y": 1.0, "z": 1.0}`. A value of 1.0 indicates that the box size matches the detected bounds.

### DIFF
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -288,6 +288,9 @@ public class AgentManager : MonoBehaviour, ActionInvokable {
         BaseAgentComponent baseAgentComponent = GameObject.FindObjectOfType<BaseAgentComponent>();
         primaryAgent = createAgentType(typeof(StretchAgentController), baseAgentComponent);
         baseAgentComponent.StretchBodyColliders.SetActive(true);
+        if (action.useFPINCollider) {
+            primaryAgent.SpawnBoxCollider(primaryAgent.gameObject, action.colliderScaleRatio);
+        }
     }
 
     private void SetUpStretchABController(ServerAction action) {
@@ -2135,7 +2138,8 @@ public class ServerAction {
     public Vector3 rotation;
     public Vector3 position;
     public Vector3 direction;
-
+    public Vector3 colliderScaleRatio;
+    public bool useFPINCollider;
     public bool allowAgentsToIntersect = false;
     public float handDistance;// used for max distance agent's hand can move
     public List<Vector3> positions = null;
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -1557,6 +1557,8 @@ public class AgentMetadata {
     public bool? isStanding = null;
 
     public bool inHighFrictionArea;
+
+    public Vector3 colliderSize;
     public AgentMetadata() { }
 }
 
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -2192,6 +2192,7 @@ public class ServerAction {
     public Vector3 direction;
     public Vector3 colliderScaleRatio;
     public bool useAbsoluteSize = false;
+    public bool useVisibleColliderBase = true;
     public bool allowAgentsToIntersect = false;
     public float handDistance;// used for max distance agent's hand can move
     public List<Vector3> positions = null;
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -2191,6 +2191,7 @@ public class ServerAction {
     public Vector3 position;
     public Vector3 direction;
     public Vector3 colliderScaleRatio;
+    public bool useAbsoluteSize = false;
     public bool allowAgentsToIntersect = false;
     public float handDistance;// used for max distance agent's hand can move
     public List<Vector3> positions = null;
--- a/unity/Assets/Scripts/AgentManager.cs
+++ b/unity/Assets/Scripts/AgentManager.cs
@@ -271,6 +271,9 @@ public class AgentManager : MonoBehaviour, ActionInvokable {
         action.snapToGrid = false;
         BaseAgentComponent baseAgentComponent = GameObject.FindObjectOfType<BaseAgentComponent>();
         primaryAgent = createAgentType(typeof(LocobotFPSAgentController), baseAgentComponent);
+        if (action.useFPINCollider) {
+            primaryAgent.SpawnBoxCollider(primaryAgent.gameObject, typeof(LocobotFPSAgentController), action.colliderScaleRatio);
+        }
     }
 
     private void SetUpDroneController(ServerAction action) {
@@ -289,7 +292,7 @@ public class AgentManager : MonoBehaviour, ActionInvokable {
         primaryAgent = createAgentType(typeof(StretchAgentController), baseAgentComponent);
         baseAgentComponent.StretchBodyColliders.SetActive(true);
         if (action.useFPINCollider) {
-            primaryAgent.SpawnBoxCollider(primaryAgent.gameObject, action.colliderScaleRatio);
+            primaryAgent.SpawnBoxCollider(primaryAgent.gameObject, typeof(StretchAgentController), action.colliderScaleRatio);
         }
     }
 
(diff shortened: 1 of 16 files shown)


KuoHaoZeng:
### Update

- I added `UpdateAgentBoxCollider` with an argument `colliderScaleRatio=vector3` to spawn a new box collider. It will spawn a box collider according to agent's body (after destroy the pre-existing one)

- On the other hand, `DestroyAgentBoxCollider` will just removes the existing box collider.

- No `UseFPINCollider` augment anymore!

- Nothing happens with box collider functionality anymore during the initialization or reset.

- To adjust the collider, we must explicitly call `UpdateAgentBoxCollider` or `DestroyAgentBoxCollider`

- Add `imageSynthesis.OnCameraChange()` at the end of `updateCameraProperties` to solve the segmentation mask issue with the change of FoV.

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -456,24 +458,46 @@ private enum serverTypes { WSGI, FIFO };
...
+            camera.transform.localScale = Vector3.one;
         } else {
-            //makes sure camera is child of agent, and reposition in agent relative space
-            if(camera.transform.parent != primaryAgent.transform) {
-                camera.transform.SetParent(primaryAgent.transform);
-                //for some reason reparenting sometimes makes the scale slightly off...
-                //so we are gonna force it to be uniform just in case cause floats suck
-                camera.transform.localScale = Vector3.one;
+            if (position.HasValue) {
+                camera.gameObject.transform.position = position.Value;
```
AlvaroHG, line 484: This is fine but a thing to consider: there are a lot of if else statements that end with the same code like assigning the position. 
I may try refactoring so that only those lines exist once so that is easier to debug, as much as possible.

You can try using the `??=` operator to optionally (assign a value that is null)[https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator#examples]

unity/Assets/Scripts/AgentManager.cs
```diff
@@ -528,6 +552,11 @@ private enum serverTypes { WSGI, FIFO };
             );
         }
 
+        ImageSynthesis imageSynthesis = camera.gameObject.GetComponentInChildren<ImageSynthesis>();
```
AlvaroHG, line 555: 👍

unity/Assets/Scripts/StretchAgentController.cs
```diff
@@ -268,6 +268,73 @@ public partial class StretchAgentController : ArmAgentController {
...
+//                gimbalBase.localEulerAngles = new Vector3(
+//                    gimbalBaseStartingXRotation + rollDegrees,
+//                    gimbalBaseStartingYRotation + yawDegrees,
+//                    gimbalBase.transform.localEulerAngles.z
+//                );
+//            }
+//            actionFinished(true);
+//        }
+
+        public void RotateCameraMount(float degrees, bool secondary = false) {
```
AlvaroHG, line 293: Do we need this funtionality on this branch?

AlvaroHG:
Nothing jumps out, and looks good. But what I recommend to make sure things work as expected, and easier for the reviewers to understand how the code should work is creating some unit tests :)
LGTM after some minor comments

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -7275,6 +7223,110 @@ ProceduralAsset asset
...
+            } else if (agentType == typeof(StretchAgentController)) {
+                meshRenderers = this.baseAgentComponent.StretchVisCap.GetComponentsInChildren<MeshRenderer>();
+            }
+            foreach (MeshRenderer meshRenderer in meshRenderers) {
+                bounds.Encapsulate(meshRenderer.bounds);
+            }
+            return bounds;
+        }
+
+        public void SpawnBoxCollider(GameObject agent, Type agentType, Vector3 scaleRatio, bool useAbsoluteSize = false, bool useVisibleColliderBase = false) {
```
winthos, line 7242: Since this is now a helper function to help spawn the agent colliders, I'd change the name to be a bit more descripting and also lead with a lowercase so its not confused with an action that should be called directly

Something like `spawnAgentBoxCollider`

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -7275,6 +7223,110 @@ ProceduralAsset asset
...
+                visibleBox.name = "VisibleBox";
+                visibleBox.transform.position = new Vector3(bounds.center.x, 0.05f, bounds.center.z);
+                visibleBox.transform.localScale = colliderSize;
+                visibleBox.transform.parent = agent.transform;
+            }
+
+            this.transform.rotation = originalRotation;
+        }
+
+        public void DestroyAgentBoxCollider(ServerAction action){
```
winthos, line 7306: I would replace the `ServerAction action` parameter into the more explicit syntax we have adopted that lists the exact parameters used to avoid passing in the entire ServerAction class.

This one doesn't appear to actually need any parameters so just leaving it with no parameters should be fine.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -7275,6 +7223,110 @@ ProceduralAsset asset
...
+                GameObject.Destroy(triggeredEncapsulatingBox);
+            }
+            if (visibleBox != null) {
+                GameObject.Destroy(visibleBox);
+            }
+            actionFinished(true);
+            return;
+        }
+
+        public void UpdateAgentBoxCollider(ServerAction action) {
```
winthos, line 7323: Also here I would replace the ServerAction action parameter into the more explicit syntax, so something like 

```public void UpdateAgentBoxCollider(Vector3 colliderScaleRatio, bool useAbosluteSize, bool useVisibleColliderBase)```

Similarly I would remove these added members of the ServerAction class as to not bloat it even more since we are trying to move away from having to use it.

**Review by winthos — CHANGES_REQUESTED** [review]
Looking good! We have a few syntax things to clean up and adding in some new extra collision checks via box colliders that should be done as part of the collider generation steps.

File: unity/Assets/Scripts/StretchAgentController.cs (hunk 3 above)
Comment (KuoHaoZeng) [reply unity/Assets/Scripts/StretchAgentController.cs:293]: it shows up here probably because I merged the code from `physics-det-new-cam-adjust`
shall I remove this?

Comment (AlvaroHG) [reply unity/Assets/Scripts/StretchAgentController.cs:293]: I see I think we can remove it

File: unity/Assets/Scripts/BaseFPSAgentController.cs [hunk 7]
```diff
@@ -7275,6 +7223,133 @@ ProceduralAsset asset
...
+            return bounds;
+        }
+
+        public void spawnAgentBoxCollider(GameObject agent, Type agentType, Vector3 scaleRatio, bool useAbsoluteSize = false, bool useVisibleColliderBase = false) {
+            // Store the current rotation
+            Vector3 originalPosition = this.transform.position;
+            Quaternion originalRotation = this.transform.rotation;
+
+            // Move the agent to a safe place and align the agent's rotation with the world coordinate system
+            this.transform.position = new Vector3(originalPosition.x + 100, originalPosition.y + 100, originalPosition.z + 100);
```
AlvaroHG, line 7248: Would store the `100` in a constant variable for the class, and use it here and below, maybe a Vector3 `agentSpawnOffset` ?

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -7275,6 +7223,160 @@ ProceduralAsset asset
...
+                scaleRatio.y * bounds.size.y,
+                scaleRatio.z * bounds.size.z
+            );
+            if (useAbsoluteSize) {
+                colliderSize = scaleRatio;
+            }
+
+            GameObject noneTriggeredEncapsulatingBox = new GameObject("NonTriggeredEncapsulatingBox");
+            // noneTriggeredEncapsulatingBox.transform.position = new Vector3(bounds.center.x, bounds.center.y, agent.transform.position.z);
+            noneTriggeredEncapsulatingBox.transform.position = bounds.center;
```
AlvaroHG, line 7310: This is set to `bounds.center` which computed using agent position  with the added +100, wouldn't we want `newBoxCenter` here which removes the offset?

Same comment for below's `triggeredEncapsulatingBox` position.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -7275,6 +7223,160 @@ ProceduralAsset asset
...
+            // Get the agent's bounds
+            var bounds = GetAgentBounds(agent, agentType);
+
+            Debug.Log($"the global position of the agent bounds is: {bounds.center:F8}");
+
+            // Check if the spawned boxCollider is colliding with other objects
+            int layerMask = LayerMask.GetMask("SimObjVisible", "Procedural1", "Procedural2", "Procedural3", "Procedural0");
+            
+            Vector3 newBoxCenter = new Vector3(bounds.center.x - 100f, bounds.center.y - 100f, bounds.center.z - 100f);
+            Debug.Log($"the center for the new box should be at the agent's original position but is: {newBoxCenter:F8}");
```
AlvaroHG, line 7266: Not necessarily true `Bounds.Encapsulate` might find a more fitting box center after adding the mesh geo to encapsulate.

PR #1200, opened 2024-03-23
Adding TryReachObject and GetTouchingPoses actions to the stretch age…

Lucaweihs:
…nt (useful for finding positions from which an object is graspable). Also fixed a bug in the wrist rotation action.


Lucaweihs:
@AlvaroHG @winthos @elimvb All tests passed with a local build, merged.

PR #1230, opened 2024-08-09
Fixing collision checks to work correctly with convex colliders.

Lucaweihs:
The old `isObjectColliding` function did not check for mesh colliders as all dynamic objects previously used only primitive colliders. This changed with Objaverse and so this function needed to be updated.

This commit also includes a change to how the skybox is setup when a procedural house is loaded with a uniform color skybox. The idea here is that we don't want the skybox to effect lighting in the scene as this can result in some really weird behavior (i.e. objects rendered on a pure white background are very overexposed).


AlvaroHG:
Looks good!

PR #1231, opened 2024-08-12
Removing `Collider-Based` visibility scheme option

winthos:
Removes all references and use of the collider based visibility scheme used to query sim objects from cameras. Moving forward this scheme will not be supported, and all visibility checks will now be completely distance-based from the camera (ie: objects will only be potentially visible if they are within maxVisibleDistance from the camera based purely by raycast distance)

Additionally, updating the function to query objects from a camera to be called `GetVisibleObjectsFromCamera` since the old name `GetVisibleObjects` was too easily confused with other object-visibility helper functions. The old`GetVisibleObjects` action will remain but will have a deprecation warning message thrown for those who may still use it.

Finally, added unit tests for`GetVisibleObjectsFromCamera`


AlvaroHG:
Ok one request, can you throw an exception on `AgentManger.cs` `Initialize` explicitly saying `Deprecated parametar 'collider' for visibility, only ... supported` so that it clear. 

LGTM after that

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -4547,66 +4538,90 @@ protected SimObjPhysics[] GetAllVisibleSimObjPhysics(
...
-                        $"Visibility scheme {visibilityScheme} is not implemented. Must be 'distance' or 'collider'."
+                        $"Visibility scheme {this.visibilityScheme} is not implemented for third party cameras. Default visibility scheme should be 'Distance'."
                     );
                 }
-            } else {
-                visSchemeEnum = this.visibilityScheme;
             }
-            return visSchemeEnum;
+
+            //only check visibility for objects with these ids otherwise check them all
```
elimvb, line 4570: "otherwise check them all" has no code-based follow-through. However, I'm not sure you'd ever want to check everything, since it goes against the entire purpose of the function. For that reason, I'd alter the comment rather than the code.

elimvb (commented):
Pretty solid. Just a few minor changes.

unity/Assets/Scripts/BaseFPSAgentController.cs
```diff
@@ -4547,69 +4538,95 @@ protected SimObjPhysics[] GetAllVisibleSimObjPhysics(
...
+            else {
                 camera = m_Camera;
             }
 
+            if (this.visibilityScheme != VisibilityScheme.Distance) {
+                throw new System.NotImplementedException(
+                    $"Visibility scheme {this.visibilityScheme} is not compatible. Visibility scheme should be 'Distance'."
+                );
+            }
+
```
Lucaweihs, line 4629: This is a bug. Someone can pass in `visibilityScheme="Collider"` and no error would be thrown because we're referencing `this.visibilityScheme` and not the local `visibilityScheme` parameter.

My recommendation would be to completely remove the `visibilityScheme` argument to this function. This will cause peoples' code to break if they specified it before but that's ok.

unity/Assets/UnitTests/TestThirdPartyCameraAndMainCamera.cs
```diff
@@ -191,6 +191,67 @@ public IEnumerator TestUpdateThirdPartyCamera()
...
+#if UNITY_EDITOR
+            foreach (string obj in visibleObjects)
+            {
+                Debug.Log(obj);
+            }
+#endif
+
+            Assert.AreEqual(visibleObjects.Count, 1);
+            Assert.AreEqual(visibleObjects[0], "Apple|-00.47|+01.15|+00.48");
+        }
```
Lucaweihs, line 253: Please also add a test for the main camera.

Lucaweihs (changes requested):

PR #1234, opened 2024-08-14
`ChangeFOV` changes fov of segmentation, depth, etc.

AlvaroHG:
`ChangeFOV` action did not change fov for renders other than `RGB` on main camera.


winthos (changes requested):
Besides the one exception looks good to go

winthos (approved):

PR #1241, opened 2024-09-25
Updates to Teleport/TeleportFull and isStanding()

winthos:
The purpose of these updates is to prevent an edge case where, when using an arm-typed agent, calling the Teleport action would cause the isStanding value to default in a weird way, causing the main camera of the arm-typed agent to be changed unexpectedly back to a default position that only has context for the high-level-agent. This update also catches a case where, when using the high-level agent and calling `UpdateMainCamera`, the `isStanding() return became ambiguous.

- adding an override to the Teleport/TeleportFull actions in the `ArmAgentController` class and its derived classes so that an error can be thrown if passing in the `standing` bool since arm agents don't have a concept of a standing vs crouching camera position

- updating the isStanding() helper function to return `null` in cases where the High Level Agent's main camera is not in either the hard-coded standing or crouching local position.

### DIFF
--- a/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
+++ b/unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
@@ -173,24 +173,28 @@ namespace UnityStandardAssets.Characters.FirstPerson {
             return;
         }
 
-        public bool isStanding() {
+        public bool? isStanding() {
 
             //default to not standing if this isn't literally the PhysicsRemoteFPSAgentController
             //this means the metadat for isStanding should always be false for all derived classes
             if (this.GetType() != typeof(PhysicsRemoteFPSAgentController))
             {
+                return null;
+            }
+
+            //if camera is in neither the standing or crouching predetermined locations, return null
+            if(UtilityFunctions.ArePositionsApproximatelyEqual(m_Camera.transform.localPosition, standingLocalCameraPosition) == true) {
+                return true;
+            } 
+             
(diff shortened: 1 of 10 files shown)


unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
```diff
@@ -173,9 +173,31 @@ public void SetMassProperties(string objectId, float mass, float drag, float ang
...
+            //this means the metadat for isStanding should always be false for all derived classes
+            if (this.GetType() != typeof(PhysicsRemoteFPSAgentController)) {
+                return null;
+            }
+
+            //if camera is in neither the standing or crouching predetermined locations, return null
+            if (
+                UtilityFunctions.ArePositionsApproximatelyEqual(
+                    m_Camera.transform.localPosition,
+                    standingLocalCameraPosition
```
AlvaroHG, line 187: Would be good to pass in the epsilon here, as epsilons are per use basis rather than bake it in the function

AlvaroHG:
Minor comment, but LGTM!

unity/Assets/Scripts/PhysicsRemoteFPSAgentController.cs
winthos: ok i've updated it so we can pass in a specific epsilon, but it will default to Unity's Vecto3.kEpsilon

winthos: done!

